PR #8, opened 2025-06-20
formatting: run clang-format over all C/C++ files in the repository

pedro-alves-ferreira:
Several files are not formatted according to the /.clang-format file. This PR corrects formatting on all C/C++ files in the repository and introduces a build step on CI that verifies that formatting is correct. 
More than for the sake of consistency, the goal of this PR is to avoid problems caused by "format on save" creating format-only diff chunks in future PRs.


pedro-alves-ferreira:
@vt-tv Thanks, Vincent. Will do!

PR #9, opened 2025-06-20
Draft: Fix reading of data which has not been written yet

pac-work:
- The code contained an assumption that the next written grain will be the one which the reader was waiting for, which is not necessarily the case.
- This commit appears to trigger a deadlock, which I am not sure whether it was in the code before, or whether it gets introduced by these changes. However preliminary investigation suggests the problem was there from before.


pac-work:
The deadlock has been solved by changes similar to the ones proposed in #10.

vt-tv (approved):
As dicusssed on the TSC weekly call, we are going to merge this but there is more work to be done around thread safety and syncrhonisation

vt-tv:
> I am not sure how does one ask for a review here on GitHub... @KimonHoffmann ? May I ask you this way? Or is there a special button / type of comment to use?

top right of the PR page, Reviewers : click on the 'gear' icon then request PR

lib/src/internal/Instance.cpp
```diff
@@ -174,19 +174,25 @@ namespace mxl::lib
         if (writer)
         {
             auto const& id = writer->getId();
-
-            auto const lock = std::lock_guard<std::mutex>{_mutex};
-            if (auto const pos = _writers.find(id); pos != _writers.end())
+            auto removeFlowWatch = false;
             {
-                if ((*pos).second.releaseReference())
+                auto const lock = std::lock_guard<std::mutex>{_mutex};
```
KimonHoffmann: This could use CTAD to avoid mentioning `std::mutex` explicitly.

lib/src/internal/Instance.cpp
```diff
@@ -174,19 +174,25 @@ namespace mxl::lib
...
+                if (auto const pos = _writers.find(id); pos != _writers.end())
                 {
-                    if (dynamic_cast<ContinuousFlowWriter*>((*pos).second.get()) == nullptr)
+                    if ((*pos).second.releaseReference())
                     {
-                        _watcher->removeFlow(id, WatcherType::WRITER);
+                        if (dynamic_cast<ContinuousFlowWriter*>((*pos).second.get()) == nullptr)
```
KimonHoffmann: Could be more concise with:

```cpp
removeFlowWatch = (dynamic_cast<ContinuousFlowWriter*>((*pos).second.get()) == nullptr);
```

lib/src/internal/PosixDiscreteFlowReader.cpp
```diff
@@ -56,49 +56,53 @@ namespace mxl::lib
 
     mxlStatus PosixDiscreteFlowReader::getGrain(std::uint64_t in_index, std::uint64_t in_timeoutNs, GrainInfo* out_grainInfo, std::uint8_t** out_payload)
     {
+        const auto deadline = currentTime(Clock::Realtime) + Duration{static_cast<int64_t>(in_timeoutNs)};
```
KimonHoffmann: Please use the `std::` qualified variants of the `stdint` types in C++ code.

lib/src/internal/PosixDiscreteFlowReader.cpp
```diff
@@ -56,49 +56,53 @@ namespace mxl::lib
...
             {
-                // FIXME: This code is hopelessly broken. As it assumes that the
-                //      next write provides the index of interest, which is not
-                //      guaranteed.
-                if (waitUntilChanged(&flowInfo->discrete.syncCounter, flowInfo->discrete.syncCounter, Duration(in_timeoutNs)))
+                // We remember the sync counter before checking the head index, otherwise we would introduce a race condition:
+                // 1. We check the header index, data won't be available yet.
+                // 2. Writer writes the data and updates the counter.
+                // 3. If we used the current value of the counter for the futex, we would delay everything by 1 grain.
+                auto previousSyncCounter = flowInfo->discrete.syncCounter;
```
KimonHoffmann, line 74: Access to the sync counter, as well as the head index should use atomic operations in order to be well defined, but this is something we should tackle separately, as it requires some finessing to get the C & C++ notions of atomics to match up.

lib/tests/test_flows_timing.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include <ctime>
+#include <filesystem>
+#include <thread>
+#include <catch2/catch_test_macros.hpp>
+#include <mxl/flow.h>
+#include <mxl/mxl.h>
+#include <mxl/time.h>
+#include <uuid.h>
+
+#include <iostream>
```
KimonHoffmann: The include order does not quite match our anticipated standard:

1. Header for TU, if applicable
2. C++ variants of standard C headers (`<c*>`)
3. Standard C++ headers
4. Standard OS headers
5. Non-local library headers
6. Local library headers
7. Local headers (Using double quotes)

lib/tests/test_flows_timing.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+namespace fs = std::filesystem;
```
KimonHoffmann: This alias is unused and we have more recently deprecated the use of such aliases more broadly.

KimonHoffmann (changes requested):
Thanks a lot for taking this issue on! Looks good to me, I only have a few minor, mostly syntactic remarks. The only substantial concern I have is about the use of atomics, as noted below, but I think this is more appropriately handled in a separate PR.

Regarding the mode of operation: Please apply any changes resulting from this or other reviews by amending the original commits in order to keep the history nice and lean.

lib/src/internal/Instance.cpp
pac-work: This line is actually not mine, it was already there and it is on several places in that file. I will fix this in a separate commit in the whole file then, to be consistent (hope that is OK, leaving this discussion opened just that you know about me making this a separate commit).

lib/src/internal/Instance.cpp
pac-work: Using the proposed version instead.

lib/src/internal/PosixDiscreteFlowReader.cpp
pac-work: Fixing.

lib/src/internal/PosixDiscreteFlowReader.cpp
pac-work: Agree, when writing the code, I was a bit surprised with lack of mechanism for protecting the compiler from optimizing out, ensuring atomicity etc. etc. etc. . I take it this has to happen for more places in the codebase anyway, so not doing anything in these regards now. Will be happy to discuss with you if you want so.

lib/tests/test_flows_timing.cpp
pac-work: Removing, thanks, missed this one.

lib/tests/test_flows_timing.cpp
pac-work: The `iostream` was something I forgot from some debugging stage, that is gone now. I tried to reorder the rest according to the mentioned order. Maybe it would be worth if this was written down somewhere, in some coding guidelines? Or auto-checked? Similar to the use of `std::` for `stdint` types. Is there a place where I should put these?

KimonHoffmann: There is a rudimentary attempt in the `.clang-format` config to ensure the order. Not sure where there might be a good place to document this...

lib/tests/test_flows_timing.cpp
```diff
@@ -0,0 +1,86 @@
+#include "../src/internal/Time.hpp"
+#include "Utils.hpp"
```
KimonHoffmann: These two should go to the bottom of the list.

Having `Time.hpp` at the top would would apply to a fictitious `Time.cpp` TU.

lib/tests/test_flows_timing.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+#ifndef __APPLE__
+#   include <Device.h>
+#   include <Packet.h>
+#   include <PcapFileDevice.h>
+#   include <ProtocolType.h>
+#   include <RawPacket.h>
+#   include <UdpLayer.h>
+#endif
```
KimonHoffmann: This block should also go below the standard haeaders. In order to not clutter things up I would not interleave the unconditionally included standard headers with this conditional block, but rather move it completely below the other standard includes.

I **would** move the local, double-quoted includes mentioned above even further down, below the block, though.

lib/tests/test_flows_timing.cpp
pac-work: I was hoping that clang format would have taken care of these things, sorry for missing this one. Fixing manually.

lib/tests/test_flows_timing.cpp
pac-work: Well, all of these were based on an existing test. Hope it is correct now.

PR #20, opened 2025-07-01
Fix signaling to wake up all readers, not a single one

ko80:
There seems to be wrong signaling since wakeOne() invokes do_wake_all() while wakeAll() invokes do_wake_one(). It means that when the writer releases the grain, only one reader is notified while other readers are not.

Change wakeOne() to make it calling do_wake_one() and wakeAll() to call do_wake_all().


ko80:
@vt-tv @lapointejp I'd appreciate it if you review this change.

PR #23, opened 2025-07-02
fix: correct flow writer cancel method to use cancel instead of commit

moleksy:
There seems to be a logical error in mxlFlowWriterCancelSamples(), it invokes commit() on the writer, which is committing the samples rather than discarding them. This means that calls to cancel a write actually ended up committing the data.

This patch changes the implementation to call cancel() instead of commit(), restoring the intended behaviour for cancelled samples.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,6 +19,7 @@ permissions:
   issues: read
   checks: write
   pull-requests: write
+  packages: write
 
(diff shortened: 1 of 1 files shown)


KimonHoffmann (approved):
Another casualty of my recent hasty, sweeping code changes.
Thanks for catching and fixing this!

PR #25, opened 2025-07-02
Add rust to the containers.

vt-tv:


.devcontainer/Dockerfile.debiantrixie
```diff
@@ -76,3 +76,5 @@ RUN git clone https://github.com/microsoft/vcpkg \
  && ./vcpkg/bootstrap-vcpkg.sh --disableMetrics
 
 ENV VCPKG_ROOT="/home/${USERNAME}/vcpkg"
+
```
KimonHoffmann, line 80: Why not use `rustup` from the official package sources, i.e. `apt-get install -y --no-install-recommends rustup`?

vt-tv: To make sure we have a consistant version.  I am pretty sure the versions of rustup packaged with our legacy ubuntu containers or amazonlinux is really old. Same logic as for clang : we pull it from llvm.org instead of relying on older distro packages.

KimonHoffmann: `rustup` is a tool to install rust toolchains and not a specific rust toolchain version. More modern distributions (such as Debian 'Trixie', on which this comment was made) have it in thier official package repositories and in those cases I think we should obtain it from there.

We can do whatever we deem necessary in order to pull the legacy distributions ahead, but I would certainly not want them to drag down everything else to the lowest common denominator.

vt-tv: trixie packages are quite recent. but even unbuntu 24.04 is already 4 versions behind.
Rustup is the recommended way to install rust and cargo according to the rust official docs : https://www.rust-lang.org/tools/install#:~:text=Using%20rustup%20(Recommended)

If you insist on using platform packages for trixie I'll change it but I fear this is the wrong approach.

vt-tv: I would suggest so refine the installation command to specify an exact version so we have consistent builds across all OSes. `curl --proto '=https' --tls1.2 -sSf https://sh.rustup.rs/ | sh -s --default-toolchain=1.88.0`

KimonHoffmann: Yes, `rustup` is the recommended way to install rust. The `rustup` version is not related to the rust toolchain version(s) installed with it. So not being 100% current is not going to be a problem. And to that point:

* Trixe: 1.27
* Questing: 1.27
* Plucky: 1.27
* Oracular: 1.26
* Noble: 1.26

All seem recent enough to me, given that 1.28 is the current upstream version.

vt-tv: Draft PR updated to use rustup from package managers on modern distros.  Configured each dockerfile to use a specific rust toolchain version in order to have predictable and repeatable builds.

.devcontainer/Dockerfile.amazonlinux
```diff
@@ -63,3 +63,4 @@ RUN git clone https://github.com/microsoft/vcpkg \
 
 ENV VCPKG_ROOT=/home/$USERNAME/vcpkg
 
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain=1.88.0
```
garethsb, line 66: It's a pity there's not an easy way to just prohibit TLS 1.0 and 1.1, rather than force TLS 1.2. ChatGPT suggests `--tls-max 1.3` but I don't know if that's really supported!

garethsb (approved):
LGTM

PR #28, opened 2025-07-03
Implement a simple json based configuration mechanism. 

vt-tv:
Implement a simple json based configuration mechanism.  Hook up the history_duration option


vt-tv:
Based on the ticket discussion this seems to the be the favored approach: domain-level history duration if configured + fallback to   hardcoded value if not.

lib/src/internal/Instance.cpp
```diff
@@ -40,6 +43,31 @@ namespace mxl::lib
...
+        /// \param options The options json string to parse
+        /// \param config The parsed json object
+        /// \return true if the parsing was successful, false otherwise
+        bool parseOptionsJson(std::string const& options, picojson::object& config)
+        {
+            picojson::value parsed;
+            std::string err = picojson::parse(parsed, options);
+            if (!err.empty())
+            {
+                return false;
```
lapointejp, line 59: Feels to me we might want to print the error for convenience ?

lib/src/internal/Instance.cpp
```diff
@@ -301,4 +330,70 @@ namespace mxl::lib
...
+                    }
+                }
+                else
+                {
+                    MXL_ERROR("Failed to parse domain specific options: {}", options);
+                }
+            }
+        }
+
+        // If we have am instance level options string, parse it as well.
```
lapointejp: am -> an

lib/src/internal/Instance.cpp
```diff
@@ -301,4 +330,70 @@ namespace mxl::lib
...
+        if (!options.empty())
+        {
+            picojson::object config;
+            if (parseOptionsJson(options, config))
+            {
+                // We are not considering MXL_HISTORY_DURATION_TAG here. we don't want per-instance history durations.
```
lapointejp, line 382: Perhaps we should think of documenting options somewhere ? Especially if some options are only parsed in certain scenarios.

lib/src/internal/Instance.hpp
```diff
@@ -97,6 +97,10 @@ namespace mxl::lib
         ///
         std::size_t garbageCollect() const;
 
+        /// Accessor for the history duration value
+        /// \return The history duration in nanoseconds
+        std::uint64_t getHistoryDuration() const;
```
lapointejp: Maybe name this method `getHistoryDurationNs()` for more clarity just by looking at the name wherever used ?

lib/tests/CMakeLists.txt
```diff
@@ -16,30 +16,35 @@ set_target_properties(mxl-tests
...
-            ../src/internal/PathUtils.cpp
-            ../src/internal/PosixContinuousFlowReader.cpp
-            ../src/internal/PosixDiscreteFlowReader.cpp
-            ../src/internal/SharedMemory.cpp
-            ../src/internal/Thread.cpp
-            ../src/internal/Time.cpp
-            ../src/internal/Timing.cpp
-            ../src/internal/Sync.cpp
-        )
+        test_domainwatcher.cpp
```
lapointejp, line 19: The change in indentation differs from other CMakes. I would go back to the extra indent.

lib/tests/Utils.cpp
```diff
@@ -14,4 +14,27 @@ std::string mxl::tests::readFile(std::filesystem::path const& filepath)
         throw std::runtime_error("Failed to open file: " + filepath.string());
     }
     return {std::istreambuf_iterator<char>(file), std::istreambuf_iterator<char>()};
+}
+
+/// Simple utility function to get the domain path for the tests.
+/// This will return a path to /dev/shm/mxl_domain on Linux and a path in the user's
+/// home directory ($HOME/mxl_domain) on macOS.
+/// \return the path to the mxl domain directory.  The directory may not exist yet.
```
lapointejp: Nitpick: **T**he path [...]

lapointejp (changes requested):

lib/src/internal/Instance.cpp
vt-tv: Agreed.

lib/tests/CMakeLists.txt
vt-tv: I'll try to first understand how this keeps happening. maybe we need something in our clang tidy. opening any cmakelists.txt file in vscode will do this at the moment

lib/src/internal/Instance.hpp
vt-tv: fixed

lib/src/internal/Instance.cpp
vt-tv: fixed

lib/src/internal/Instance.cpp
vt-tv: fixed

lib/src/internal/Instance.cpp
vt-tv: fixed

docs/configuration.md
```diff
@@ -0,0 +1,19 @@
+# Configuration
+
+## Domain level configuration
+
+Domain level configuration is stored in an optional '.options' files stored at the root of the MXL domain.  If present, the MXL SDK will look for specific options defined in the table below and configure itself accordingly.
+
+| Option        | Description                | Default Value |
+|----------------|---------------------------|---------------|
+| `urn:x-mxl:option:history_duration/v1.0"`         | Depth, in nanoseconds, of a ringbuffer         | 100'000'000ns   |
```
lapointejp, line 9: There is an extra quote at the end of the option name.

lib/tests/CMakeLists.txt
lapointejp: I think it happened again.

PR #32, opened 2025-07-08
Fix timestamp to/from index overflows. Add time roundtrip tests

vt-tv:


KimonHoffmann (commented):
The reason why I changed it away the "nano seconds per grain" calculation was that it is quite inaccurate due to truncation.

Before we change it back I would like to have wrapped my head around what that implies with regards to accumulating inaccuracies also with high rates (such as audio sampling rates).

vt-tv:
Could we at least make it _not obviously wrong_ (overflows and truncations) and then refine it ?

vt-tv:
Would you be ok if I commit this change (use 128bit unsigned int) as a temporary measure?  At least it would be correct, until we decide how we deal with this (or IF we need to change it)
```c++
extern "C"
MXL_EXPORT
uint64_t mxlTimestampToHeadIndex(Rational const* editRate, uint64_t timestamp)
{
    if ((editRate == nullptr) || (editRate->denominator == 0) || (editRate->numerator == 0))
    {
        return MXL_UNDEFINED_INDEX;
    }

    auto const nsPerGrain = (__uint128_t)editRate->denominator * (__uint128_t)1'000'000'000 / (__uint128_t)editRate->numerator;
    return timestamp / (uint64_t)nsPerGrain;
}

extern "C"
MXL_EXPORT
uint64_t mxlHeadIndexToTimestamp(Rational const* editRate, uint64_t index)
{
    // Validate the edit rate
    if ((editRate == nullptr) || (editRate->denominator == 0) || (editRate->numerator == 0))
    {
        return MXL_UNDEFINED_INDEX;
    }

    auto const nsPerGrain = (__uint128_t)editRate->denominator * (__uint128_t)1'000'000'000 / (__uint128_t)editRate->numerator;
    return (uint64_t)(index * nsPerGrain);
}
```


Another point : I really don't like the new time functions : they should really not have <head> in their name.  They are about index to/from timestamp, not HEAD index.  the maths work for any index within the ringbuffer. the head is not special.

vt-tv:
> > Another point : I really don't like the new time functions : they should really not have in their name. They are about index to/from timestamp, not HEAD index. the maths work for any index within the ringbuffer. the head is not special.
> 
> I agree with this comment :).

PR update with renamed time functions.  the functions should only refer to 'index'. Head index is an attribute of the ringbuffer flow info.  Changed conversion routines to use uint128 for now.

lib/tests/test_time.cpp
```diff
@@ -1,5 +1,6 @@
 #include <cstdint>
 #include <ctime>
+#include <iostream>
```
pac-work: Not sure the iostream is needed, probably remnant of debugging (I had one in my code as well).

lib/include/mxl/time.h
```diff
@@ -17,59 +17,59 @@ extern "C"
...
      * Index 0 is defined to be index at the beginning of the epoch as defined by SMPTE ST 2059
      *
      * \param[in] editRate The edit rate of the Flow
-     * \return The head index or MXL_UNDEFINED_INDEX if editRate is null or invalid
+     * \return The index or MXL_UNDEFINED_INDEX if editRate is null or invalid
      */
-MXL_EXPORT
-    uint64_t mxlGetCurrentHeadIndex(Rational const* editRate);
+    MXL_EXPORT
+    uint64_t mxlGetCurrentIndex(Rational const* editRate);
```
KimonHoffmann, line 27: For this specific function the word `Head` made sense, because it doesn't operate on arbitrary timestamps but the **current** time.

lib/src/time.cpp
```diff
@@ -11,55 +11,55 @@ uint64_t mxlGetTime()
...
+uint64_t mxlTimestampToIndex(Rational const* editRate, uint64_t timestamp)
 {
     if ((editRate == nullptr) || (editRate->denominator == 0) || (editRate->numerator == 0))
     {
         return MXL_UNDEFINED_INDEX;
     }
 
-    return (timestamp * editRate->numerator + 500'000'000ULL * editRate->denominator) / (1'000'000'000ULL * editRate->denominator);
+    auto const nsPerGrain = (__uint128_t)editRate->denominator * (__uint128_t)1'000'000'000 / (__uint128_t)editRate->numerator;
+    return timestamp / (uint64_t)nsPerGrain;
```
KimonHoffmann: The idea behind switching to 128 bit integers was to avoid the overflow in the original calculation and thereby avoid having to go with the inaccurate calculation via "nanoseconds per grain".

Also I would prefer C++ style casts.

So this would become:
```cpp
// ...
return static_cast<uint64_t>(timestamp * __uint128_t{editRate->numerator} + 500'000'000U * __uint128_t{editRate->denominator}) / (1'000'000'000U * __uint128_t{editRate->denominator});
```

lib/src/time.cpp
```diff
@@ -11,55 +11,55 @@ uint64_t mxlGetTime()
...
 {
     // Validate the edit rate
     if ((editRate == nullptr) || (editRate->denominator == 0) || (editRate->numerator == 0))
     {
         return MXL_UNDEFINED_INDEX;
     }
 
-    return (index * editRate->denominator * 1'000'000'000ULL) / editRate->numerator;
+    auto const nsPerGrain = (__uint128_t)editRate->denominator * (__uint128_t)1'000'000'000 / (__uint128_t)editRate->numerator;
+    return (uint64_t)(index * nsPerGrain);
```
KimonHoffmann: As above.

vt-tv:
> > Another point : I really don't like the new time functions : they should really not have in their name. They are about index to/from timestamp, not HEAD index. the maths work for any index within the ringbuffer. the head is not special.
> 
> Indeed. The word "head" snuck into the names, when I changed it away from "grain", to make it applicable to continuous flows as well. I full agree that names without either of those make more sense for those two.

PR updated already.

lib/include/mxl/time.h
```diff
@@ -17,59 +17,59 @@ extern "C"
...
      * Utility method to help compute how many nanoseconds we need to wait until the beginning of the specified index
-     * \param[in] index The head to wait for
+     * \param[in] index The index to wait for
      * \param[in] editRate The edit rate of the Flow
      * \return How many nanoseconds to wait or MXL_UNDEFINED_INDEX if editRate is null or invalid
      */
-MXL_EXPORT
-    uint64_t mxlGetNsUntilHeadIndex(uint64_t index, Rational const* editRate);
+    MXL_EXPORT
+    uint64_t mxlGetNsUntilIndex(uint64_t index, Rational const* editRate);
```
KimonHoffmann, line 36: A similar argument can be made for this function, but to a lesser extent.

KimonHoffmann (changes requested):

lib/src/time.cpp
vt-tv: I pushed new code.
- This syntax is invalid __uint128_t{editRate->numerator}. it triggers narrowing warning and fails the build.  warning: narrowing conversion of '(int64_t)editRate->Rational::denominator' from 'int64_t' {aka 'long int'} to '__int128 unsigned' [-Wnarrowing]
- I kept the clean nsPerGrain computation without adjustments. I need roundtripping to be predictable.

lib/include/mxl/time.h
vt-tv: actually no.  The current index (based on current time) _may not_ be the ringbuffer head.  the head may be slightly in the past (for example, if receveiving a stream that has origin timestamps and is delayed by a few tens or hundreds of miliseconds.

KimonHoffmann: I'm convinced.

lib/src/time.cpp
KimonHoffmann: The syntax I pasted above is fine, even GCC 10 and clang 15 accept it without any warnings, see:
https://godbolt.org/z/1ch43n1db

Your pasted error message gives rise to the suspicion that there's some other mixup involved in that error/warning message.

KimonHoffmann: If round tripping is the issue this can be achieved with the previous version by dropping the rounding step: `...  + 500'000'000U * __uint128_t{editRate->denominator}`.

vt-tv: I find it more readable this way and will be optimized away by the compiler.

KimonHoffmann: There is nothing to be optimized away, since we're talking about compile time constants. There is a semantic difference, because one construct allows for *silent implicit* narrowing, while the other doesn't. 

But again:
1. The change over to 128 bit only makes sense if we also revert to the previous calculation. The calculation that uses "nano seconds per grain" *does not suffer from overflow*.
2. The variant that uses "nano seconds per grain" is less accurate and already dropped ~1050 frames since the epoch. The problem with round tripping is not due to the calculation, but due to the added rounding step that can just be left out.

vt-tv: I assigned the PR to you. Please fix it the way you want. I find the current code correct, readable and acceptable.

KimonHoffmann: Here you go ... 
... unfortunately I can not request you as a reviewer.

lapointejp (approved):
Thanks for the changes Kimon. Looks good to me.

pac-work (approved):

PR #43, opened 2025-07-16
Feature/getting started: minor changes to make getting started easier

david3a:
devcontainers needed a gpu, created configs with and without for ubuntu 24:04

cleaned up some trailing whitespace


lapointejp:
For the GPU portion, I would keep a single UB24 container and use:
`    "hostRequirements": {
        "gpu": "optional"
    }`

This will set-up GPU only if it is possible, allowing to maintain a single devcontainer for both with and without GPU hosts.

david3a:
@lapointejp not sure how that gets added, I tried changing locally and get missing GPU errors again

lapointejp:
> @lapointejp not sure how that gets added, I tried changing locally and get missing GPU errors again

Can you try my branch ? Seems to work fine for me.
https://github.com/dmf-mxl/mxl/tree/users/lapointejp/use-optional-gpu-ub24-devcontainer

david3a:
We will need this approach for our own CI/CD pipelines anyway, so documenting it well is great. I guess we need to be aware of software supply chain security as well with any strategy.

PR #44, opened 2025-07-17
Simplify bare metal development

pac-work:
- Git ignore files generated by common IDEs.
- Git ignore `CMakeUserPresets.json` to allow simpler profile customization (for example change vcpkg path).


vt-tv (approved):
Simple enough.  Approved.

PR #48, opened 2025-07-18
Add copyright SPDX to all files.  Adopt FSFE REUSE principles

vt-tv:
As discussed during the TSC meeting.
Additionnaly, adopt the FSFE REUSE Principles. this will allow use to use the reuse lint tools in build pipelines.  The reuse annotate command will also make developer's life easier: it can easily annotate most file formats.

- https://reuse.software/
- https://github.com/marketplace/actions/reuse-compliance-check
- https://github.com/fsfe/reuse-tool


vt-tv:
> Sicne I was unable to attend last weeks TSC call, I'm not familiar with the details discussed, so please excuse if you already had this discussion, but I'm wondering whether including the URL is beneficial.
> 
> https://spdx.github.io/spdx-spec/v2.3/file-tags/#h2-file-tags-format
> 
> Includes an example where they just mention the contributors without a year, or a specific reference to a file.
> 
> The reason why the URL rubs me the wrong way is that
> 
> 1. project URLs tend to change over time, and
> 2. the reference is unversioned. So clicking the link will always bring up the most current file, not the one that matches the commit/branch/tag/dist archive at hand.

@felixpou Where did the copyright text that you provided last week come from?  Is this a recommendation from LF?  could we simplify it to `#SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project `

.devcontainer/almalinux/devcontainer.json
```diff
@@ -1,3 +1,5 @@
+// SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project https://github.com/dmf-mxl/mxl/contributors.md
+// SPDX-License-Identifier: Apache-2.0
```
garethsb: C and C++ style comments are not an official part of JSON and rejected by many JSON parsers/validators. I don't like the sidecar .license files because they bloat the directory view, but I think we probably should for formats that don't support comments...

docs/Media eXchange Layer.png.license
```diff
@@ -0,0 +1,3 @@
+SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project https://github.com/dmf-mxl/mxl/contributors.md
+
+SPDX-License-Identifier: Apache-2.0
```
garethsb: Not sure why it's putting a blank line between these in the sidecar files but not in the normal headers?

```suggestion
SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project https://github.com/dmf-mxl/mxl/contributors.md
SPDX-License-Identifier: Apache-2.0
```

lib/tests/data/audio_flow.json
```diff
@@ -1,17 +1,19 @@
...
-    "sample_rate": {
-        "numerator": 48000
-    },
-    "channel_count": 1,
-    "bit_depth": 32,
-    "parents": [],
-    "source_id": "2aa143ac-0ab7-4d75-bc32-5c00c13d186f",
-    "device_id": "169feb2c-3fae-42a5-ae2e-f6f8cbce29cf"
+  "$copyright": "SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project https://github.com/dmf-mxl/mxl/contributors.md",
+  "$license": "SPDX-License-Identifier: Apache-2.0",
```
garethsb: I really don't think we should do this? We should use sidecar .license files.

lib/tests/data/data_flow.json
```diff
@@ -13,4 +15,4 @@
     "numerator": 30000,
     "denominator": 1001
   }
-}
+}
```
garethsb, line 18: ⬇️ Personal preference to keep files with newlines...

garethsb (changes requested):
Agree with @KimonHoffmann, not sure the URL is beneficial.
But biggest problem is with `//` comments or `$copyright` in the JSON files.

docs/Media eXchange Layer.png.license
vt-tv: That seems like the standard behaviour of the RIFE tool.

.devcontainer/almalinux/devcontainer.json
vt-tv: I was careful to use C style comments in files that are explicitly JSONC only (such as the devcontainer files).  You will see that other json files do not have them (such as the vcpkg.json file)

lib/tests/data/data_flow.json
vt-tv: this is what the .clang-format file defaults to.  that would require a change to teh clang-format file and reprocess the repo.

lib/tests/data/audio_flow.json
vt-tv: I don't think it really matters since these are our own test files.  but I don't mind if you think it is important (picojson does not support jsonc hence the extra attributes)

felixpou:
> @felixpou Where did the copyright text that you provided last week come from? Is this a recommendation from LF? could we simplify it to `#SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project `

https://reuse.software/faq/#copyright-symbol

I think this is flexible, but good that we agree on one general way. The simplification you propose is also given as an example.

lib/tests/data/data_flow.json
KimonHoffmann: I previously preferred newlines at EOF, but since git hates trailing newlines I adapted my habits in this regard and now like it the other way ;).

KimonHoffmann:
> > @felixpou Where did the copyright text that you provided last week come from? Is this a recommendation from LF? could we simplify it to `#SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project `
> 
> https://reuse.software/faq/#copyright-symbol
> 
> I think this is flexible, but good that we agree on one general way. The simplification you propose is also given as an example.

Of the options shown there I'd prefer the third option.

garethsb (approved):

lapointejp (approved):
All good :)

PR #50, opened 2025-07-21
Tighten FlowManager error handling and enrich exception contexts

moleksy:
- CreateTemporaryFlowDirectory: throw a filesystem_error with errno on mkdtemp failure.
- PublishFlowDirectory: ensure correct directory perms, then rename(); exceptions bubble up with context.
- SanitizeFlowFormat: map unsupported formats to UNSPECIFIED for better downstream checks in create*Flow().
- WriteFlowDescriptor: now throws filesystem_error if the JSON file can’t be opened or written, instead of silently dropping data.
- createDiscreteFlow & createContinuousFlow: • Sanitize and validate the requested mxlDataFormat up front. • Wrap each I/O/mmap step in try/catch that rethrows with a clear, specific runtime_error message (e.g. “Failed to mmap…”, “Could not create grain directory”, “Failed to open grain [i]”). • On any exception, the temporary directory is remove_all()’ed via a noexcept cleanup in the catch block.

- openFlow/openDiscreteFlow/openContinuousFlow: • Invalid AccessMode now throws std::invalid_argument immediately. • Any SharedMemory or per-grain/channel errors are caught and re-wrapped as runtime_error with detailed context. • Missing files (flow, grain dir, channel buffer) throw filesystem_error with std::errc::no_such_file_or_directory.

- deleteFlow(uuid) & deleteFlow(unique_ptr): • delete_all is wrapped in a try/catch that returns false on any filesystem_error (no unexpected throws). • Logs an MXL_ERROR on failure but preserves the bool return contract.

- listFlows(): • Catches filesystem_error during directory iteration and rethrows it with the failing path and error code. • Catches other std::exception (e.g. bad UUID parsing) and wraps it in a runtime_error.

- Tests (test_flowmanager.cpp) updated to expect these richer exceptions and to cover: • create*Flow format‐mismatch • create*Flow JSON/write failures • openFlow invalid access mode • openFlow missing files • deleteFlow returning false on FS errors • listFlows throwing on directory‐iterator errors

These changes do not alter the public API but significantly improve the debuggability and robustness of every file operation within FlowManager.

### DIFF
--- a/lib/src/internal/FlowManager.cpp
+++ b/lib/src/internal/FlowManager.cpp
@@ -3,11 +3,11 @@
 #include <fstream>
 #include <ios>
 #include <stdexcept>
+#include <system_error>
 #include <unistd.h>
 #include <mxl/flow.h>
 #include <mxl/mxl.h>
 #include <mxl/time.h>
-#include <system_error>
 #include "Logging.hpp"
 #include "PathUtils.hpp"
 #include "SharedMemory.hpp"
@@ -33,8 +33,10 @@ namespace mxl::lib
             auto pathBuffer = (base / ".mxl-tmp-XXXXXXXXXXXXXXXX").string();
             if (::mkdtemp(pathBuffer.data()) == nullptr)
             {
+                auto const error = errno;
+                MXL_ERROR("FlowManager: mkdtemp failed for path '{}' (errno {}: {})", base.string(), error, std::strerror(error));
                 throw std::filesystem::filesystem_error{
-                    "Could not create temporary directory.", base, std::error_code{errno, std::generic_category()}
+                    "FlowManager: Could not create temporary directory.", base, std::error_code{error, std::generic_category()}
                 };
             }
             return pathBuffer;
@@ -67,8 +69,9 @@ namespace mxl::lib
             }
             else
             {
+                MXL_ERROR("FlowManager: Failed to create flow resource definition file '{}'", flowJsonFile.string());
                 throw std::filesystem::filesystem_error(
-                    "Failed to create flow resource definition.", flowJsonFile, std::make_error_code(std::errc::io_error));
+                    "FlowManager: Failed to create flow resource definition.", flowJsonFile, std::make_error_code(std::errc::io_error));
             }
         }
 
@@ -92,8 +95,9 @@ namespace mxl::lib
     {
         if (!exists(in_mxlDomain) || !is_directory(in_mxlDomain))
         {
+            MXL_ERROR("FlowManager: Domain path '{}' does not exist or is not a directory", in_mxlDomain.string());
             throw std::filesystem::filesystem_error(
-                "Path does not exist or is not a directory", in_mxlDomain, std::make_error_code(std::errc::no_such_file_or_directory));
+                "FlowManager: Path does not exist or is not a directory", in_mxlDomain, std::make_error_code(std::errc::no_such_file_or_directory));
         }
     }
 
@@ -106,7 +110,8 @@ namespace mxl::lib
         flowFormat = sanitizeFlowFormat(flowFormat);
         if (!mxlIsDiscreteDataFormat(flowFormat))
         {
-            throw std::runtime_error("Attempt to create discrete flow with unsupported or non matching format.");
+            MXL_ERROR("FlowManager: Attempt to create discrete flow '{}' with unsupported format {}", uuidString, static_cast<int>(flowFormat));
+            throw std::runtime_error("FlowManager: Attempt to create discrete flow with unsupported or non matching format.");
         }
 
         auto const tempDirectory = createTemporaryFlowDirectory(_mxlDomain);
@@ -119,8 +124,9 @@ namespace mxl::lib
             auto readAccessFile = makeFlowAccessFilePath(tempDirectory);
             if (auto out = std::ofstream{readAccessFile, std::ios::out | std::ios::trunc}; !out)
             {
+                MXL_ERROR("FlowManager: Failed to create flow access file '{}'", readAccessFile.string());
                 throw std::filesystem::filesystem_error(
-                    "Failed to create flow access file.", readAccessFile, std::make_error_code(std::errc::file_exists));
+                    "FlowManager: Failed to create flow access file.", readAccessFile, std::make_error_code(std::errc::file_exists));
             }
 
             auto flowData = std::make_unique<DiscreteFlowData>(makeFlowDataFilePath(tempDirectory).string().c_str(), AccessMode::CREATE_READ_WRITE);
@@ -136,7 +142,11 @@ namespace mxl::lib
             info.discrete.syncCounter = 0;
 
             auto const grainDir = makeGrainDirectoryName(tempDirectory);
-            create_directory(grainDir);
+            if (!std::filesystem::create_directory(grainDir))
+            {
+                MXL_ERROR("FlowManager: Could not create grain directory '{}'", grainDir.string());
+                throw std::filesystem::filesystem_error("FlowManager: Could not create grain directory", grainDir, std::make_error_code(std::errc::io_error));
+            }
 
             for (auto i = std::size_t{0}; i < grainCount; ++i)
             {
@@ -152,7 +162,16 @@ namespace mxl::lib
                 gInfo.deviceIndex = -1;
             }
 
-            publishFlowDirectory(tempDirectory, makeFlowDirectoryName(_mxlDomain, uuidString));
+            auto const finalDir = makeFlowDirectoryName(_mxlDomain, uuidString);
+            try
+            {
+                publishFlowDirectory(tempDirectory, finalDir);
+            }
+            catch (std::exception const& e)
+            {
+                MXL_ERROR("FlowManager: Failed to publish flow directory from '{}' to '{}': {}", tempDirectory.string(), finalDir.string(), e.what());
+                throw std::filesystem::filesystem_error("FlowManager: Failed to publish flow directory", finalDir, std::make_error_code(std::errc::io_error));
+            }
 
             return flowData;
         }
@@ -177,7 +196,8 @@ namespace mxl::lib
         flowFormat = sanitizeFlowFormat(flowFormat);
         if (!mxlIsContinuousDataFormat(flowFormat))
         {
-            throw std::runtime_error("Attempt to create continuous flow with unsupported or non matching format.");
+            MXL_ERROR("FlowManager: Attempt to create continuous flow '{}' with unsupported format {}", uuidString, static_cast<int>(flowFormat));
+            throw std::runtime_error("FlowManager: Attempt to create continuous flow with unsupported or non matching format.");
         }
 
         auto const tempDirectory = createTemporaryFlowDirectory(_mxlDomain);
@@ -186,7 +206,16 @@ namespace mxl::lib
             // Write the json file to disk.
             writeFlowDescriptor(tempDirectory, flowDef);
 
-            auto flowData = std::make_unique<ContinuousFlowData>(makeFlowDataFilePath(tempDirectory).string().c_str(), AccessMode::CREATE_READ_WRITE);
+            auto flowData = std::unique_ptr<ContinuousFlowData>{};
+            try
+            {
+                flowData = std::make_unique<ContinuousFlowData>(makeFlowDataFilePath(tempDirectory).string().c_str(), AccessMode::CREATE_READ_WRITE);
+            }
+            catch (std::exception const& e)
+            {
+                MXL_ERROR("FlowManager: Failed to create continuous flow data for '{}': {}", uuidString, e.what());
+                throw std::runtime_error(std::string("FlowManager: Failed to mmap continuous flow data: ") + e.what());
+            }
 
             auto& info = *flowData->flowInfo();
             info.version = 1;
@@ -194,14 +223,31 @@ namespace mxl::lib
 
             info.common = initCommonFlowInfo(flowId, flowFormat);
 
-            info.continuous = {};
+            info.continuous = ContinuousFlowInfo{};
             info.continuous.sampleRate = sampleRate;
             info.continuous.channelCount = channelCount;
             info.continuous.bufferLength = bufferLength;
 
-            flowData->openChannelBuffers(makeChannelDataFilePath(tempDirectory).string().c_str(), sampleWordSize);
+            try
+            {
+                flowData->openChannelBuffers(makeChannelDataFilePath(tempDirectory).string().c_str(), sampleWordSize);
+            }
+            catch (std::exception const& e)
+            {
+                MXL_ERROR("FlowManager: Failed to open channel buffers for continuous flow '{}': {}", uuidString, e.what());
+                throw std::runtime_error(std::string("FlowManager: Failed to open channel buffers: ") + e.what());
+            }
 
-            publishFlowDirectory(tempDirectory, makeFlowDirectoryName(_mxlDomain, uuidString));
+            auto const FinalDir = makeFlowDirectoryName(_mxlDomain, uuidString);
+            try
+            {
+                publishFlowDirectory(tempDirectory, FinalDir);
+            }
+            catch (std::exception const& e)
+            {
+                MXL_ERROR("FlowManager: Failed to publish continuous flow directory from '{}' to '{}': {}", tempDirectory.string(), FinalDir.string(), e.what());
+                throw std::runtime_error(std::string("FlowManager: Failed to publish flow dir: ") + e.what());
+            }
 
             return flowData;
         }
@@ -217,7 +263,8 @@ namespace mxl::lib
     {
         if (in_mode == AccessMode::CREATE_READ_WRITE)
         {
-            throw std::invalid_argument("Attempt to open flow with invalid access mode.");
+            MXL_ERROR("FlowManager: Attempt to open flow '{}' with invalid access mode CREATE_READ_WRITE", uuids::to_string(in_flowId));
+            throw std::invalid_argument("FlowManager: Attempt to open flow with invalid access mode.");
         }
 
         auto uuid = uuids::to_string(in_flowId);
@@ -226,23 +273,60 @@ namespace mxl::lib
         // Verify that the flow file exists.
         if (auto const flowFile = makeFlowDataFilePath(base); exists(flowFile))
         {
-            auto flowSegment = SharedMemoryInstance<Flow>{flowFile.string().c_str(), in_mode, 0U};
-            if (auto const flowFormat = flowSegment.get()->info.common.format; mxlIsDiscreteDataFormat(flowFormat))
+            auto flowSegment = SharedMemoryInstance<Flow>{};
+            try
             {
-                return openDiscreteFlow(base, std::move(flowSegment));
+                flowSegment = SharedMemoryInstance<Flow>(flowFile.string().c_str(), in_mode, 0U);
+            }
+            catch (std::exception const& e)
+            {
+                MXL_ERROR("FlowManager: Failed to open flow data segment '{}': {}", flowFile.string(), e.what());
+                throw std::filesystem::filesystem_error("FlowManager: Failed to open flow data segment", flowFile, std::make_error_code(std::errc::io_error));
+            }
+
+            auto const* flow = flowSegment.get();
+            if (!flow)
+            {
+                MXL_ERROR("FlowManager: Failed to access flow data for '{}': invalid shared memory segment", uuid);
+                throw std::runtime_error("FlowManager: Failed to access flow data: invalid shared memory segment");
+            }
+            
+            auto const flowFormat = flow->info.common.format;
+            if (mxlIsDiscreteDataFormat(flowFormat))
+            {
+                try
+                {
+                    return openDiscreteFlow(base, std::move(flowSegment));
+                }
+                catch (std::exception const& e)
+                {
+                    MXL_ERROR("FlowManager: Failed to open discrete flow '{}': {}", uuid, e.what());
+                    throw std::filesystem::filesystem_error("FlowManager: Failed to open discrete flow", base, std::make_error_code(std::errc::io_error));
+                }
             }
             else if (mxlIsContinuousDataFormat(flowFormat))
             {
-                return openContinuousFlow(base, std::move(flowSegment));
+                try
+                {
+                    return openContinuousFlow(base, std::move(flowSegment));
+                }
+                catch (std::exception const& e)
+                {
+                    MXL_ERROR("FlowManager: Failed to open continuous flow '{}': {}", uuid, e.what());
+                    throw std::runtime_error(std::string("FlowManager: Failed to open continuous flow: ") + e.what());
+                }
             }
             else
             {
-                throw std::runtime_error("Attempt to open flow with unsupported data format.");
+                // This should never happen for a valid flow.
+                MXL_ERROR("FlowManager: Attempt to open flow '{}' with unsupported data format {}", uuid, static_cast<int>(flowFormat));
+                throw std::runtime_error("FlowManager: Attempt to open flow with unsupported data format.");
             }
         }
         else
         {
-            throw std::filesystem::filesystem_error("Flow file not found", flowFile, std::make_error_code(std::errc::no_such_file_or_directory));
+            MXL_ERROR("FlowManager: Flow file not found for '{}' at '{}'", uuid, flowFile.string());
+            throw std::filesystem::filesystem_error("FlowManager: Flow file not found", flowFile, std::make_error_code(std::errc::no_such_file_or_directory));
         }
     }
 
@@ -257,18 +341,35 @@ namespace mxl::lib
             auto const grainDir = makeGrainDirectoryName(flowDir);
             if (exists(grainDir) && is_directory(grainDir))
             {
-                // Open each grain
+                // Open each grain with per-item error handling
                 for (auto i = 0U; i < grainCount; ++i)
                 {
                     auto const grainPath = makeGrainDataFilePath(grainDir, i).string();
                     MXL_TRACE("Opening grain: {}", grainPath);
-                    flowData->emplaceGrain(grainPath.c_str(), 0U);
+
+                    try
+                    {
+                        flowData->emplaceGrain(grainPath.c_str(), /*payloadSize=*/0U);
+                    }
+                    catch (std::exception const& e)
+                    {
+                        MXL_ERROR("FlowManager: Failed to open grain [{}] for flow '{}': {}", i, flowDir.string(), e.what());
+                        throw std::filesystem::filesystem_error(std::string("FlowManager: Failed to open grain [") + std::to_string(i) + "]", 
+                            makeGrainDataFilePath(grainDir, i), std::make_error_code(std::errc::io_error));
+                    }
+                    catch (...)
+                    {
+                        MXL_ERROR("FlowManager: Failed to open grain [{}] for flow '{}': unknown error", i, flowDir.string());
+                        throw std::filesystem::filesystem_error(std::string("FlowManager: Failed to open grain [") + std::to_string(i) + "]: unknown error", 
+                            makeGrainDataFilePath(grainDir, i), std::make_error_code(std::errc::io_error));
+                    }
                 }
             }
             else
             {
+                MXL_ERROR("FlowManager: Grain directory not found for flow at '{}': '{}'", flowDir.string(), grainDir.string());
                 throw std::filesystem::filesystem_error(
-                    "Grain directory not found.", grainDir, std::make_error_code(std::errc::no_such_file_or_directory));
+                    "FlowManager: Grain directory not found.", grainDir, std::make_error_code(std::errc::no_such_file_or_directory));
             }
         }
 
@@ -279,7 +380,27 @@ namespace mxl::lib
         SharedMemoryInstance<Flow>&& sharedFlowInstance)
     {
         auto flowData = std::make_unique<ContinuousFlowData>(std::move(sharedFlowInstance));
-        flowData->openChannelBuffers(makeChannelDataFilePath(flowDir).string().c_str(), 0U);
+
+        // Verify that the channel‐buffers file actually exists before attempting to open it
+        auto const channelPath = makeChannelDataFilePath(flowDir);
+        if (!std::filesystem::exists(channelPath))
+        {
+            MXL_ERROR("FlowManager: Channel buffer file not found for flow at '{}': '{}'", flowDir.string(), channelPath.string());
+            throw std::filesystem::filesystem_error(
+                "FlowManager: Channel buffer file not found.", channelPath, std::make_error_code(std::errc::no_such_file_or_directory));
+        }
+
+        // Open the channel buffers (may throw on I/O or mmap failure)
+        try
+        {
+            flowData->openChannelBuffers(channelPath.string().c_str(), /*payloadSize=*/0U);
+        }
+        catch (std::exception const& e)
+        {
+            MXL_ERROR("FlowManager: Failed to open continuous channel buffers at '{}': {}", channelPath.string(), e.what());
+            throw std::runtime_error(std::string("FlowManager: Failed to open continuous channel buffers: ") + e.what());
+        }
+
         return flowData;
     }
 
@@ -295,7 +416,21 @@ namespace mxl::lib
             // Close the flow
             flowData.reset();
 
-            return deleteFlow(id);
+            // Attempt filesystem deletion, catching any errors
+            try
+            {
+                return deleteFlow(id);
+            }
+            catch (std::exception const& e)
+            {
+                MXL_ERROR("FlowManager: Failed to delete flow {}: {}", uuids::to_string(id), e.what());
+                return false;
+            }
+            catch (...)
+            {
+                MXL_ERROR("FlowManager: Failed to delete flow {}: unknown error", uuids::to_string(id));
+                return false;
+            }
         }
         return false;
     }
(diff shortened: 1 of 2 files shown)


moleksy:
@KimonHoffmann @vt-tv 
I'd appreciate it if you could review this PR.
Thank you!

lib/src/internal/FlowManager.cpp
```diff
@@ -33,8 +33,10 @@ namespace mxl::lib
             auto pathBuffer = (base / ".mxl-tmp-XXXXXXXXXXXXXXXX").string();
             if (::mkdtemp(pathBuffer.data()) == nullptr)
             {
+                auto const error = errno;
+                MXL_ERROR("FlowManager: mkdtemp failed for path '{}' (errno {}: {})", base.string(), error, std::strerror(error));
                 throw std::filesystem::filesystem_error{
-                    "Could not create temporary directory.", base, std::error_code{errno, std::generic_category()}
+                    "FlowManager: Could not create temporary directory.", base, std::error_code{error, std::generic_category()}
```
lapointejp: I'm not completely opposed to prefixing logs to track the class more easily, but at the same time I think it should be consistent across the code base.
I'm also wondering if there really is value in this, given that spdlog will catch the file & line number of the log.

Topic for a TSC call @vt-tv ?

lib/src/internal/FlowManager.cpp
```diff
@@ -92,8 +95,9 @@ namespace mxl::lib
     {
         if (!exists(in_mxlDomain) || !is_directory(in_mxlDomain))
         {
+            MXL_ERROR("FlowManager: Domain path '{}' does not exist or is not a directory", in_mxlDomain.string());
             throw std::filesystem::filesystem_error(
-                "Path does not exist or is not a directory", in_mxlDomain, std::make_error_code(std::errc::no_such_file_or_directory));
+                "FlowManager: Path does not exist or is not a directory", in_mxlDomain, std::make_error_code(std::errc::no_such_file_or_directory));
```
lapointejp: Nitpick: missing period at the end of the exception message, which differs from the rest of the exceptions. I know it was already missing before, but while we're here :).

lib/src/internal/FlowManager.cpp
```diff
@@ -136,7 +142,12 @@ namespace mxl::lib
             info.discrete.syncCounter = 0;
 
             auto const grainDir = makeGrainDirectoryName(tempDirectory);
-            create_directory(grainDir);
+            if (!std::filesystem::create_directory(grainDir))
+            {
+                MXL_ERROR("FlowManager: Could not create grain directory '{}'", grainDir.string());
+                throw std::filesystem::filesystem_error(
+                    "FlowManager: Could not create grain directory", grainDir, std::make_error_code(std::errc::io_error));
```
lapointejp: Missing period at the end of the exception message. I'll stop commenting on further occurrences and let you do a pass if that's okay.

lib/src/internal/FlowManager.cpp
```diff
@@ -186,22 +208,51 @@ namespace mxl::lib
...
+                flowData->openChannelBuffers(makeChannelDataFilePath(tempDirectory).string().c_str(), sampleWordSize);
+            }
+            catch (std::exception const& e)
+            {
+                MXL_ERROR("FlowManager: Failed to open channel buffers for continuous flow '{}': {}", uuidString, e.what());
+                throw std::runtime_error(std::string("FlowManager: Failed to open channel buffers: ") + e.what());
+            }
 
-            publishFlowDirectory(tempDirectory, makeFlowDirectoryName(_mxlDomain, uuidString));
+            auto const FinalDir = makeFlowDirectoryName(_mxlDomain, uuidString);
```
lapointejp: Should lowercase the first F of finalDir.

lib/tests/test_flowmanager.cpp
```diff
@@ -202,3 +206,345 @@ TEST_CASE("Flow Manager : Create Audio Flow Structure", "[flow manager]")
...
+    //
+    // 10) unsupported formats should be rejected
+    //
+    auto const badId = *uuids::uuid::from_string("44444444-4444-4444-4444-444444444444");
+    REQUIRE_THROWS_AS(manager->createDiscreteFlow(badId, flowDef1, MXL_DATA_FORMAT_UNSPECIFIED, 1, grainRate, 128), std::runtime_error);
+    REQUIRE_THROWS_AS(manager->createContinuousFlow(badId, flowDef2, MXL_DATA_FORMAT_VIDEO, sampleRate, 1, 4, 1024), std::runtime_error);
+}
+
+// Helper to make a unique temp domain
+static std::filesystem::path makeTempDomain()
```
lapointejp: Consider declaring this in an anonymous namespace before test cases ?

lib/tests/test_flowmanager.cpp
```diff
@@ -202,3 +206,345 @@ TEST_CASE("Flow Manager : Create Audio Flow Structure", "[flow manager]")
...
+    // Manually remove the descriptor JSON to simulate corruption
+    auto flowDir = makeFlowDirectoryName(domain, uuids::to_string(id));
+    auto descFile = makeFlowDescriptorFilePath(flowDir);
+    REQUIRE(std::filesystem::exists(descFile));
+    std::filesystem::remove(descFile);
+
+    // Current behavior: listFlows still reports the ID (only checks directory structure)
+    // \todo : DESIGN DECISION NEEDED - Should corrupted flows be:
+    //   a) Filtered out from listFlows() for safety, or
+    //   b) Listed but marked as corrupted, or
```
lapointejp: I think b would be my preferred option. @vt-tv, @KimonHoffmann opinions ?

lapointejp (changes requested):
Thank you for your contribution !
I've commented on a few nitpicks - nothing major, but would be good for consistency.

lib/src/internal/FlowManager.cpp
vt-tv: I would prefer not doing it since SPDLOG should be able to prefix the log location (filename, lineno) automatically.

moleksy: @lapointejp @vt-tv Should I remove it then from DomainWatcher too?

lib/tests/test_flowmanager.cpp
vt-tv: I would simply filter them out.

lib/src/internal/FlowManager.cpp
```diff
@@ -152,7 +163,17 @@ namespace mxl::lib
...
+            auto const finalDir = makeFlowDirectoryName(_mxlDomain, uuidString);
+            try
+            {
+                publishFlowDirectory(tempDirectory, finalDir);
+            }
+            catch (std::exception const& e)
+            {
+                MXL_ERROR("FlowManager: Failed to publish flow directory from '{}' to '{}': {}", tempDirectory.string(), finalDir.string(), e.what());
+                throw std::filesystem::filesystem_error(
+                    "FlowManager: Failed to publish flow directory", finalDir, std::make_error_code(std::errc::io_error));
```
KimonHoffmann: I don't like swallowing the original exception in favor of a generic `std::filesystem_error`.

If you want to add a log statement in this case this is fine by me, but then I would choose a pattern like this:

```cpp
try
{
    // ...
}
catch (std::exception const& e)
{
    MXL_ERROR("Failed to publish flow directory from '{}' to '{}': {}", tempDirectory.string(), finalDir.string(), e.what());
    throw;
}
catch (...)
{
    MXL_ERROR("Failed to publish flow directory from '{}' to '{}': {}", tempDirectory.string(), finalDir.string(), "An unknown error occurred.");
    throw;
}
```

lib/src/internal/FlowManager.cpp
```diff
@@ -177,7 +198,8 @@ namespace mxl::lib
         flowFormat = sanitizeFlowFormat(flowFormat);
         if (!mxlIsContinuousDataFormat(flowFormat))
         {
-            throw std::runtime_error("Attempt to create continuous flow with unsupported or non matching format.");
+            MXL_ERROR("FlowManager: Attempt to create continuous flow '{}' with unsupported format {}", uuidString, static_cast<int>(flowFormat));
+            throw std::runtime_error("FlowManager: Attempt to create continuous flow with unsupported or non matching format.");
```
KimonHoffmann: Please use curly braces for initialization.

lib/src/internal/FlowManager.cpp
KimonHoffmann: Please use curly braces for initialization.

lib/src/internal/FlowManager.cpp
```diff
@@ -119,8 +124,9 @@ namespace mxl::lib
             auto readAccessFile = makeFlowAccessFilePath(tempDirectory);
             if (auto out = std::ofstream{readAccessFile, std::ios::out | std::ios::trunc}; !out)
             {
+                MXL_ERROR("FlowManager: Failed to create flow access file '{}'", readAccessFile.string());
                 throw std::filesystem::filesystem_error(
-                    "Failed to create flow access file.", readAccessFile, std::make_error_code(std::errc::file_exists));
+                    "FlowManager: Failed to create flow access file.", readAccessFile, std::make_error_code(std::errc::file_exists));
```
KimonHoffmann: Please use curly braces for initialization.

lib/src/internal/FlowManager.cpp
```diff
@@ -106,7 +110,8 @@ namespace mxl::lib
         flowFormat = sanitizeFlowFormat(flowFormat);
         if (!mxlIsDiscreteDataFormat(flowFormat))
         {
-            throw std::runtime_error("Attempt to create discrete flow with unsupported or non matching format.");
+            MXL_ERROR("FlowManager: Attempt to create discrete flow '{}' with unsupported format {}", uuidString, static_cast<int>(flowFormat));
+            throw std::runtime_error("FlowManager: Attempt to create discrete flow with unsupported or non matching format.");
```
KimonHoffmann: Please use curly braces for initialization.

lib/src/internal/FlowManager.cpp
KimonHoffmann: Please use curly braces for initialization.

lib/src/internal/FlowManager.cpp
```diff
@@ -67,8 +69,9 @@ namespace mxl::lib
             }
             else
             {
+                MXL_ERROR("FlowManager: Failed to create flow resource definition file '{}'", flowJsonFile.string());
                 throw std::filesystem::filesystem_error(
-                    "Failed to create flow resource definition.", flowJsonFile, std::make_error_code(std::errc::io_error));
+                    "FlowManager: Failed to create flow resource definition.", flowJsonFile, std::make_error_code(std::errc::io_error));
```
KimonHoffmann: Please use curly braces for initialization.

lib/src/internal/FlowManager.cpp
KimonHoffmann: Please use curly braces for initialization.

lib/src/internal/FlowManager.cpp
```diff
@@ -186,22 +208,51 @@ namespace mxl::lib
...
+            auto flowData = std::unique_ptr<ContinuousFlowData>{};
+            try
+            {
+                flowData = std::make_unique<ContinuousFlowData>(makeFlowDataFilePath(tempDirectory).string().c_str(), AccessMode::CREATE_READ_WRITE);
+            }
+            catch (std::exception const& e)
+            {
+                MXL_ERROR("FlowManager: Failed to create continuous flow data for '{}': {}", uuidString, e.what());
+                throw std::runtime_error(std::string("FlowManager: Failed to mmap continuous flow data: ") + e.what());
+            }
```
KimonHoffmann: 1. As above wrt. rethrowing the original exception instead of replacing it with something generic.
2. Please use curly braces for initialization of the `std::string`, as well as the exception.

lib/src/internal/FlowManager.cpp
```diff
@@ -186,22 +208,51 @@ namespace mxl::lib
...
 
             auto& info = *flowData->flowInfo();
             info.version = 1;
             info.size = sizeof info;
 
             info.common = initCommonFlowInfo(flowId, flowFormat);
 
-            info.continuous = {};
+            info.continuous = ContinuousFlowInfo{};
```
KimonHoffmann: Please keep the original implicit form, or is there a specific reason why that needed to be replaced?

lib/src/internal/FlowManager.cpp
```diff
@@ -186,22 +208,51 @@ namespace mxl::lib
...
-            flowData->openChannelBuffers(makeChannelDataFilePath(tempDirectory).string().c_str(), sampleWordSize);
+            try
+            {
+                flowData->openChannelBuffers(makeChannelDataFilePath(tempDirectory).string().c_str(), sampleWordSize);
+            }
+            catch (std::exception const& e)
+            {
+                MXL_ERROR("FlowManager: Failed to open channel buffers for continuous flow '{}': {}", uuidString, e.what());
+                throw std::runtime_error(std::string("FlowManager: Failed to open channel buffers: ") + e.what());
+            }
```
KimonHoffmann: As above wrt to rethrowing the original exception.

lib/src/internal/FlowManager.cpp
```diff
@@ -226,23 +278,63 @@ namespace mxl::lib
...
             {
-                return openDiscreteFlow(base, std::move(flowSegment));
+                flowSegment = SharedMemoryInstance<Flow>(flowFile.string().c_str(), in_mode, 0U);
+            }
+            catch (std::exception const& e)
+            {
+                MXL_ERROR("FlowManager: Failed to open flow data segment '{}': {}", flowFile.string(), e.what());
+                throw std::filesystem::filesystem_error(
+                    "FlowManager: Failed to open flow data segment", flowFile, std::make_error_code(std::errc::io_error));
+            }
```
KimonHoffmann: 1. As above wrt to rethrowing the original exception.
2. I tend to prefer the conciseness of the original and am not sure whether it's worth sacrificing this for the added log statement alone.

lib/src/internal/FlowManager.cpp
```diff
@@ -226,23 +278,63 @@ namespace mxl::lib
...
+
+            auto const* flow = flowSegment.get();
+            if (!flow)
```
KimonHoffmann: I'm not sure whether this can actually happen, because the constructor would show in this case. Have you checked?

lib/src/internal/FlowManager.cpp
```diff
@@ -226,23 +278,63 @@ namespace mxl::lib
...
-                return openContinuousFlow(base, std::move(flowSegment));
+                try
+                {
+                    return openContinuousFlow(base, std::move(flowSegment));
+                }
+                catch (std::exception const& e)
+                {
+                    MXL_ERROR("FlowManager: Failed to open continuous flow '{}': {}", uuid, e.what());
+                    throw std::runtime_error(std::string("FlowManager: Failed to open continuous flow: ") + e.what());
+                }
```
KimonHoffmann: As above with regards to rethrowing.

lib/src/internal/FlowManager.cpp
```diff
@@ -257,18 +349,38 @@ namespace mxl::lib
...
+                            std::make_error_code(std::errc::io_error));
+                    }
+                    catch (...)
+                    {
+                        MXL_ERROR("FlowManager: Failed to open grain [{}] for flow '{}': unknown error", i, flowDir.string());
+                        throw std::filesystem::filesystem_error(
+                            std::string("FlowManager: Failed to open grain [") + std::to_string(i) + "]: unknown error.",
+                            makeGrainDataFilePath(grainDir, i),
+                            std::make_error_code(std::errc::io_error));
+                    }
```
KimonHoffmann: As above with regards to rethrowing.

lib/src/internal/FlowManager.cpp
```diff
@@ -279,7 +391,27 @@ namespace mxl::lib
...
+        // Open the channel buffers (may throw on I/O or mmap failure)
+        try
+        {
+            flowData->openChannelBuffers(channelPath.string().c_str(), /*payloadSize=*/0U);
+        }
+        catch (std::exception const& e)
+        {
+            MXL_ERROR("FlowManager: Failed to open continuous channel buffers at '{}': {}", channelPath.string(), e.what());
+            throw std::runtime_error(std::string("FlowManager: Failed to open continuous channel buffers: ") + e.what());
+        }
```
KimonHoffmann: As above wrt to rethrowing the original exception.

lib/src/internal/FlowManager.cpp
```diff
@@ -295,7 +427,21 @@ namespace mxl::lib
...
+            catch (std::exception const& e)
+            {
+                MXL_ERROR("FlowManager: Failed to delete flow {}: {}", uuids::to_string(id), e.what());
+                return false;
+            }
+            catch (...)
+            {
+                MXL_ERROR("FlowManager: Failed to delete flow {}: unknown error", uuids::to_string(id));
+                return false;
+            }
```
KimonHoffmann: As above wrt to rethrowing the original exception.

lib/src/internal/FlowManager.cpp
```diff
@@ -305,8 +451,28 @@ namespace mxl::lib
...
+        {
+            auto const removed = std::filesystem::remove_all(flowPath);
+            if (removed == 0)
+            {
+                MXL_WARN("Flow not found or already deleted: {}", uuid);
+                return false;
+            }
+            return true;
+        }
+        catch (std::filesystem::filesystem_error const& e)
```
KimonHoffmann: Please catch `std::exception` instead, because the call may also show `std::bad_alloc`.

lib/src/internal/FlowManager.cpp
```diff
@@ -305,8 +451,28 @@ namespace mxl::lib
...
-        auto const base = std::filesystem::path{_mxlDomain};
-        return (remove_all(makeFlowDirectoryName(base, uuid)) != 0);
+        // Compute the flow directory path
+        auto const flowPath = makeFlowDirectoryName(_mxlDomain, uuid);
+        try
+        {
+            auto const removed = std::filesystem::remove_all(flowPath);
+            if (removed == 0)
+            {
+                MXL_WARN("Flow not found or already deleted: {}", uuid);
```
KimonHoffmann: I'm not sure this warrants a warning. This may be benign and intentional by the caller, which is why the information is passed on to it via the returned boolean.

lib/tests/test_flowmanager.cpp
```diff
@@ -1,16 +1,31 @@
...
 
 using namespace mxl::lib;
 
+namespace
+{
+    // Helper to make a unique temp domain
+    std::filesystem::path makeTempDomain()
+    {
+        char tmpl[] = "/dev/shm/mxl_test_domainXXXXXX";
+        REQUIRE(mkdtemp(tmpl) != nullptr); // mkdtemp is in global namespace
```
KimonHoffmann: Please add full qualification for `mkdtemp`.

lib/tests/test_flowmanager.cpp
KimonHoffmann: I'd tend to to keep the current behaviour, through checks for corrupted flows are error prone (i.e. prone to miss the more subtle aspect of of what it means for a specific type of flow to be "valid") and lead to a lot of duplicated logic.

If a consumer tries to open the flow it will get an error (from the prime authority that can not get out of sync). If a crashed producer tries to recreate the flow it would clean up the directory beforehand.

KimonHoffmann (changes requested):
Forgot to tick "Request changes" before, sorry.

lib/src/internal/FlowManager.cpp
moleksy: You are right, upon inspection, it shouldn't happen, I'll remove it.

lib/src/internal/FlowManager.cpp
lapointejp: > @lapointejp @vt-tv Should I remove it then from DomainWatcher too?

Yes I would. Just to keep consistency.

vt-tv (approved):
Thanks for applying the requested changes.

lib/src/internal/FlowManager.cpp
```diff
@@ -139,7 +141,10 @@ namespace mxl::lib
             info.discrete.syncCounter = 0;
 
             auto const grainDir = makeGrainDirectoryName(tempDirectory);
-            create_directory(grainDir);
+            if (!std::filesystem::create_directory(grainDir))
```
KimonHoffmann: Please keep the ADL based overload resolution instead of fully qualifying the call.

lib/src/internal/FlowManager.cpp
```diff
@@ -230,7 +237,10 @@ namespace mxl::lib
         if (auto const flowFile = makeFlowDataFilePath(base); exists(flowFile))
         {
             auto flowSegment = SharedMemoryInstance<Flow>{flowFile.string().c_str(), in_mode, 0U};
-            if (auto const flowFormat = flowSegment.get()->info.common.format; mxlIsDiscreteDataFormat(flowFormat))
+
+            auto const* flow = flowSegment.get();
+            auto const flowFormat = flow->info.common.format;
+            if (mxlIsDiscreteDataFormat(flowFormat))
```
KimonHoffmann: I don't think pulling the definitions into the outer scope, just to have a separate variable name for the result of `flowSegment.get()` is an improvement. Could you please revert this change?

lib/src/internal/FlowManager.cpp
```diff
@@ -282,7 +294,18 @@ namespace mxl::lib
         SharedMemoryInstance<Flow>&& sharedFlowInstance)
     {
         auto flowData = std::make_unique<ContinuousFlowData>(std::move(sharedFlowInstance));
-        flowData->openChannelBuffers(makeChannelDataFilePath(flowDir).string().c_str(), 0U);
+
+        // Verify that the channel‐buffers file actually exists before attempting to open it
+        auto const channelPath = makeChannelDataFilePath(flowDir);
+        if (!std::filesystem::exists(channelPath))
```
KimonHoffmann: Prefer ADL to a fully qualified call here.

lib/src/internal/FlowManager.cpp
```diff
@@ -282,7 +294,18 @@ namespace mxl::lib
...
+        {
+            throw std::filesystem::filesystem_error{
+                "Channel buffer file not found.", channelPath, std::make_error_code(std::errc::no_such_file_or_directory)};
+        }
```
KimonHoffmann: More generally I think this entire check is superfluous as I'm pretty sure the call to `openChannelBuffers()` below will raise an error if the file is absent.

So this check is ultimately only covering a subset of the potential failure conditions (wrong type of dirent would be another, for example) and on top of that is prone to races.

For these reasons I'd suggest removing the check entirely.

lib/src/internal/FlowManager.cpp
```diff
@@ -298,7 +321,21 @@ namespace mxl::lib
...
+            catch (std::exception const& e)
+            {
+                MXL_ERROR("Failed to delete flow {}: {}", uuids::to_string(id), e.what());
+                return false;
+            }
+            catch (...)
+            {
+                MXL_ERROR("Failed to delete flow {}: unknown error", uuids::to_string(id));
+                return false;
+            }
```
KimonHoffmann: All this exception handling and error reporting is superfluous, because the invoked `deleteFlow()` overload already swallows and reports all exceptions that may occur.

lib/tests/test_flowmanager.cpp
```diff
@@ -205,3 +220,329 @@ TEST_CASE("Flow Manager : Create Audio Flow Structure", "[flow manager]")
...
+    auto const flowId3 = *uuids::uuid::from_string("33333333-3333-3333-3333-333333333333");
+    REQUIRE_THROWS_AS(manager->openFlow(flowId3, AccessMode::READ_ONLY), std::filesystem::filesystem_error);
+
+    //
+    // 8) listFlows skips invalid directories
+    //
+    {
+        // manually drop a bogus folder
+        auto invalidDir = domain / "not-a-valid-uuid.mxl-flow";
+        REQUIRE(std::filesystem::create_directory(invalidDir));
```
KimonHoffmann: Can probably use the unqualified variant of `create_directory()`.

lib/tests/test_flowmanager.cpp
```diff
@@ -205,3 +220,329 @@ TEST_CASE("Flow Manager : Create Audio Flow Structure", "[flow manager]")
...
+    auto const id = *uuids::uuid::from_string("dddddddd-dddd-dddd-dddd-dddddddddddd");
+    auto const def = mxl::tests::readFile("data/v210_flow.json");
+    auto const rate = Rational{50, 1};
+
+    auto f = mgr->createDiscreteFlow(id, def, MXL_DATA_FORMAT_VIDEO, 1, rate, 64);
+    f.reset();
+    REQUIRE(mgr->listFlows().size() == 1);
+
+    // Make domain read-only
+    std::filesystem::permissions(domain,
```
KimonHoffmann: Can probably use the unqualified variant of `permissions()`.

lib/tests/test_flowmanager.cpp
```diff
@@ -205,3 +220,329 @@ TEST_CASE("Flow Manager : Create Audio Flow Structure", "[flow manager]")
...
+        std::filesystem::perms::owner_all | std::filesystem::perms::group_all | std::filesystem::perms::others_all,
+        std::filesystem::perm_options::remove);
+
+    // DeleteFlow should respect filesystem permissions
+    // When domain is read-only, deletion operations should fail gracefully
+    // The method returns false and logs an error instead of throwing
+    REQUIRE(mgr->deleteFlow(id) == false);
+
+    // Restore full permissions for cleanup and verification
+    std::filesystem::permissions(domain, std::filesystem::perms::owner_all, std::filesystem::perm_options::add);
```
KimonHoffmann: As above.

lib/tests/test_flowmanager.cpp
```diff
@@ -205,3 +220,329 @@ TEST_CASE("Flow Manager : Create Audio Flow Structure", "[flow manager]")
...
+    auto const def = mxl::tests::readFile("data/v210_flow.json");
+    auto const rate = Rational{50, 1};
+
+    // CreateFlow should respect filesystem permissions
+    // When domain is unwritable, creation operations should fail
+    // This test verifies proper permission enforcement
+    REQUIRE_THROWS_AS(mgr->createDiscreteFlow(id, def, MXL_DATA_FORMAT_VIDEO, 1, rate, 8), std::filesystem::filesystem_error);
+
+    // restore perms so we can clean up
+    std::filesystem::permissions(domain, std::filesystem::perms::owner_all, std::filesystem::perm_options::add);
```
KimonHoffmann: As above.

KimonHoffmann (changes requested):
Thank you for applying all the previously requested changes so diligently!
Unfortunately I have a few more changes to request, but nothing serious.

lib/tests/test_flowmanager.cpp
moleksy: Addressed in 7a3e08a153fb90136eabe7f158d4e656b24fc327

lib/tests/test_flowmanager.cpp
moleksy: Addressed in 7a3e08a153fb90136eabe7f158d4e656b24fc327

lib/tests/test_flowmanager.cpp
moleksy: Addressed in 7a3e08a153fb90136eabe7f158d4e656b24fc327

lib/tests/test_flowmanager.cpp
moleksy: Addressed in 7a3e08a153fb90136eabe7f158d4e656b24fc327

lib/src/internal/FlowManager.cpp
moleksy: Addressed in 7a3e08a153fb90136eabe7f158d4e656b24fc327

lib/src/internal/FlowManager.cpp
moleksy: Addressed in 7a3e08a153fb90136eabe7f158d4e656b24fc327

lib/src/internal/FlowManager.cpp
moleksy: Addressed in 7a3e08a153fb90136eabe7f158d4e656b24fc327

lib/src/internal/FlowManager.cpp
moleksy: Addressed in 7a3e08a153fb90136eabe7f158d4e656b24fc327

lib/src/internal/FlowManager.cpp
moleksy: Addressed in 7a3e08a153fb90136eabe7f158d4e656b24fc327

moleksy:
@KimonHoffmann could you review the changes? Thank you!

lib/src/internal/FlowManager.cpp
KimonHoffmann: You also removed the exception handling from the `deleteFlow()` overload it was delegating to leading to the situation where exceptions can now escape from both overloads.

I liked the original intention of catching all exception in these specific methods, I was just referring to the exception handling being superfluous in the one overload that ultimately delegated to the other, because that one was already taking care of the exceptions that may be thrown.

lib/src/internal/FlowManager.cpp
```diff
@@ -308,8 +321,15 @@ namespace mxl::lib
         auto uuid = uuids::to_string(flowId);
         MXL_TRACE("Delete flow: {}", uuid);
 
-        auto const base = std::filesystem::path{_mxlDomain};
-        return (remove_all(makeFlowDirectoryName(base, uuid)) != 0);
+        // Compute the flow directory path
+        auto const flowPath = makeFlowDirectoryName(_mxlDomain, uuid);
+        auto const removed = std::filesystem::remove_all(flowPath);
```
KimonHoffmann: Can be an unqualified call. Please also see my comment regarding the exception handling in the other thread.

lib/tests/test_flowmanager.cpp
```diff
@@ -205,3 +220,329 @@ TEST_CASE("Flow Manager : Create Audio Flow Structure", "[flow manager]")
...
     REQUIRE(!exists(flowDirectory));
 }
+
+TEST_CASE("Flow Manager : Open, List, and Error Conditions", "[flow manager]")
+{
+    auto const domain = mxl::tests::getDomainPath();
+    // start clean
+    auto ec = std::error_code{};
+    remove_all(domain, ec);
+    REQUIRE(std::filesystem::create_directory(domain));
```
KimonHoffmann: Can be unqualified.

lib/tests/test_flowmanager.cpp
```diff
@@ -205,3 +220,329 @@ TEST_CASE("Flow Manager : Create Audio Flow Structure", "[flow manager]")
...
+    // Create and publish
+    auto flow = mgr->createDiscreteFlow(id, def, MXL_DATA_FORMAT_VIDEO, 1, rate, 64);
+    flow.reset();
+    REQUIRE(mgr->listFlows().size() == 1);
+
+    // Manually remove the descriptor JSON to simulate corruption
+    auto flowDir = makeFlowDirectoryName(domain, uuids::to_string(id));
+    auto descFile = makeFlowDescriptorFilePath(flowDir);
+    REQUIRE(std::filesystem::exists(descFile));
+    std::filesystem::remove(descFile);
```
KimonHoffmann: Can both be unqualified.

lib/tests/test_flowmanager.cpp
```diff
@@ -205,3 +220,329 @@ TEST_CASE("Flow Manager : Create Audio Flow Structure", "[flow manager]")
...
+
+    // Make domain read-only
+    permissions(domain,
+        std::filesystem::perms::owner_all | std::filesystem::perms::group_all | std::filesystem::perms::others_all,
+        std::filesystem::perm_options::remove);
+
+    // DeleteFlow should respect filesystem permissions
+    // When domain is read-only, deletion operations should throw filesystem_error
+    // This test verifies proper permission enforcement
+    REQUIRE_THROWS_AS(mgr->deleteFlow(id), std::filesystem::filesystem_error);
```
KimonHoffmann: Once exception handling is reintroduced to the primary overload this test probably needs to change?!

felixpou:
### TSC 2025-08-07
- this is following its review process

moleksy:
@KimonHoffmann I will address it on 08/18-19/2025 as I'm on vacation right now. Thank you!

lib/src/internal/FlowManager.cpp
moleksy: fixed in ca238872b90f3ec1f89cd6cda08023b2cf97f74a

lib/tests/test_flowmanager.cpp
moleksy: fixed in ca238872b90f3ec1f89cd6cda08023b2cf97f74a

lib/tests/test_flowmanager.cpp
moleksy: fixed in ca238872b90f3ec1f89cd6cda08023b2cf97f74a

lib/tests/test_flowmanager.cpp
moleksy: Actually, this test should remain valid even with the current implementation. We're not adding exception handling back to the primary overload, we're removing artificial exception suppression. The filesystem operation itself still throws when it encounters permission errors, and we now let that exception flow through naturally rather than catching and logging it. I believe this aligns with the "either log OR throw, not both" philosophy.

KimonHoffmann: Yes, but what I meant with my change request was something slightly different.

Before:
1. Overload 1
   1. Performs actual operation
   2. Swallows all exceptions
   3. Returns false on any error encountered
2. Overload 2
   1. Delegates to overload 1  
   2. Swallows all exceptions
   3. Returns false on any error encountered

Requested change:
Get rid of 2.2 & 2.3, because there is no way 2.1 can return with an exception due to 1.2 & 1.3.

I still agree that no additional logging should happen, but I liked the fact of your original change that the function effectively became `noexcept` and just indicated success via its return value. The way it is now a caller still needs to handle both error cases (false return **and** exception), which is a bit unwieldy.

moleksy: @KimonHoffmann I think 59a3eb72ae901a6a2cefce592fb7ac01636ccb36 is now in line with what you meant.

PR #52, opened 2025-07-24
Users/lapointejp/use optional gpu ub24 devcontainer

lapointejp:
Changed devcontainers to only use GPU if it is detected.


KimonHoffmann (approved):
Looks good to me, thank you!

Issue #67, opened 2025-08-05
The mxl-gst-videotestsrc does not handle delayed and / or non-linearized data


pac-work:
When we have data which are not written to MXL in nicely linearized fashion and / or have some delay, the test playback application does not handle that well. Example of such a data is an output from a decoder.

To fix this, following changes are proposed:
- Allow read timeout customization - compensates for the delayed data availability.
- Timestamp the buffers pushed to Gstreamer - compensates for the non-linearly spaced frames availability.

PR will follow...

pac-work:
Fixed by https://github.com/dmf-mxl/mxl/pull/68.

PR #68, opened 2025-08-05
Improve mxl-gst-videosink

pac-work:
Solves #67.

### DIFF
--- a/tools/mxl-gst/videosink.cpp
+++ b/tools/mxl-gst/videosink.cpp
@@ -171,12 +171,14 @@ public:
         gst_element_set_state(_pipeline, GST_STATE_PLAYING);
     }
 
-    void pushSample(GstBuffer* gst_buffer, uint8_t* payload, size_t payload_len)
+    void pushSample(uint8_t* payload, size_t payload_len)
     {
+        GstBuffer* gst_buffer{gst_buffer_new_allocate(nullptr, payload_len, nullptr)};
+
         GstMapInfo map;
         gst_buffer_map(gst_buffer, &map, GST_MAP_WRITE);
-
         ::memcpy(map.data, payload, payload_len);
+        gst_buffer_unmap(gst_buffer, &map);
 
         int ret;
         g_signal_emit_by_name(_appsrc, "push-buffer", gst_buffer, &ret);
@@ -185,7 +187,8 @@ public:
             MXL_ERROR("Error pushing buffer to appsrc");
             return;
         }
-        gst_buffer_unmap(gst_buffer, &map);
+
+        gst_buffer_unref(gst_buffer);
     }
 
 private:
@@ -236,7 +239,6 @@ int real_main(int argc, char** argv, void*)
         .frame_rate = descriptor_parser.getGrainRate()};
 
     GstreamerPipeline gst_pipeline(gst_config);
-    GstBuffer* gst_buffer{nullptr};
 
     int exit_status = EXIT_SUCCESS;
     mxlStatus ret;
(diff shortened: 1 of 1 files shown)


tools/mxl-gst/videosink.cpp
```diff
@@ -194,6 +205,11 @@ class GstreamerPipeline final
     GstElement* _videoscale{nullptr};
     GstElement* _autovideosink{nullptr};
     GstElement* _pipeline{nullptr};
+    /// Indicates the MXL timestamp at the moment the pipeline got started. Is used to calculate the PTS of the buffers based on the pipeline's
+    /// running time.
+    std::uint64_t _baseTimestamp{0};
+    /// Indicates the latency at which we are getting the data. We adjust PTS values based on this information to ensure fluent playback.
+    std::uint64_t _highestLatency{0};
```
lapointejp: What's the unit of the latency ? Perhaps clarify in the variable name ?

tools/mxl-gst/videosink.cpp
```diff
@@ -212,6 +228,11 @@ int real_main(int argc, char** argv, void*)
     domainOpt->required(true);
     domainOpt->check(CLI::ExistingDirectory);
 
+    std::optional<std::uint64_t> readTimeout;
```
lapointejp: I would append Ns at the end of variable name to clarify. Also, does ns really make sense for a user supplied argument ? I would argue that microseconds are probably enough for anything a user would want to do ?

tools/mxl-gst/videosink.cpp
pac-work: I have appended Ns.

tools/mxl-gst/videosink.cpp
pac-work: I did append the Ns and kept the value in ns.

Agree that it may be strange, but I wanted to make this value consistent with the value `urn:x-mxl:option:history_duration/v1.0` user needs to enter in `<mxl_domain>/.options`. That one is supposed to be in ns and user most likely set it when they need to change the read timeout. I can change the unit here to Ms, if you want to, but I preferred consistency of the user facing parameters.

Also, @lapointejp , could you add some other reviewer, so we get this in? My [other changes](https://github.com/dmf-mxl/mxl/pull/79) which allow "hack synchronized" audio - video playback (audio has to be available before video) are dependent on this...

tools/mxl-gst/videosink.cpp
```diff
@@ -165,18 +165,28 @@ class GstreamerPipeline final
...
     }
 
-    void pushSample(GstBuffer* gst_buffer, uint8_t* payload, size_t payload_len)
+    void pushSample(uint8_t* payload, size_t payload_len, std::uint64_t timestamp, std::uint64_t currentTime)
     {
+        auto const latencyNs = currentTime > timestamp ? currentTime - timestamp : 0;
+        if (latencyNs > _highestLatencyNs)
+        {
+            _highestLatencyNs = latencyNs;
+            MXL_WARN("Latency increase detected: {} ns", latencyNs);
```
lapointejp: Haven't tested myself, but it seems to me like this could be trigger happy at first ? Maybe we need to warn only if the jump of a significant amount ?
Does it not always warn on going from zero to non-zero as well ?

pac-work: Yes, you are totally right - it always prints like 4 or 5 lines at the beginning... and sometimes even later. On the other hand, it kind of feels like a useful information. Each of those printouts will mean glitch in the playback. Maybe making it just `MXL_INFO`?

lapointejp: Yes, info would be a good start in my opinion. I still find it a little spammy that we log 4-5 times early on while it stabilizes, but I can live with that.

pac-work: Changed this to info.

KimonHoffmann (approved):
Looks good to me, thank you!

lapointejp (approved):

PR #78, opened 2025-08-08
Add inter-host memory sharing implementation

jonasohland:
Hi everyone!

My colleague @mlefebvre1 and I are happy to present to you this draft of an inter-host memory sharing extension to mxl using RDMA. 

This is based on previous work that we did on an internal project, but is a complete rewrite with a completely different design,  and contains no code related our internal project. Currently supported is sharing of complete grains of discrete flows using hardware that can be addressed by userspace verbs and also the libfabric tcp provider for debugging. AWS EFA is not supported right now, but we plan to add support for it during the coming week.

For now, the extension is built into a separate shared object `libmxl-fabrics.so` that will be built in addition to the mxl base library when the project is configured with `-DMXL_FABRICS_OFI=ON`. This will build the extension and a demo application in `tools/mxl-fabrics-demo` that can be used to transfer a single flow from one node to another. 

We expect this PR draft to be a starting point for a discussion around the topic.
We already see some issues that we would like to include in this draft and we hope the community can give us some helpful feedback on this:

#### 1. Grain file access and memory protection flags
If you take a closer look at the changes we have done to the shared memory abstraction in mxl, you will be shocked and horrified to discover that we changed the code so the grain files will always be opened with `O_WRITE`, and their contents mmap'd with `PROT_READ | PROT_WRITE` regardless if we open the flow for reading or writing. The reason for that is that `PROT_WRITE` is required for memory registration with the verbs api, even if the registration is done as a source for the initiator of a remote write operation. And for `PROT_WRITE` the file needs to be opened with `O_WRITE`. We don't know exactly why this is required, but my guess is that it has to do with changes that are made to flags on the backing memory pages at the time of registration. We are investigating if there is anything that can be done about this, but at the time of writing this PR, we have not found a workaround.
We know this change cannot be accepted and asking for your advice on this. Our wish was to use the mxl `FlowWriter` and `FlowReader` classes to access the underlying memory so we don't have to implement the logic behind it a second time, but if that's what we have to do, then that is what we have to do. But even then it would be not very nice for the user of a program that effectively only wants to do a read and send of the contents of some file to need write permission for that file.

#### 2. Distribution of the extension, inclusion into base library
Right now the implementation behind the symbols in `fabrics.h` are built into a separate shared object `libmxl-fabrics.so`. But that's just what was the easiest for us during development. We are open to any other format. 
This could also be implemented in the base library with the downside of mxl then having a link-time dependency on libfabric.
Another option would be an entirely separate project hosted in a separate repository that links to upstream mxl.
There was also an idea floating around that there should not be a library at all, but instead just a service to share flows on the network that would be maintained by the community.

#### 3. Continuous flow (audio) support
There is no support for continuous flows in mxl-fabrics at the moment. This was out of scope for us for this first draft. I am going to be honest and tell you that it was not immediately obvious to me how to use the continuous flow API when I first looked at it. So we would probably need some support from a knowledgeable person to make progress on this. 

#### 4. Additional libraries in `vcpkg.json`
You will see that we have pulled some additional dependencies into vcpkg.json. We realize there is already a json library used in mxl. But we like the one we have used here more and it is very frustration-free and easy to use. We would like to encourage you to consider adopting this library instead of the one that is used in mxl right now. If you are open to this, we would create a separate issue for that. Otherwise we will change our code to use the library that is already used in mxl.
We also need a library for base64 encoding. For now we have included one directly in the project in Base64.hpp. But I see some clang-tidy warnings in this one, and we are probably violating the original projects license, so I will remove that header and include the library through vcpkg.

#### 5. Kubernetes/Container testing
I have been developing in a VM on Proxmox and @mlefebvre1 was working on a bare-metal server. So we have not yet tested this inside a container and on Kubernetes. But we have previous experience running libfabric based apps on Kubernetes (kubeadm-based with the sriov-network-operator, openshift and also Nvidias Holoscan for Media), so if anyone wants to try this, feel free to contact me.

#### 6. Lacking documentation of internal classes/functions
We need to improve both on the documentation in `fabrics.h` and the headers in `lib/fabrics/src/internal`. I will be working on this next week, but if there is anything unclear to you in the code right now, don't hold back with the comments.

#### 7. Error reporting
We have extended the status enum in `mxl.h` to include some additional error codes that we thought we needed. But these also include some that would probably be of use in the base library as alternatives to `MXL_ERR_UNKNOWN`.  We have also defined our own Exception class that can be thrown together with a status code that will be returned at the api surface when it is caught there.

#### 8. Rust bindings
We like rust and want to contribute bindings to this library also. We also think that a possible reference application for inter-host sharing should be written in rust. 
By the way: Both @mlefebvre and I have been mainly writing rust in the past years, so our C++ might have been little bit rusty and maybe  resembles a more data-oriented approach than the classic OOP and Clean Code(TM) style. 

#### 9. Connection management and control
Connection and management is out of scope for this project in our opinion. Right now, the connection establishment flow is very simple. The *Target* is created and returns a `TargetInfo` object that can be serialized using a function in the library. This object contains the *Targets* implementation-specific fabric address, information about addressable memory regions and an access key. It can then be passed to the *Initiator*. The *Initiator* will then establish a connection to the *Target* and begin transferring grains. We think this is simple enough for someone to build a robust connection management on top of it. It might be Kubernetes CRDs, a coordination server or something that works peer-to-peer, we think its out of scope for this library.

---

We know this is a very large PR, and maybe we have to split it into multiple parts, even though I would rather not do it. But to get you started right now with the code, here is a quick intro on what can be found where:

`lib/include/mxl/fabrics.h` The main header that contains the complete user facing API. 
`tools/mxl-fabrics-demo/demo.cpp` Uses this API to implement a simple network transfer application. It is good source to learn how the library should be used.

`lib/fabrics/src/fabrics.cpp` Contains the middle layer between the C api and the C++ code the implements the bulk of the funcionality.
`lib/fabrics/src/internal/*` is where all the implementation lives.

Some file in there that might be interesting to you:
`lib/fabrics/src/internal/RCTarget.{h,c}pp` implements the target using reliable connection oriented queue pairs. 
`lib/fabrics/src/internal/RCInitiator.{h,c}pp` implements the initiator using reliable connection oriented queue pairs.
`lib/fabrics/src/internal/Completion*` code around the completion queue and completions.
`lib/fabrics/src/internal/Event*` code around the event queue and its events.
`lib/fabrics/src/internal/{Passive,}Endpoint.{h,c}pp` abstracts the libfabric endpoint primitive.

We are looking forward to your comments!


.devcontainer/Dockerfile
KimonHoffmann, line 1: Similar updates should be applied to the other Docker files.

CMakeLists.txt
```diff
@@ -27,6 +27,8 @@ project(mxl
     LANGUAGES CXX C
 )
 
+option(MXL_FABRICS "Enable the mxl-fabrics module" FALSE)
```
KimonHoffmann: Could maybe be called `MXL_ENABLE_FABRICS`?!

lib/CMakeLists.txt
```diff
@@ -27,9 +32,19 @@ set_target_properties(mxl
             VERSION                      "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}"
             SOVERSION                    "${PROJECT_VERSION_MAJOR}"
     )
+set_target_properties(mxl-static
+        PROPERTIES
+            POSITION_INDEPENDENT_CODE    ON
+            VISIBILITY_INLINES_HIDDEN    OFF
+            C_VISIBILITY_PRESET          default
+            CXX_VISIBILITY_PRESET        default
```
KimonHoffmann: If I'm not mistaken visibility settings have no effect for static libraries therefor it might be best to just omit all visibility related properties.

lib/CMakeLists.txt
```diff
@@ -78,17 +96,31 @@ target_include_directories(mxl PUBLIC
...
+        PRIVATE
+            stduuid
+            spdlog::spdlog
+            fmt::fmt
+            picojson::picojson
+    )
 
 # Alias trace to libtrace::libtrace so that this library can be used
 # in lieu of a module from the local source tree
 add_library(${PROJECT_NAME}::mxl ALIAS mxl)
```
KimonHoffmann, line 83: It probably makes sense to also add an alias for `mxl-static`.

lib/CMakeLists.txt
```diff
@@ -78,17 +96,31 @@ target_include_directories(mxl PUBLIC
...
+
+add_subdirectory(fabrics)
```
KimonHoffmann: Adding this sub-directory could/should be predicated on `MXL_ENABLE_FABRICS`.

lib/fabrics/ofi/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+            CXX_VISIBILITY_PRESET        hidden
+            C_EXTENSIONS                 OFF
+            CXX_EXTENSIONS               OFF
+            VERSION                      "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}"
+            SOVERSION                    "${PROJECT_VERSION_MAJOR}"
+    )
+
+target_link_libraries(mxl-fabrics
+    PUBLIC
+        mxl-static
```
KimonHoffmann: Why not link against the dynamic library here?

lib/CMakeLists.txt
```diff
@@ -54,8 +69,11 @@ target_sources(mxl
             src/internal/Thread.cpp
             src/internal/Time.cpp
             src/internal/Timing.cpp
-        )
 
+)
+
+target_sources(mxl PRIVATE ${MXL_SOURCES})
+target_sources(mxl-static PRIVATE ${MXL_SOURCES})
```
KimonHoffmann: The two variants can probably be handled more cleanly by referring both to one common cmake "object library".

lib/fabrics/ofi/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+)
+
+target_include_directories(mxl-fabrics
+    PUBLIC
+        "${CMAKE_CURRENT_SOURCE_DIR}/../../src"
+)
+
+target_include_directories(mxl-fabrics-static
+    PUBLIC
+        "${CMAKE_CURRENT_SOURCE_DIR}/../../src"
```
KimonHoffmann: These two should not be `PUBLIC`, I suppose.

Unrelated to this PR in particular, but worth mentioning:
In general we should create a separate (object) library of the internal mxl C++ library and have the tests, as well as this library officially link against that one instead of each of them grabbing files out of the library source directory.

**Edit:** I saw below that you already did that as `mxl-internal`. That's ag great start, but we should flesh that out a little more so that fs level accesses like this one are no longer necessary.

lib/fabrics/ofi/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+        src/internal/Target.cpp
+        src/internal/RCTarget.cpp
+        src/internal/RCInitiator.cpp
+        src/internal/RDMTarget.cpp
+        src/internal/RDMInitiator.cpp
+        src/internal/TargetInfo.cpp
+)
+
+target_sources(mxl-fabrics PRIVATE ${mxl_fabrics_sources})
+target_sources(mxl-fabrics-static PRIVATE ${mxl_fabrics_sources})
```
KimonHoffmann: As remarked above this is probably best handled by creating a common object library.

tools/mxl-fabrics-demo/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+find_package(stduuid CONFIG REQUIRED)
+find_package(spdlog CONFIG REQUIRED)
+
+# Add the test executable
+add_executable(mxl-fabrics-demo
+    demo.cpp
+)
+
+target_link_libraries(mxl-fabrics-demo PRIVATE
+    mxl
```
KimonHoffmann: Probably not necessary, when mxl is a `PUBLIC` link library of `mcx-fabrics`.

lib/tests/fabrics/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+    )
+
+target_sources(mxl-fabrics-tests
+    PRIVATE
+        test_basics.cpp
+)
+
+target_link_libraries(mxl-fabrics-tests
+    PRIVATE
+        mxl
```
KimonHoffmann: Should maybe be `mxl-static` and is probably not necessary, if `mxl-static` is a `PUBLIC` dependency of `mxl-fabrics-static`.

lib/internal/include/mxl-internal/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
...
+target_compile_features(mxl-internal
+        PRIVATE
+            cxx_std_20
+    )
+set_target_properties(mxl-internal
+        PROPERTIES
+            POSITION_INDEPENDENT_CODE    ON
+            VISIBILITY_INLINES_HIDDEN    OFF
+            C_VISIBILITY_PRESET          default
+            CXX_VISIBILITY_PRESET        default
```
KimonHoffmann: Can/should be left out, as mentioned above.

lib/internal/include/mxl-internal/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
...
+
+if (NOT TARGET picojson::picojson)
+    find_package(picojson REQUIRED)
+endif()
+
+target_include_directories(mxl-internal PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>
+    $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
+)
```
KimonHoffmann: This currently makes no sense, as headers are not separated for the internal library and the mentioned directory does not exist, but I'm all in favor of actually promoting the internal parts into a fully fledged (object) library, with proper file layout.

lib/fabrics/ofi/src/internal/Region.hpp
```diff
@@ -0,0 +1,87 @@
...
+    class Region
+    {
+    public:
+        explicit Region(std::uintptr_t base, size_t size)
+            : base(base)
+            , size(size)
+            , _iovec(iovecFromRegion(base, size))
+        {}
+
+        virtual ~Region() = default;
```
KimonHoffmann: Why have a `virtual` destructor for a monomorphic type?

lib/fabrics/ofi/src/internal/Region.hpp
```diff
@@ -0,0 +1,87 @@
...
+#include <bits/types/struct_iovec.h>
+#include "internal/FlowData.hpp"
+#include "mxl/fabrics.h"
+
+namespace mxl::lib::fabrics::ofi
+{
+    class Region
+    {
+    public:
+        explicit Region(std::uintptr_t base, size_t size)
```
KimonHoffmann: Could  probably `constexpr` and should be `noexcept`.

lib/fabrics/ofi/src/internal/Region.hpp
```diff
@@ -0,0 +1,87 @@
...
+        size_t size;
+
+        [[nodiscard]]
+        ::iovec const* as_iovec() const noexcept;
+
+        [[nodiscard]]
+        ::iovec to_iovec() const noexcept;
+
+    private:
+        static ::iovec iovecFromRegion(std::uintptr_t, size_t) noexcept;
```
KimonHoffmann: Can probably be `constexpr`.
`size_t` should be fully qualified.

lib/fabrics/ofi/src/internal/Region.hpp
```diff
@@ -0,0 +1,87 @@
...
+        explicit Region(std::uintptr_t base, size_t size)
+            : base(base)
+            , size(size)
+            , _iovec(iovecFromRegion(base, size))
+        {}
+
+        virtual ~Region() = default;
+
+        std::uintptr_t base;
+        size_t size;
```
KimonHoffmann: Should be fully qualified.

lib/fabrics/ofi/src/internal/Region.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+        std::vector<Region> _inner;
+
+        std::vector<::iovec> _iovecs;
+    };
+
+    class RegionGroups
+    {
+    public:
+        static RegionGroups fromFlow(FlowData& flow);
```
KimonHoffmann: Baking a dependency to `FlowData` into this rather generic type seems a bit wild to me. This could be a free function in a separate, dedicated header.

lib/fabrics/ofi/src/internal/Region.hpp
```diff
@@ -0,0 +1,87 @@
...
+    public:
+        explicit RegionGroup() = default;
+
+        explicit RegionGroup(std::vector<Region> inner)
+            : _inner(std::move(inner))
+            , _iovecs(iovecsFromGroup(_inner))
+        {}
+
+        [[nodiscard]]
+        std::vector<Region> const& view() const noexcept;
```
KimonHoffmann: Could make good use of ranges instead of exposing the internal vector.

lib/fabrics/ofi/src/internal/Region.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+        [[nodiscard]]
+        ::iovec const* as_iovec() const noexcept;
```
KimonHoffmann: Could make good use of ranges to communicate both the start and the end pointer as one.

lib/fabrics/ofi/src/internal/Region.hpp
```diff
@@ -0,0 +1,87 @@
...
+    {
+    public:
+        static RegionGroups fromFlow(FlowData& flow);
+
+        static RegionGroups* fromAPI(mxlRegions) noexcept;
+        [[nodiscard]]
+        mxlRegions toAPI() noexcept;
+
+        [[nodiscard]]
+        std::vector<RegionGroup> const& view() const noexcept;
```
KimonHoffmann: Can make use of ranges.

lib/fabrics/ofi/src/fabrics.cpp
```diff
@@ -0,0 +1,853 @@
...
+#include "internal/Initiator.hpp"
+#include "internal/Provider.hpp"
+#include "internal/Region.hpp"
+#include "internal/Target.hpp"
+#include "internal/TargetInfo.hpp"
+#include "mxl/flow.h"
+#include "mxl/platform.h"
+
+namespace
+{
```
KimonHoffmann, line 32: I don't see the purpose of this anonymous namespace and I think it should be removed.

lib/fabrics/ofi/src/internal/Address.hpp
```diff
@@ -0,0 +1,88 @@
...
+        void const* raw() const noexcept;
+
+        /**
+         * Byte-length of the raw address data.
+         */
+        [[nodiscard]]
+        std::size_t size() const noexcept;
+
+    private:
+        explicit FabricAddress(std::vector<uint8_t> addr);
```
KimonHoffmann: `std::uin8_t`

lib/fabrics/ofi/src/internal/Address.cpp
KimonHoffmann, line 1: Some instances of `uint8_t` and `size_t` miss `std::` qualification here.

lib/src/internal/SharedMemory.cpp
KimonHoffmann, line 1: As long as this can not be avoided, both changes in this file should have explanatory comments inline.

lib/include/mxl/fabrics.h
```diff
@@ -0,0 +1,305 @@
...
+    } mxlFabricsProvider;
+
+    /// Address of a logical network endpoint. This is analogous to a hostname and port number in classic ipv4 networking.
+    /// The actual values for node and service vary between providers, but often an ip address as the node value and a port number as the service
+    /// value are sufficient.
+    typedef struct mxlEndpointAddress_t
+    {
+        char const* node;
+        char const* service;
+    } mxlEndpointAddress;
```
KimonHoffmann, line 57: The lifetime requirements and ownership of these pointer is probably worth a mention in the description.

lib/fabrics/ofi/src/internal/LocalRegion.hpp
```diff
@@ -0,0 +1,55 @@
...
+    class RemoteRegionGroup;
+
+    struct LocalRegion
+    {
+        uint64_t addr;
+        size_t len;
+        void* desc;
+
+        [[nodiscard]]
+        ::iovec toIov() const noexcept;
```
KimonHoffmann: Should probably use the same naming convention as `Region`.

lib/fabrics/ofi/src/internal/LocalRegion.hpp
```diff
@@ -0,0 +1,55 @@
...
+            , _descs(descFromGroup(_inner))
+        {}
+
+        [[nodiscard]]
+        std::vector<LocalRegion> const& view() const noexcept;
+
+        [[nodiscard]]
+        iovec const* iovec() const noexcept;
+        [[nodiscard]]
+        size_t count() const noexcept;
```
KimonHoffmann: Same comments about using ranges as for `Region`.

KimonHoffmann (commented):
Publishing this partial review as **comments** before I leave for two weeks.
This is based on reviewing roughly 50% of the files.

A few general syntactic remarks that I stopped commenting on after the first one or two instances:
* The fixed width integer types as well as `size_t`,`ptrdiff_t` and the likes should be qualified with the `std` namespace.
* In some cases it would be more readable to use right hand init, i.e `auto x = Type{val}` instead of `Type x{val}`.
* I like ordering the cisibility specifiers from less to more restrictive (i.e. `public`, `protected`, `private` -- all fine as far as I could see) and separating type definitions, methods and members of a class by repeating the respective visibility specifier, i.e.
```cpp
class C
{
    // ...
    public:
        using TypeA = ...;
        using TypeB = ...;
     
    public:
        void methodA(...);
        
    private:
        void helperMethod();
         
    private:
        int _member;
};
```

And a general question regarding the use of reflcpp: Isn't this way of serializing target infos prone to compatibility problems down the line? Wouldn't it be less heavy weight and more resilient to define our own custom serialization format for those?

Thanks a lot for the time, effort and care poured into this high quality addition to the library!

mlefebvre1:
**Progress Update: 1. Grain file access and memory protection flags**

We identified and resolved the issue with memory region registration. The problem stemmed from libfabric's default memory registration cache, which was overriding our specified memory access permissions.

When using the verbs provider with caching enabled, memory region registration flows through `verb_mr_cache_add_region`, which calls `vrb_mr_reg_common`. This function then invokes ibv_reg_mr with given access flags: IBV_ACCESS_LOCAL_WRITE, IBV_ACCESS_REMOTE_WRITE, IBV_ACCESS_REMOTE_ATOMIC, and IBV_ACCESS_REMOTE_READ. This forced requirement for both read and write permissions conflicted with our intended memory protection settings which only requires read permissions.

The solution is to disable the memory registration cache entirely. Since we only perform memory registration during initialization rather than runtime, the cache provides no benefit for our use case.

mlefebvre1:
> And a general question regarding the use of reflcpp: Isn't this way of serializing target infos prone to compatibility problems down the line? Wouldn't it be less heavy weight and more resilient to define our own custom serialization format for those?

I'm not sure how we can get compatibility problems by using this library, because we are explicitly naming fields. For an example, you can check `TargetInfo.hpp`.

lib/src/internal/SharedMemory.cpp
mlefebvre1: This was resolved and is not needed anymore.

lib/fabrics/ofi/src/internal/Region.hpp
mlefebvre1: The implementation casts the `std::uintptr_t` to `void*`. As far as I know this is forbidden in constexpr context.

lib/fabrics/ofi/src/internal/Region.hpp
mlefebvre1: Makes use of iovecFromRegion which can't be made constexpr

lib/fabrics/ofi/src/internal/Region.hpp
mlefebvre1: moved to a seperate function : `regionGroupsfromFlow`

lib/fabrics/ofi/src/internal/Region.hpp
mlefebvre1: implemented ranges for all the classes that had a `view` method in Region.hpp, LocalRegion.hpp and RemoteRegion.hpp

lib/fabrics/ofi/src/fabrics.cpp
mlefebvre1: removed

lib/fabrics/ofi/src/internal/Region.hpp
KimonHoffmann: `std::bitcast`?

PR #79, opened 2025-08-11
Synchronized audio and video playback in gst-videosink

pac-work:
Only the last commit is relevant here, remaining ones are already in different PRs.

There is still an audio glitch in the moment the audio buffer rolls over (when the second buffer is of non-zero size). This is currently being investigated. It is unsure, whether the error is here, in the MXL itself, or somewhere else.


KimonHoffmann (approved):
Looks good to me, thank you for contributing this!

sviengkhou (approved):

Issue #81, opened 2025-08-13
Audio buffers corruption


pac-work:
While adding support to the current GStreamer sink ([issue](https://github.com/dmf-mxl/mxl/issues/45), [PR](https://github.com/dmf-mxl/mxl/pull/79)), it has been discovered that audio data get corrupted sometimes.

This appears this happens on the buffer roll-over. The test which demonstrates this problem better is available [here](https://github.com/dmf-mxl/mxl/pull/80).

pac-work:
> It'd be great if you could rebase your PR on this and see whether it makes your newly added test consistently pass.

Will do so tomorrow. Today it is the same for me as for you - taking care of my son :)

PR #86, opened 2025-08-14
Fix audio buffer corruption caused by invalid range return

KimonHoffmann:
Fix issue #81 that was caused by an invalid base pointer being used for the second range fragment.
This PR also introduces a check on the number of samples requested to ensure the entire requested range is valid.


pac-work:
I haven't yet rebased my test to see, but - isn't writer the same story?

lib/src/internal/PosixContinuousFlowReader.cpp
```diff
@@ -30,7 +30,7 @@ namespace mxl::lib
             {
                 auto const minIndex = (headIndex >= (_bufferLength / 2U)) ? (headIndex - (_bufferLength / 2U)) : std::uint64_t{0};
 
-                if (index >= minIndex)
+                if ((index >= minIndex) || ((index - minIndex) > count))
```
garethsb: if `!(index >= minIndex)` then `index - minIndex` is logically negative... except actually there will be underflow because these are unsigned variables... what am I missing?

KimonHoffmann: You're missing that I really should not do such things in a hurry and use `&&` instead of `||`. 😉
Good catch!

KimonHoffmann: I've fixed this in the most recent version of the commits.

lib/src/internal/PosixContinuousFlowReader.cpp
```diff
@@ -30,7 +30,7 @@ namespace mxl::lib
             {
                 auto const minIndex = (headIndex >= (_bufferLength / 2U)) ? (headIndex - (_bufferLength / 2U)) : std::uint64_t{0};
 
-                if (index >= minIndex)
+                if ((index >= minIndex) && ((index - minIndex) > count))
```
garethsb: Would this be appropriate?

```suggestion
                if ((index >= minIndex) && ((index - minIndex) >= count))
```

And maybe same in the writer condition? `(count <= (_bufferLength / 2))` ?

KimonHoffmann: Of course! Sorry for the noise ... I should be way more thorough with my changes.
I updated the commit accordingly.

PR #94, opened 2025-08-15
added absolute index and time stamp to grain

david3a:
absolute time stamp to grain, allows origination time to propagate accurately
added absolute index to grain, allows verification that grain contains expected index
added structure size checks to test cases for flows to ensure all hit page size multiples

### DIFF
--- a/lib/include/mxl/flow.h
+++ b/lib/include/mxl/flow.h
@@ -136,8 +136,12 @@ extern "C"
         /// How many bytes in the grain are currently valid (commited).  This is typically used when writing slices.
         /// A grain is complete when commitedSize == grainSize
         uint32_t commitedSize;
+        /// absolute index of grain last written to this buffer
+        uint64_t grainIndex;
+        /// time stamp of grain last written to this buffer
+        uint64_t grainTimeStamp;
         /// User data space
-        uint8_t userData[4068];
+        uint8_t userData[4044];
     } GrainInfo;
 
     typedef struct mxlFlowReader_t* mxlFlowReader;
--- a/lib/include/mxl/flow.h
+++ b/lib/include/mxl/flow.h
@@ -291,6 +291,18 @@ MXL_EXPORT
     MXL_EXPORT
     mxlStatus mxlFlowReaderGetSamples(mxlFlowReader reader, uint64_t index, size_t count, mxlWrappedMultiBufferSlice* payloadBuffersSlices);
 
+    ///
+    /// retrieve the current range of absolute grain indexes available
+    ///
+    /// \param[in] reader flow reader to access
+    /// \param[out] oldest_index pointer to oldest valid grain
+    /// \param[out] newest_index pointer to newest valid grain
+    ///
+    /// \return A status code describing the outcome of the call.
+    ///
+    MXL_EXPORT
+    mxlStatus mxlFlowReaderGetGrainRange(mxlFlowReader reader, std::uint64_t *oldest_index, std::uint64_t *newest_index);
+
     /**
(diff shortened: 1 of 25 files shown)


lib/include/mxl/flow.h
```diff
@@ -136,8 +136,12 @@ extern "C"
         /// How many bytes in the grain are currently valid (commited).  This is typically used when writing slices.
         /// A grain is complete when commitedSize == grainSize
         uint32_t commitedSize;
+        /// absolute index of grain last written to this buffer
+        uint64_t grainIndex;
+        /// time stamp of grain last written to this buffer
+        uint64_t grainTimeStamp;
         /// User data space
-        uint8_t userData[4068];
+        uint8_t userData[4044];
```
KimonHoffmann, line 144: In order to align the true size of the struct without tail padding to 4K this needs to be 4048.

The actually size of the fields would leave even more space (4052 bytes) if it weren't for 4 bytes of padding before `grainIndex`. 

I think we are still in a phase where we don't need to maintain backwards compatibility so we could opt for recording these fields in a way that
1. makes sense and
2. avoids padding.

At a quick glance the padding could be avoided while still maintaining sensible grouping by sorting the fields like this:

* version
* size
* grainIndex
* grainTimeStamp
* flags
* payloadLocation
* deviceIndex
* grainSize
* commitedSize

To keep related fields aligned to 8 byte boundaries we could also opt to spend 64 bits on flags instead of 32.

lib/tests/test_flows.cpp
```diff
@@ -206,6 +207,11 @@ TEST_CASE("Data Flow : Create/Destroy", "[mxl flows]")
     uint64_t index = mxlTimestampToIndex(&rate, now);
     REQUIRE(index != MXL_UNDEFINED_INDEX);
 
+    /// validate grain and flow structure sizes
+    REQUIRE(sizeof(GrainInfo) == 0x1000);
+    REQUIRE(sizeof(FlowInfo) == 0x1000);
+    REQUIRE(sizeof(mxl::lib::Grain) == 0x2000);
+
```
KimonHoffmann, line 215: Checking these is a great idea, but in order to avoid pitfalls as the one above we should probably also check for `std::has_unique_object_representations_v<...>` of each type.

There are non-padding related reasons why this trait might not be fulfilled, but as long as those are not a factor we can put it to good use here.

KimonHoffmann (changes requested):
Thank you for this contribution. I only have a few small remarks.

lib/include/mxl/flow.h
david3a: checking the sizeof in the test framework has led me to wonder if compile flags are consistent, the test module reports the structure as 4096 bytes with 4044 padding.

KimonHoffmann: This is unrelated to compiler settings.

By introducing fields of type `unit64_t` the alignment of the structure has increased from 4 bytes to 8 bytes. With a `userData` field of 4044 bytes the total size of the structure is 4092 bytes (including 4 bytes of padding before `grainIndex`), which is then increased with additional 4 bytes of tail padding to 4096 bytes to satisfy the alignment constraint.

You can try this yourself: If you increase the size of `userData` to 4048 bytes the test will still pass, because you have traded 4 bytes of tail padding with the 4 additional bytes of user data.

This is also why I suggested testing for `std::has_unique_object_representations_v<...>` in addition to the plain size requirements. The way tail padding works this size test alone will succeed for any value of user data bytes in the range [4041, 4048], and thus not quite assert what it is supposed to.

PR #99, opened 2025-08-22
Prefix all C types with mxl.  Fix compilation on macOS (mkdtemp)

vt-tv:


lib/include/mxl/flowinfo.h
```diff
@@ -116,7 +116,7 @@ extern "C"
      * The flow shared memory will be located in {mxlDomain}/{flowId}
      * where {mxlDomain} is a filesystem location available to the application
      */
-    typedef struct FlowInfo
+    typedef struct mxlFlowInfo
```
garethsb: Just noticing that we have a little inconsistency between `typedef struct mxlGrainInfo_t { ... } mxlGrainInfo;` but `typedef struct mxlFlowInfo { ... } mxlFlowInfo;` (no `_t` on the (unused) struct type name).

lib/src/internal/PosixContinuousFlowWriter.cpp
```diff
@@ -62,8 +62,8 @@ namespace mxl::lib
     {
         if (_flowData)
         {
-            auto const flowInfo = _flowData->flowInfo();
-            flowInfo->continuous.headIndex = _currentIndex;
+            auto const mxlFlowInfo = _flowData->flowInfo();
+            mxlFlowInfo->continuous.headIndex = _currentIndex;
```
garethsb: ⬇️ Not a type name, so I'd leave these as they were?

lib/tests/test_flowmanager.cpp
```diff
@@ -9,6 +9,7 @@
 #include <memory>
 #include <string>
 #include <thread>
+#include <unistd.h>
```
garethsb, line 12: ?

lib/include/mxl/flow.h
```diff
@@ -149,10 +149,10 @@ extern "C"
     /// \param[in] instance The mxl instance created using mxlCreateInstance
     /// \param[in] flowDef A flow definition in the NMOS Flow json format.  The flow ID is read from the <id> field of this json object.
     /// \param[in] options Additional options (undefined). \todo Specify and used the additional options.
-    /// \param[out] info A pointer to a FlowInfo structure.  If not nullptr, this structure will be updated with the flow information after the flow
-    /// is created.
+    /// \param[out] info A pointer to a mxlFlowInfo structure.  If not nullptr, this structure will be updated with the flow information after the
```
garethsb: Being a pedant...

```suggestion
    /// \param[out] info A pointer to an mxlFlowInfo structure.  If not nullptr, this structure will be updated with the flow information after the
```

lib/include/mxl/flow.h
```diff
@@ -173,19 +173,20 @@ MXL_EXPORT
      * Get a copy of the current descriptive header of a Flow
      *
      * \param[in] reader A valid flow reader
-     * \param[out] info A valid pointer to a FlowInfo structure. on return, the structure will be updated with a copy of the current flow info value.
+     * \param[out] info A valid pointer to a mxlFlowInfo structure. on return, the structure will be updated with a copy of the current flow info
```
garethsb: Being a pedant...
```suggestion
     * \param[out] info A valid pointer to an mxlFlowInfo structure. on return, the structure will be updated with a copy of the current flow info
```

lib/tests/test_flowmanager.cpp
vt-tv: required for mkdtemp. compilation was broken on darwin.

lib/include/mxl/flowinfo.h
vt-tv: made all structs definition consistant. thanks for picking this up.

lib/src/internal/PosixContinuousFlowWriter.cpp
vt-tv: good catch.  fixed.

KimonHoffmann (approved):
Looks good to me, with just one principal thing in the naming I find bit off.

The usage of `TypeName_t` is exactly opposite to the way it is usually employed, where `_t` is a postfix indicating that it's a `t`ypedef.

```cpp
// Common pattern
typedef struct Name { ... } Name_t;

// Here ...
typedef struct Name_t { ... } Name;

// Alternative I've seen around a few times
typedef strutc Name_s { ... } Name;
```

KimonHoffmann:
I find both `typedef struct Name { ...} Name` and `typedef struct Name_s { ... } Name` to be equally good alternatives with a minor preference for the latter.

garethsb (approved):

PR #101, opened 2025-08-22
add a function to detect if a flow is stale

vt-tv:


lib/src/flow.cpp
```diff
@@ -39,6 +42,59 @@ mxlStatus mxlCreateFlow(mxlInstance instance, char const* flowDef, char const* /
...
+                if ((flowId != nullptr) && uuids::uuid::is_valid_uuid(flowId))
+                {
+                    auto const id = uuids::uuid::from_string(flowId);
+                    if (id.has_value())
+                    {
+                        auto const domain = cppInstance->getDomain();
+
+                        // Try to obtain an exclusive lock on the flow data file.  If we can obtain one it means that no
+                        // other process is writing to the flow.
+                        auto flowDataFile = mxl::lib::makeFlowDataFilePath(domain, flowId);
```
pac-work, line 64: Will this really solve the problem of readers connected to the stale flows? Imagine the following sequence:

1. reader is waiting for data
2. writer stops and deletes the flow
3. writer is restarted and creates new flow with the same id
4. reader wakes up after timeout and calls this function
  a. the function returns that everyting is OK, because at this place it just opens a new lock file
  b. reader still does not know...

Or am I wrong?

sviengkhou: @vt-tv  this function will hold a lock to fd.  and any subsequent attempt to call this function with the exact flow id will fail.  When will fd be released from the exclusive lock?  The release has to be explicitly done no?
Something like 
`flock(fd, LOCK_UN |LOCK_NB)`

lib/src/flow.cpp
```diff
@@ -39,6 +42,58 @@ mxlStatus mxlCreateFlow(mxlInstance instance, char const* flowDef, char const* /
...
+
+                        int fd = open(flowDataFile.c_str(), O_RDONLY | O_CLOEXEC);
+                        if (fd < 0)
+                        {
+                            MXL_ERROR("Failed to open flow data file {} : {}", flowDataFile.string(), std::strerror(errno));
+                            return MXL_ERR_FLOW_NOT_FOUND;
+                        }
+
+                        // Try to obtain an exclusive lock on the file descriptor. Do not block if the lock cannot be obtained.
+                        bool active = flock(fd, LOCK_EX | LOCK_NB) < 0;
```
sviengkhou, line 74: @vt-tv this function will hold a lock to fd. and any subsequent attempt to call this function with the exact flow id will fail. When will fd be released from the exclusive lock? The release has to be explicitly done no?
Something like
flock(fd, LOCK_UN |LOCK_NB)

lib/src/flow.cpp
sviengkhou: TY @vt-tv  for 

`When you call flock(fd, LOCK_EX) or flock(fd, LOCK_SH), the kernel records the lock against the underlying open file description referenced by fd.

If you call close(fd), and that was the last descriptor in your process (or any forked child process) that referred to that open file description, the kernel automatically releases the lock.

If you duplicate the file descriptor (via dup, dup2, fcntl(fd, F_DUPFD, …)), those descriptors share the same open file description — closing only one of them does not release the lock; you need to close all of them for the lock to be released.`

lib/src/flow.cpp
vt-tv: I think the lock is associated to the fd.  man 2 flock:

> Locks created by flock() are associated with an open file description (see open(2)). This means that duplicated file descriptors (created by, for example, fork(2) or dup(2)) refer to the same lock, and this lock may be modified or released using any of these descriptors. Furthermore, the lock is released either by an explicit LOCK_UN operation on any of these duplicate descriptors, or when all such descriptors have been closed.

sviengkhou (approved):

osmarbento-AWS:
looks good! well-implemented feature with good architecture, tests covered and proper API design. clean refactoring removes unused code while adding valuable functionality.

PR #116, opened 2025-08-30
Users/vt tv/115 add file access checks and meaningful error codes when flows cant be accessed or created

vt-tv:


sviengkhou (approved):

pac-work (approved):
Hello, I thought I approved above... not sure where it got lost.

pac-work:
>  > @pac-work I think you now have permissions to approve PRs. please confirm

> Hello, I thought I approved above... not sure where it got lost.

@vt-tv So as it seems, I can approve, but it does not really mean anything. Not sure whether that is useful or not.

lib/src/flow.cpp
```diff
@@ -31,6 +32,21 @@ mxlStatus mxlCreateFlow(mxlInstance instance, char const* flowDef, char const* /
         }
         return MXL_ERR_INVALID_ARG;
     }
+    catch (std::filesystem::filesystem_error const& e)
+    {
+        MXL_ERROR("Failed to create flow : {}", e.what());
+        auto code = e.code();
+        if (code == std::errc::permission_denied || code == std::errc::operation_not_permitted || code == std::errc::read_only_file_system)
```
KimonHoffmann: I'm a big fan to put separate sets of parentheses around binary sub-expressions, as I think it makes the code more readable and less prone to wonky precedence conflicts.

KimonHoffmann (approved):
Looks good to me!
Just a small syntactical remark from my end.

lib/src/flow.cpp
```diff
@@ -31,6 +32,21 @@ mxlStatus mxlCreateFlow(mxlInstance instance, char const* flowDef, char const* /
         }
         return MXL_ERR_INVALID_ARG;
     }
+    catch (std::filesystem::filesystem_error const& e)
+    {
+        MXL_ERROR("Failed to create flow : {}", e.what());
+        auto code = e.code();
```
KimonHoffmann: Can/should be const.

vt-tv: fixed

PR #124, opened 2025-09-04
Test that mxlIsFlowActive detects writer restart

pac-work:
This is a failing test to demonstrate problem with stale readers and how `mxlIsFlowActive` does not solve this problem.


vt-tv:
@pac-work  There is another PR that attempts to resolve this issue.  #135.  It also has a few unit tests to detect this issue.

Issue #128, opened 2025-09-04
`watch` command incompatible with docker compose log output in examples


jptrainor:
The following command at the last line of `examples/Dockerfile.reader.txt` would not produce output when running in a terminal on my Unbuntu 24.04 desktop:

```
CMD ["watch", "/app/mxl-info", "-d", "/domain", "-l"]
```

I tracked that down to incompatiblity of the watch command with the docker compose log ouput that is redirected to the terminal. The watch command sends terminal clear screen characters before each command execution. Somehow that is confusing the docker compose logger and no output results in the terminal.  Note that this may be `watch` implementation, and terminal program dependent therefore it is possible that this did work somewhere. It didn't work for me, however.

An alternative that avoids the `watch` command's clear screen complication is:

```
CMD["sh", "-c", "while true; do /app/mxl-info -d /domain -l || true; sleep 2; done"]
```

See also: related issue 127

vt-tv:
Fixed by MR #131

PR #131, opened 2025-09-05
Fix continuous flow latency reporting in mxl-info

### LINKED ISSUE #121: Grain latency always using discrete head index even in case of continuous buffers.
https://github.com/dmf-mxl/mxl/blob/9519e9e9d363b1856473694c828b8ab8ef099e8c/tools/mxl-info/main.cpp#L69

### LINKED ISSUE #127: examples incorrect reference to /app/mxl-gst-info
In the file `examples/Dockerfile.reader.txt` there is an errant reference to `/app/mxl-gst-info`. It should be `/app/mxl-info`:

Current:
```
CMD ["watch",  "/app/mxl-gst-info", "-d", "/domain", "-l"]
```

Should be:
```
CMD ["watch",  "/app/mxl-info", "-d", "/domain", "-l"]
```

vt-tv:
Fix #121 

- Latency reporting of continuous buffers was using discrete buffer index
- Remove the nanoseconds based latency reporting. it was using the flow mtime which is irrelevant. we may want to actually compute the value if someone cares about it.  reporting the latency as grains or samples.
- Help debug timing issues but displaying latency values in red (out of ringbuffer) yellow (on the edge) or green (in the ringbuffer)

Also fix #127 and #128 (example docker compose reader dockerfile syntax for recent distros)

### DIFF
--- a/tools/mxl-info/main.cpp
+++ b/tools/mxl-info/main.cpp
@@ -5,17 +5,20 @@
 #include <cstdlib>
 #include <ctime>
 #include <filesystem>
+#include <fstream>
 #include <iostream>
 #include <sstream>
 #include <string>
 #include <unistd.h>
 #include <uuid.h>
 #include <CLI/CLI.hpp>
+#include <fmt/color.h>
 #include <fmt/format.h>
 #include <gsl/span>
 #include <mxl/flow.h>
 #include <mxl/mxl.h>
 #include <mxl/time.h>
+#include <picojson/picojson.h>
 #include <sys/file.h>
 #include "../../lib/src/internal/PathUtils.hpp"
 
@@ -47,11 +50,28 @@ std::ostream& operator<<(std::ostream& os, mxlFlowInfo const& info)
        << '\t' << fmt::format("{: >18}: {}", "Format", getFormatString(info.common.format)) << '\n'
        << '\t' << fmt::format("{: >18}: {:0>8x}", "Flags", info.common.flags) << '\n';
 
+    auto const now = mxlGetTime();
+
     if (mxlIsDiscreteDataFormat(info.common.format))
     {
         os << '\t' << fmt::format("{: >18}: {}/{}", "Grain rate", info.discrete.grainRate.numerator, info.discrete.grainRate.denominator) << '\n'
            << '\t' << fmt::format("{: >18}: {}", "Grain count", info.discrete.grainCount) << '\n'
            << '\t' << fmt::format("{: >18}: {}", "Head index", info.discrete.headIndex) << '\n';
+
+        auto const currentIndex = mxlTimestampToIndex(&info.discrete.grainRate, now);
+        auto const latency = currentIndex - info.discrete.headIndex;
+
+        fmt::color color{fmt::color::green};
+        if (latency > static_cast<std::uint64_t>(info.discrete.grainCount))
+        {
+            color = fmt::color::red;
+        }
+        else if (latency == static_cast<std::uint64_t>(info.discrete.grainCount))
+        {
+            color = fmt::color::yellow;
+        }
+
+        os << '\t' << fmt::format(fmt::fg(color), "{: >18}: {}", "Latency (grains)", latency) << std::endl;
     }
     else if (mxlIsContinuousDataFormat(info.common.format))
     {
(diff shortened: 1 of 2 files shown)


sviengkhou (approved):

tools/mxl-info/main.cpp
```diff
@@ -47,11 +50,28 @@ std::ostream& operator<<(std::ostream& os, mxlFlowInfo const& info)
...
     if (mxlIsDiscreteDataFormat(info.common.format))
     {
         os << '\t' << fmt::format("{: >18}: {}/{}", "Grain rate", info.discrete.grainRate.numerator, info.discrete.grainRate.denominator) << '\n'
            << '\t' << fmt::format("{: >18}: {}", "Grain count", info.discrete.grainCount) << '\n'
            << '\t' << fmt::format("{: >18}: {}", "Head index", info.discrete.headIndex) << '\n';
+
+        auto const currentIndex = mxlTimestampToIndex(&info.discrete.grainRate, now);
+        auto const latency = currentIndex - info.discrete.headIndex;
+
+        fmt::color color{fmt::color::green};
```
KimonHoffmann: Could we use right hand type & init here, like so
`auto color = fmt::color{fmt::color::green}`, or probably even `auto color = fmt::color::green`?

tools/mxl-info/main.cpp
```diff
@@ -62,19 +82,66 @@ std::ostream& operator<<(std::ostream& os, mxlFlowInfo const& info)
...
-    }
 
-    auto const now = mxlGetTime();
-    auto const currentIndex = mxlTimestampToIndex(&info.discrete.grainRate, now);
-    os << '\t' << fmt::format("{: >18}: {}", "Latency (grains)", currentIndex - info.discrete.headIndex) << '\n'
-       << '\t' << fmt::format("{: >18}: {}", "Latency (ns)", now - info.common.lastWriteTime) << std::endl;
+        auto const currentIndex = mxlTimestampToIndex(&info.continuous.sampleRate, now);
+        auto const latency = currentIndex - info.continuous.headIndex;
+
+        fmt::color color{fmt::color::green};
```
KimonHoffmann: As above.

tools/mxl-info/main.cpp
```diff
@@ -62,19 +82,66 @@ std::ostream& operator<<(std::ostream& os, mxlFlowInfo const& info)
...
+
+        os << '\t' << fmt::format(fmt::fg(color), "{: >18}: {}", "Latency (samples)", latency) << std::endl;
+    }
 
     return os;
 }
 
+std::string getFlowLabel(std::filesystem::path const& descPath)
+{
+    std::string label{"Unnamed flow"};
```
KimonHoffmann: As above for all the variables and this function as well?

tools/mxl-info/main.cpp
```diff
@@ -62,19 +82,66 @@ std::ostream& operator<<(std::ostream& os, mxlFlowInfo const& info)
...
+
+    picojson::value v;
+    std::string err = picojson::parse(v, content);
+    if (!err.empty())
+    {
+        return label;
+    }
+
+    auto const& obj = v.get<picojson::object>();
+    auto flowTypeIt = obj.find("label");
```
KimonHoffmann: Can probably be const.

tools/mxl-info/main.cpp
```diff
@@ -62,19 +82,66 @@ std::ostream& operator<<(std::ostream& os, mxlFlowInfo const& info)
...
+    if (flowTypeIt == obj.end() || !flowTypeIt->second.is<std::string>())
```
KimonHoffmann: Having parentheses around the binary (`==`) sub-expression makes this a little easier to read.

vt-tv: Fixed

tools/mxl-info/main.cpp
vt-tv: Fixed

tools/mxl-info/main.cpp
vt-tv: Fixed

tools/mxl-info/main.cpp
vt-tv: Fixed

tools/mxl-info/main.cpp
vt-tv: fixed

PR #135, opened 2025-09-16
Add an invalid flow detection mechanism

vt-tv:
This is an issue that was initially reported by @pac-work  and that ended up being the most annoying issue at IBC. Every vendor suffered from this issue.

The issue occurs then an upstream writer restarts and recreates the flow resources using a fixed UUID.  A reader has no way to detect that the memory mapped structures are not being updated anymore (the tempfs data will point to different inodes)

This PR implement a simple detection logic based on capturing the flow data file inode value when opening a flow.  Then, when a flow reader reading operations (get grain, get samples)

### DIFF
--- a/lib/src/internal/FlowManager.cpp
+++ b/lib/src/internal/FlowManager.cpp
@@ -89,7 +89,7 @@ namespace mxl::lib
             result.format = format;
 
             // Get the inode of the flow data file
-            struct stat st;
+            struct ::stat st;
             if (::stat(flowDataPath.string().c_str(), &st) != 0)
             {
                 auto const error = errno;


lib/src/internal/FlowManager.cpp
```diff
@@ -128,14 +129,30 @@ namespace mxl::lib
...
+            auto const flowDataPath = makeFlowDataFilePath(tempDirectory);
+            auto flowData = std::make_unique<DiscreteFlowData>(flowDataPath.string().c_str(), AccessMode::CREATE_READ_WRITE);
 
             auto& info = *flowData->flowInfo();
             info.version = 1;
             info.size = sizeof info;
 
             info.common = initCommonFlowInfo(flowId, flowFormat);
 
+            // Get the inode of the flow data file
```
pac-work: Can this become part of the `initCommonFlowInfo`? That way it would not have to be copy+pasted for both the discrete and continuous.

lib/src/internal/PosixDiscreteFlowReader.cpp
```diff
@@ -140,16 +145,38 @@ namespace mxl::lib
...
+        // We may have attempted to read out of range on an invalid flow. let the reader know.
+        if (status != MXL_STATUS_OK && !isFlowValid())
+        {
+            status = MXL_ERR_FLOW_INVALID;
+        }
+
+        return status;
+    }
+
+    bool PosixDiscreteFlowReader::isFlowValid() const
```
pac-work, line 166: Not sure what is the MXL's project policy for use of multiple base class + virtual inheritance, but this one could theoretically refactored out into some common `PosixFlowReader` and thus avoid code duplication. But no strong opinion on this one - pick your poison - either multiple / virtual inheritance or a bit of code duplication.

pac-work (changes requested):
I think the move of the inode number initialization code into `initCommonFlowInfo` is worth it, the other one is really up to consideration and personal preference.

Otherwise - I have tested the code with our product and all works fine. Restarting the writer forced our product to re-connect, which shows that this works at least for one of the MXL users ;-) Thanks a lot, this will improve quality of life a lot.

lib/src/internal/FlowManager.cpp
vt-tv: fixed.

lib/src/internal/PosixDiscreteFlowReader.cpp
vt-tv: AFAIK the current approach is ok and will resolve to the right implementation at the cost of a little duplication.

pac-work (approved):
Thanks for resolving my previous comment.

My meaningless approval attached, @vt-tv ... somebody now needs to click it for real ;-)

lib/src/internal/PosixDiscreteFlowReader.cpp
pac-work: Yes, it works, my comment was only about potential to remove the little code duplication. All good in my eyes.

lib/src/internal/FlowManager.cpp
```diff
@@ -87,6 +88,20 @@ namespace mxl::lib
             result.lastReadTime = result.lastWriteTime;
             result.format = format;
 
+            // Get the inode of the flow data file
+            struct stat st;
```
jonasohland:
```suggestion
            struct ::stat st;
```

lib/src/internal/FlowManager.cpp
```diff
@@ -87,6 +88,20 @@ namespace mxl::lib
...
+            if (stat(flowDataPath.string().c_str(), &st) != 0)
```
jonasohland:
```suggestion
            if (::stat(flowDataPath.string().c_str(), &st) != 0)
```

vt-tv: fixed

lib/src/internal/FlowManager.cpp
vt-tv: fixed

vt-tv:
The new policies in place now force me to merge the PRs using the bypass rules option.  This is a known issue.

PR #136, opened 2025-09-16
Fix buildling.md dockerfile links

vt-tv:


vt-tv:
> I thought we already approved this and applied it to the main branch. The links [here](https://github.com/dmf-mxl/mxl/blob/main/docs/Building.md) look fine.

Can you try to browse the markdown links in a local checkout using VSCode? I think the github UI is trying to be very smart and resolves it by file name, not path.

PR #140, opened 2025-09-16
Take into account the vcpkg.json file when building the image hash

vt-tv:
Also : keep up with dependency updates (fmt and catch2).


KimonHoffmann (commented):
Looks good to me!

KimonHoffmann (approved):

PR #143, opened 2025-09-17
Perform C headers check at build time

vt-tv:
Add a dummy target that pulls in the sdk public header files and compiles them using the C17 language standard and strict options.   First commit adds the mechanism and proves that it works (see build history).  Second commit fixes the issues found in the previous build.


sviengkhou (approved):

lib/tests/CMakeLists.txt
```diff
@@ -99,3 +99,26 @@ catch_discover_tests(mxl-tests)
...
+    C_STANDARD_REQUIRED ON
+    C_EXTENSIONS OFF
+)
+
+target_include_directories(mxl-all-headers PRIVATE
+    "${CMAKE_CURRENT_SOURCE_DIR}/../include"
+)
+
+target_compile_options(mxl-all-headers PRIVATE
+    -std=c17
```
KimonHoffmann: This seems redundant, as the properties already specify C17, isn't it?

KimonHoffmann (changes requested):
Good idea to provide the conformance check in that manner!
Just one question/remark below.

lib/tests/CMakeLists.txt
vt-tv: Good catch.  Fixed.

PR #157, opened 2025-09-26
Strict json validation. Require label, grouphint.  rework tests

vt-tv:


lib/src/internal/FlowParser.cpp
```diff
@@ -97,9 +106,91 @@ namespace mxl::lib
             {
                 result.denominator = static_cast<std::int64_t>(it->second.get<double>());
             }
+
+            // Normalize the rational. We should realistically only see x/1 or x/1001 here.
+            auto g = std::gcd(result.numerator, result.denominator);
```
sviengkhou, line 111: I would do this just to be safe, sinc gcd can return '0' in very specific cases.
`if (g != 0) {
    result.numerator   /= g;
    result.denominator /= g;
}`

vt-tv: Modified.

tools/mxl-info/main.cpp
```diff
@@ -128,18 +128,20 @@ std::ostream& operator<<(std::ostream& os, mxlFlowInfo const& info)
     return os;
 }
 
-std::string getFlowLabel(std::filesystem::path const& descPath)
+void getFlowDetails(std::filesystem::path const& descPath, std::string& label, std::string& groupHint)
```
jonasohland, line 131: Either mark arguments that do return-by-reference like: `labelOut`, or return a `std::pair` and receive in a structured binding decl.

vt-tv: I started using the in/out syntax and was recommended to stop doing this in a previous pr.

jonasohland: Well, its personal preference I guess :)

PR #159, opened 2025-10-01
Build shared library and tests from common object library

jonasohland:
Create a cmake object library that contains all library objects. The shared library and the tests are linked to this object lib. Avoids having to build the objects twice.

Also adds a linker flag to avoid exposing symbols from the static libraries that we link into our library to be exposed on the library surface. 

Before:
```
nm -gD build/Linux-Clang-Debug/lib/libmxl.so | count
7585
```

After:
```
nm -gD build/Linux-Clang-Debug/lib/libmxl.so | count
355
```

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -5,30 +5,26 @@ include(GNUInstallDirs)
 
 # Configure the version.h header file base on the current project version
 configure_file(
-    ${CMAKE_CURRENT_SOURCE_DIR}/include/mxl/version.h.in
-    ${CMAKE_CURRENT_BINARY_DIR}/include/mxl/version.h
-    @ONLY
-)
+        ${CMAKE_CURRENT_SOURCE_DIR}/include/mxl/version.h.in
+        ${CMAKE_CURRENT_BINARY_DIR}/include/mxl/version.h
+        @ONLY
+    )
 
-# Files to be compiled into the library
-add_library(mxl SHARED)
-target_compile_features(mxl
-        PRIVATE
+add_library(mxl-objects OBJECT)
+target_compile_features(mxl-objects
+        PUBLIC
             cxx_std_20
     )
-set_target_properties(mxl
-        PROPERTIES
-            POSITION_INDEPENDENT_CODE    ON
-            VISIBILITY_INLINES_HIDDEN    ON
-            C_VISIBILITY_PRESET          hidden
-            CXX_VISIBILITY_PRESET        hidden
-            C_EXTENSIONS                 OFF
-            CXX_EXTENSIONS               OFF
-            VERSION                      "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}"
-            SOVERSION                    "${PROJECT_VERSION_MAJOR}"
+set_target_properties(mxl-objects
+        PROPERTIES 
+            POSITION_INDEPENDENT_CODE ON      # Needed to bulid shared library later
+            VISIBILITY_INLINES_HIDDEN ON
+            C_VISIBILITY_PRESET       hidden
+            CXX_VISIBILITY_PRESET     hidden
+            C_EXTENSIONS              OFF
+            CXX_EXTENSIONS            OFF
     )
-
-target_sources(mxl
+target_sources(mxl-objects
         PRIVATE
             src/flow.cpp
             src/mxl.cpp
@@ -54,8 +50,7 @@ target_sources(mxl
             src/internal/Thread.cpp
             src/internal/Time.cpp
             src/internal/Timing.cpp
-        )
-
+    )
 
 if (NOT TARGET stduuid)
     find_package(stduuid CONFIG REQUIRED)
@@ -73,20 +68,45 @@ if (NOT TARGET picojson::picojson)
     find_package(picojson REQUIRED)
 endif()
 
-target_include_directories(mxl PUBLIC
-    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
-    $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>
-    $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
-)
-target_link_libraries(mxl
+target_include_directories(mxl-objects
         PRIVATE
+            $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+            $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>
+    )
+target_link_libraries(mxl-objects
+        PUBLIC
             stduuid
             spdlog::spdlog
             fmt::fmt
             picojson::picojson
     )
 
-# Alias trace to libtrace::libtrace so that this library can be used
+add_library(mxl SHARED)
+target_compile_features(mxl
+        PRIVATE
+            cxx_std_20
+    )
+set_target_properties(mxl
+        PROPERTIES
+            VERSION                      "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}"
+            SOVERSION                    "${PROJECT_VERSION_MAJOR}"
+    )
+target_link_options(mxl
+        PRIVATE
+            "-Wl,--exclude-libs,ALL"
+    )
+
+target_include_directories(mxl 
+        PUBLIC
+            $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
+    )
+
+target_link_libraries(mxl
+        PRIVATE
+            mxl-objects
+    )
+
+# Alias mxl to mxl::mxl so that this library can be used
 # in lieu of a module from the local source tree
 add_library(${PROJECT_NAME}::mxl ALIAS mxl)
 add_subdirectory(tests)
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -26,9 +26,6 @@ set_target_properties(mxl-objects
     )
 target_sources(mxl-objects
         PRIVATE
-            src/flow.cpp
-            src/mxl.cpp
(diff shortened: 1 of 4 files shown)


lib/CMakeLists.txt
```diff
@@ -10,52 +10,44 @@ configure_file(
...
-            src/internal/PosixDiscreteFlowReader.cpp
-            src/internal/PosixDiscreteFlowWriter.cpp
-            src/internal/PosixFlowIoFactory.cpp
-            src/internal/SharedMemory.cpp
-            src/internal/Sync.cpp
-            src/internal/Thread.cpp
-            src/internal/Time.cpp
-            src/internal/Timing.cpp
-        )
+target_compile_features(mxl-objects PUBLIC cxx_std_20)
```
KimonHoffmann: Could you please keep the indenting as it was before, i.e.:

```
target_compile_features(mxl-objects
        PUBLIC
            cxx_std_20
    )
```

lib/CMakeLists.txt
```diff
@@ -73,18 +65,38 @@ if (NOT TARGET picojson::picojson)
...
+    PRIVATE
+        stduuid
+        spdlog::spdlog
+        fmt::fmt
+        picojson::picojson
+)
+
+add_library(mxl SHARED)
+
+target_link_libraries(mxl PRIVATE mxl-objects)
```
KimonHoffmann: Could you please use similar indenting as the other declarations, i,e,:

```cmake
target_link_libraries(mxl
    PRIVATE
        mxl-objects
)
```

lib/CMakeLists.txt
```diff
@@ -73,18 +65,38 @@ if (NOT TARGET picojson::picojson)
...
-    $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>
-    $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
+target_include_directories(mxl-objects 
+    PUBLIC
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>
+)
+
+target_link_libraries(mxl-objects
+    PRIVATE
```
KimonHoffmann: I'm not really sure whether `PRIVATE` is correct here (since an object library is not really a true library). I think `PUBLIC` would be more appropriate, because downstream dependents might need to interact with those libraries as well, when using functionality from `mxl-objects`.

lib/CMakeLists.txt
```diff
@@ -73,18 +65,38 @@ if (NOT TARGET picojson::picojson)
...
+
+set_target_properties(mxl
+    PROPERTIES
+        VERSION                      "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}"
+        SOVERSION                    "${PROJECT_VERSION_MAJOR}"
+)
+
+target_link_options(mxl PRIVATE
+    "-Wl,--exclude-libs,ALL"
+)
```
KimonHoffmann: Could you please use similar indenting as used in the other invocations, i.e.:

```cmake
target_link_options(mxl
    PRIVATE
        "-Wl,--exclude-libs,ALL"
)
```

lib/CMakeLists.txt
```diff
@@ -73,18 +65,38 @@ if (NOT TARGET picojson::picojson)
...
+
+target_link_libraries(mxl-objects
+    PRIVATE
+        stduuid
+        spdlog::spdlog
+        fmt::fmt
+        picojson::picojson
+)
+
+add_library(mxl SHARED)
```
KimonHoffmann, line 81: Even though 1. `mxl` does not currently have any sources of it's own and 2. the feature is inherited from the `mxl-objects` dependency, I think it makes sense to keep the explicit mention of

```cmake
target_compile_features(mxl
    PRIVATE
        cxx_std_20
)
```

For future extensions where `mxl` might gain a few additional sources again, and to force an explicit clash if one target is updated without taking care of the others.

lib/CMakeLists.txt
```diff
@@ -10,52 +10,44 @@ configure_file(
...
-            POSITION_INDEPENDENT_CODE    ON
-            VISIBILITY_INLINES_HIDDEN    ON
-            C_VISIBILITY_PRESET          hidden
-            CXX_VISIBILITY_PRESET        hidden
-            C_EXTENSIONS                 OFF
-            CXX_EXTENSIONS               OFF
-            VERSION                      "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}"
-            SOVERSION                    "${PROJECT_VERSION_MAJOR}"
-    )
+add_library(mxl-objects OBJECT
```
KimonHoffmann: Could you please keep the previous structure where the target declaration and adding the sources to it is separated?

lib/CMakeLists.txt
```diff
@@ -10,52 +10,44 @@ configure_file(
...
+    src/flow.cpp
+    src/mxl.cpp
+    src/time.cpp
```
KimonHoffmann: I think these three are conceptually `PRIVATE` to `mxl` rather than being part of `mxl-objects`.

lib/CMakeLists.txt
```diff
@@ -73,18 +65,38 @@ if (NOT TARGET picojson::picojson)
...
 
-target_include_directories(mxl PUBLIC
-    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
-    $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>
-    $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
+target_include_directories(mxl-objects 
+    PUBLIC
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>
+)
```
KimonHoffmann: If the three C API implementation files move back to `mxl`, as suggested above, then these two would also move back there.

What this library does not yet, but should declare as an include directory is the source directory where the internal headers are located.

tools/mxl-gst/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ if (gstreamer_FOUND)
     endif ()
 
 
-    add_executable(mxl-gst-videotestsrc)
+    add_executable(mxl-gst-videotestsrc videotestsrc.cpp)
```
KimonHoffmann: Could you please keep the previous separation of the declaration of the target and its sources?

tools/mxl-gst/CMakeLists.txt
```diff
@@ -37,11 +37,9 @@ if (gstreamer_FOUND)
                 C_EXTENSIONS                 OFF
                 CXX_EXTENSIONS               OFF
         )
-    target_sources(mxl-gst-videotestsrc
+    target_link_libraries(mxl-gst-videotestsrc
```
KimonHoffmann: Could you merge this into the `target_link_libraries` call below?

tools/mxl-gst/CMakeLists.txt
```diff
@@ -56,7 +54,7 @@ if (gstreamer_FOUND)
             )
 
 
-    add_executable(mxl-gst-videosink)
+    add_executable(mxl-gst-videosink videosink.cpp)
```
KimonHoffmann: As above.

tools/mxl-gst/CMakeLists.txt
```diff
@@ -70,11 +68,9 @@ if (gstreamer_FOUND)
                 C_EXTENSIONS                 OFF
                 CXX_EXTENSIONS               OFF
         )
-    target_sources(mxl-gst-videosink
+    target_link_libraries(mxl-gst-videosink
```
KimonHoffmann: As above.

tools/mxl-gst/CMakeLists.txt
```diff
@@ -89,7 +85,7 @@ if (gstreamer_FOUND)
                 PkgConfig::gstreamer-video
         )
 
-    add_executable(mxl-gst-looping-filesrc)
+    add_executable(mxl-gst-looping-filesrc looping_filesrc.cpp)
```
KimonHoffmann: As above.

tools/mxl-gst/CMakeLists.txt
```diff
@@ -103,11 +99,9 @@ if (gstreamer_FOUND)
         C_EXTENSIONS OFF
         CXX_EXTENSIONS OFF
     )
-    target_sources(mxl-gst-looping-filesrc
+    target_link_libraries(mxl-gst-looping-filesrc
```
KimonHoffmann: As above.

tools/mxl-info/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@
 
 include(GNUInstallDirs)
 
-add_executable(mxl-info)
+add_executable(mxl-info main.cpp)
```
KimonHoffmann: As above.

tools/mxl-info/CMakeLists.txt
```diff
@@ -17,10 +17,9 @@ set_target_properties(mxl-info
             C_EXTENSIONS                 OFF
             CXX_EXTENSIONS               OFF
     )
-target_sources(mxl-info
+target_link_libraries(mxl-info
```
KimonHoffmann: As above.

lib/CMakeLists.txt
jonasohland: But then what do you do if you want to link to the `mxl` target dynamically? The executables unter `tools/` for example: They would then all have the objects built into them instead of linking to `libmxl.so`.
If you want to statically link all the code into your target, you should link to `mxl-objects` explicitly.

KimonHoffmann (approved):

lapointejp (approved):

PR #165, opened 2025-10-04
fix SIGABRT in DomainWatcher

GalleryUK:
fix SIGABRT because the vector was only reserved and not resized before directly accessing the members.


vt-tv (approved):

jonasohland (approved):

lib/src/internal/DomainWatcher.cpp
```diff
@@ -279,6 +279,7 @@ namespace mxl::lib
                 std::lock_guard<std::mutex> lock(_mutex);
                 watchCount = _watches.size();
                 eventsToMonitor.reserve(watchCount);
```
garethsb, line 281: There is no point in a reserve immediately before a resize, should have just changed it to the latter. Or more in keeping with the original code, might have been to add a `push_back` before the `op[]` below.

Issue #12, opened 2025-06-23
AV Synchronisation


vt-tv:
@KimonHoffmann  Please describe the work you are currently doing around synchronisation between granular flows and continuous flows.

KimonHoffmann:
In order to allow a media function to deal with multiple flows in parallel it needs a way to determine a source time offset at which it can operate on grains and continuous buffer slices without the need for waiting.

IMO this is best achieved by having the producer of each flow specify a (configurable) flow delay (relative to the origin time) that is communicated to consumers e.g. via the FlowInfo header.

Data arriving after this configured delay is to be considered "lost" and must not be commited. Instead it is the responsibility of the producer to properly deal with lost data, by e.g. incrementing the head index as if the data arrived, but marking the corresponding grains as "unavailable", or e.g. by zeroing out the affected slices in a continuous buffer.

With this information a consumer of multiple flows (that is interested in properly aligned flows) can determine the maximum expected delay of all flows it operates on and use that as its working point.

KimonHoffmann:
Thank you for moving the discussion here.

While I have yet to benchmark the `futex_wake` call myself, if it falls into the mentioned ballpark then I think there is nothing that would keep us from offering a synchronization interface for audio flows that follows the same semantics as the video an ANC flows.

But just for completeness be aware that while 200ns seems to be negligible on first glance it can quickly add up in systems with multiple audio flows. E.g. if you have a producer that implements -30/-31 RXs that handles 64 streams per thread (a reasonable quantity in order to achieve a reasonable density) and the streams are using 125µs packet times the time taken just to signal the received data adds up to a little bit more than 10% of the available time slice.

Using the `futex_waitv()` syscall unfortunately does not help a whole lot with waiting for multiple streams (aside from it only supporting up to 128 futexes to concurrently wait on), because  it wakes up when *any* of the futexes gets signals, not when *all* of them got signaled The rational behind this is obvious, because this kind of API *can not* be implemented client code, while the "wait for all" operation can be easily achieved by doing something like this:

```cpp
for (auto& futex : futexesToWaitOn)
{
    futex.wait();
}
```

The problem here ist that while this operation is *semantically* equivalent, it has quite abhorrent performance characteristics, because it (in the worst case) incurs a kernel mode transition for each futex.

[...]

PR #168, opened 2025-10-07
Change partial grain mechanism: byte-indexed to slice-indexed

### LINKED ISSUE #161: Reconsider commitedSize definition for partial grain/frame
In MXL, partial grain/frame commits are implemented by setting `commitedSize` to a value lower than `grainSize` in the `mxlGrainInfo` structure.

The current Fabrics API implementation uses the "immediate data" feature available with RDMA Remote Write to communicate the grain index being transferred to the target. The immediate data field is limited to **32 bits**, but since ring buffer addressing uses simple modulo arithmetic, we can easily recover the actual grain index by using the current time for the MSB and the immediate data for the LSB.

To support partial grains through the Fabrics API, we need to communicate `commitedSize` to the target. I see two possible approaches:

1. Transfer `commitedSize` in a separate message (1× Remote Write + 1× Send, or 2× Remote Write)
2. Squeeze both `commitedSize` and `grainIndex` into the 32-bit immediate data field

Analyzing the bit requirements for larger formats:

V210 3840×2160: 22,389,120 bytes → 25 bits required
V210 7680×4320: 89,015,040 bytes → 27 bits required

This would be even worse for 4:4:4 16-bit formats if we decide to support them in the future. Taking the worst case of V210 7680×4320, this leaves only **5 bits** for `grainIndex`, supporting a maximum of 32 ring buffer entries (approximately half a second for a 60Hz stream). This doesn't scale well.

I don't think we actually need the byte-level precision that `commitedSize` currently provides. A coarser granularity could free up bits and would allow a more reasonable `grainIndex` range.

### LINKED ISSUE #166: mxlGrainInfo "commitdSize" minor typo
The field "commitedSize" in mxlGrainInfo should have two "t"s.  The typo is in the comment that documents the field name and in the field name itself. 

```
/// A grain is complete when commitedSize == grainSize
 uint32_t commitedSize;
```

There are other instances of the "commited" typo as follows:
```
% cd .../mxl
% find . -type f -name \* -exec grep -i commited {} /dev/null \; | cut -d\: -f1 | sort | uniq
./lib/include/mxl/flow.h
./lib/src/internal/PosixDiscreteFlowWriter.cpp
./lib/tests/test_flows_timing.cpp
./lib/tests/test_flows.cpp
./tools/mxl-gst/looping_filesrc.cpp
./tools/mxl-gst/videosink.cpp
./tools/mxl-gst/videotestsrc.cpp
```

jonasohland:
Changes how partial grains are committed. Instead of tracking the valid part of the grain as a number of bytes, it now is a number of slices. 
The definition of a slice depends on the data format. For video, a slice is now defined as a single line of a frame. For data, its defined as a single byte of payload. The length in bytes of a slice can be queried from the new field: `mxlDiscreteFlowInfo.sliceLength`. 

From `mxlGrainInfo` the `committedSize` was removed and is replaced by 2 new fields: `totalSlices` and `validSlices`. `totalSlices` represents the maxium number of slices that can be committed to the grain, `validSlices` is the number of slices that have been committed (and are now valid).

This is how slices relate to the grain size:
```
grainInfo.grainSize == grainInfo.totalSlices * flowInfo.sliceLength
``` 

The full grain is valid when:
```
grainInfo.validSlices == grainInfo.totalSlices
```

We have also moved the `commitBatchSize` and `syncBatchSize` from the continuous flow info to the common flow info header, because we think they could be useful for sub-frame synchronization in video flows too. There is currently no way to configure these values, so that would be TBD and part of the discussion in #12.

Because these are breaking changes in the flow structures in shared memory, we have also increased the version number. This will have no effect, and still lead to broken consumers when the consumers expect previous version flow headers because the library did not check the versions until now. We have also added a version check.

This fixes #166 (be removing the field).
And it fixes #161. `validSlices` is a uint16_t, which is more than enough for all our formats, and fits into the 32-bit immediate data that we have to cram it into together with a ring-buffer index.


lib/include/mxl/flow.h
```diff
@@ -122,7 +122,7 @@ extern "C"
 
     typedef struct mxlGrainInfo_t
     {
-        /// Version of the structure. The only currently supported value is 1
+        /// Version of the structure. The only currently supported value is 2
```
vt-tv, line 125: should we bother with this for the time being?  until v1 is released I would leave this to 1.  we are not guaranteeing any binary compatiblity across versions until v1 is released.

lib/src/internal/FlowParser.cpp
```diff
@@ -27,6 +27,9 @@ namespace mxl::lib
         constexpr auto MAX_WIDTH = 7680u;  // 8K UHD
         constexpr auto MAX_HEIGHT = 4320u; // 8K UHD
 
+        // Grain size when the grain data format is "data"
+        constexpr auto DATA_FORMAT_GRAIN_SIZE = 4096;
```
vt-tv, line 31: Data means ANC in the spirit of 2110-40 / SDI VANC lines.  so technically is can be as big as : 32 lines * 2 bytes per pixel * width.

jonasohland:
> Just a terminology question. should we use 'size' or 'length'? we seem to be using the words interchangably, even in a single statement:
> 
> grainInfo.grainSize == grainInfo.totalSlices * flowInfo.sliceLength

I had a this discussion with @mlefebvre1 also...

Here is my (kind of esoteric) reasoning for choosing length over size for the `sliceLength`, feel free to disregard...

Looking at other code that floats around the internet, there seems to be no definitive rule for when to use size and length.
My gut-feeling is: Length describes the extent of something in one dimension, while size is more applicable in multi-dimensional scenarios.
And since the grainSize describes the **size** of a amorphous blob of anything, the sliceLength refers to the **length** of something that feels like a one-dimensional string of bytes...

lib/include/mxl/flowinfo.h
```diff
@@ -44,6 +44,19 @@ extern "C"
...
 
+        /**
+         * The largest expected batch size in samples (for continuous flows) or slices (for discrete flows), in which new data is written to this this
+         * flow by its producer. For continuous flows, this value must be less than half of the buffer length. For discrete flows, this must be >1.
+         */
+        uint32_t commitBatchSize;
+
+        /**
+         * The largest expected batch size in samples (for continuous flows) or slices (for discrete flows), at which availability of new data is
+         * signaled to waiting consumers. This must be a multiple of the commit batch size greater or equal to 1.
```
lapointejp, line 56: I am slightly worried about this for continuous (audio) flows.
I previously enjoyed the granularity of audio being to the sample because that we allowed for the north american audio cadence that matches video. So you could commit 800 or 801 samples at a time in 60000/1001 for example.

The `commitBatchSize` is field is described as the largest back that will be written and the `syncBatchSize` field is described as a multiple of the `commitBatchSize`. I'm not sure I can achieve the fractional video rate audio cadence elegantly with these fields.

lib/include/mxl/flowinfo.h
```diff
@@ -62,6 +75,12 @@ extern "C"
          */
         uint32_t grainCount;
 
+        /**
+         * Length of a slice in bytes. A slice refers to the elemental data type that can be written and comitted to a grain.
+         * For video, this is a line of a V210 picture. For data, this is just a single byte.
```
lapointejp: This includes any padding necessary if the pictures don't line up on 6 pixel widths ?

lib/src/internal/DiscreteFlowData.hpp
```diff
@@ -56,7 +58,19 @@ namespace mxl::lib
...
-        return _grains.emplace_back(grainFilePath, mode, grainPayloadSize).get();
+        auto grain = SharedMemoryInstance<Grain>{grainFilePath, mode, grainPayloadSize};
+
+        if (!this->created())
+        {
+            // Check for the version of the grain data structure in the memory that was just mapped.
+            if (grain.get()->header.info.version != GRAIN_DATA_VERSION)
+            {
+                throw std::invalid_argument(
+                    fmt::format("Unsupported grain version: {}, supported version is: {}", grain.get()->header.info.version, GRAIN_DATA_VERSION));
```
lapointejp: Nitpick: grain *data structure* version ?
It's not the grain itself that is the wrong version.

lapointejp (commented):
I haven't gone through all the details, but I did a quick run through and highlighted a few points I think should be discussed before going through everything else.

lib/include/mxl/flowinfo.h
jonasohland: `totalSlices*sliceLength == grainSize`
So it should include the padding. I will add that to the comment.

lib/src/internal/DiscreteFlowData.hpp
jonasohland: Maybe `GRAIN_HEADER_VERSION`?

lib/include/mxl/flowinfo.h
KimonHoffmann: The intention of these fields is to give the producer the ability to communicate meta information that is relevant for synchronization to downstream consumers.
* `commitBatchSize` is the **typical** (but conservative -- therefor max) batch size, in which new data is written. For a -30 RX this could for example be the packet time.
* Anticipating that signaling a futex at the commit rate might be a bit excessive I also provided for `syncBatchSize`, which indicates at what rate a consumer can expect to be signaled.

Both fields are intended to allow the consumer to gauge by how far an index it is actively waiting for might be delayed. `commitBatchSize` is mostly relevant for **synchronous** consumers that have to decide how long to sleep, while `syncBatchSize` is obviously more intended for **asynchronous** consumers.

As stated above both are conservative estimates, so there is nothing really keeping a producer from 1. committing less data than the specified maximum per write and from 2. signaling at a slightly higher (on average) rate than the specified maximum.

lib/include/mxl/flowinfo.h
```diff
@@ -44,6 +44,20 @@ extern "C"
...
+         * The largest expected batch size in samples (for continuous flows) or slices (for discrete flows), in which new data is written to this this
+         * flow by its producer. For continuous flows, this value must be less than half of the buffer length. For discrete flows, this must be
+         * greater or equal to 1.
+         */
+        uint32_t commitBatchSize;
+
+        /**
+         * The largest expected batch size in samples (for continuous flows) or slices (for discrete flows), at which availability of new data is
+         * signaled to waiting consumers. This must be a multiple of the commit batch size greater or equal to 1.
+         * \todo Will quite probably be obsoleted by new timing model.
```
KimonHoffmann: You can remove that comment, because the recent discussions have proven that this type of synchronization is necessary for many applications.

lib/include/mxl/flowinfo.h
```diff
@@ -73,7 +93,7 @@ extern "C"
         uint64_t headIndex;
 
         /** Reserved space for future extensions.  */
-        uint8_t reserved[96];
+        uint8_t reserved[88];
```
KimonHoffmann, line 95: It's rather unfortunate that the new `sliceLength` introduces four bytes of padding. Since we're changing the data structure version anyway we an also rearrange existing fields. Is there a way to rearrange the fields to avoid padding while still keeping them in a relatively sane order?

lib/include/mxl/flowinfo.h
```diff
@@ -95,19 +115,6 @@ extern "C"
...
-         */
-        uint32_t commitBatchSize;
-
-        /**
-         * The largest expected batch size in samples, at which availability of new data is signaled to waiting consumers.
-         * This must be a multiple of the commit batch size greater or equal to 1.
-         * \todo Will quite probably be obsoleted by new timing model.
-         */
-        uint32_t syncBatchSize;
-
```
KimonHoffmann, line 110: Removing these two requires increasing the padding at the end of the struct.

lib/src/internal/DiscreteFlowData.hpp
KimonHoffmann: + Curlies instead of parentheses around the exception constructor argument?

lib/src/internal/FlowManager.cpp
```diff
@@ -3,6 +3,7 @@
 
 #include "FlowManager.hpp"
 #include <cstring>
+#include <exception>
```
KimonHoffmann: This is the wrong header and should be `stdexcept` instead.
Edit: Which already exists below ... as I just realized

lib/src/internal/FlowParser.cpp
```diff
@@ -390,6 +393,68 @@ namespace mxl::lib
...
+                    auto msg = std::string{"Unsupported video media_type: "} + mediaType;
+                    throw std::invalid_argument{std::move(msg)};
+                }
+
+                return static_cast<std::size_t>((width + 47) / 48 * 128);
+            }
+
+            default: {
+                throw std::invalid_argument{"Cannot compute slice length for this data format."};
+            }
```
KimonHoffmann, line 425: According to our conventions the opening curlies should be placed on a line of their own (and can be omitted for short cases that do not introduce variables into the scope).

lib/src/internal/FlowParser.cpp
```diff
@@ -390,6 +393,68 @@ namespace mxl::lib
...
+                {
+                    return h / 2;
+                }
+
+                return h;
+            }
+            default: {
+                throw std::invalid_argument{"Cannot compute slice length for this data format."};
+            }
+        }
```
KimonHoffmann, line 461: As above.

lib/tests/test_flows.cpp
```diff
@@ -452,12 +452,23 @@ TEST_CASE("Video Flow : Slices", "[mxl flows]")
     REQUIRE(mxlFlowReaderGetInfo(reader, &fInfo1) == MXL_STATUS_OK);
     REQUIRE(fInfo1.discrete.headIndex == 0);
 
-    size_t const maxSlice = 8;
-    auto sliceSize = gInfo.grainSize / maxSlice;
-    for (size_t slice = 0; slice < maxSlice; slice++)
+    // Total number of batches that will be written
+    std::size_t const numBatches = std::ceil(static_cast<double>(gInfo.totalSlices) / static_cast<double>(fInfo1.common.commitBatchSize));
```
KimonHoffmann:
```cpp
auto const numBatches = (gInfo.totalSlices + fInfo1.common.commitBatchSize - 1U) / fInfo1.common.commitBatchSize;
```

lib/tests/test_flows.cpp
```diff
@@ -452,12 +452,23 @@ TEST_CASE("Video Flow : Slices", "[mxl flows]")
...
+    std::size_t defaultBatchSize = fInfo1.common.commitBatchSize;
+
+    for (size_t batchIndex = 0; batchIndex < numBatches; batchIndex++)
```
KimonHoffmann: `auto batchSize = std::size_t{0};`

vt-tv:
> > > Just a terminology question. should we use 'size' or 'length'? we seem to be using the words interchangably, even in a single statement:
> > > grainInfo.grainSize == grainInfo.totalSlices * flowInfo.sliceLength
> > 
> > 
> > Here is my (kind of esoteric) reasoning for choosing length over size for the `sliceLength`, feel free to disregard...
> > [...]
> > My gut-feeling is: Length describes the extent of something in one dimension, while size is more applicable in multi-dimensional scenarios. And since the grainSize describes the **size** of a amorphous blob of anything, the sliceLength refers to the **length** of something that feels like a one-dimensional string of bytes...
> 
> My -- entirely personal -- rule of thumb I (try to ) follow is: Use **length** (or in some contexts **count**), when referring to quantities of not necessarily byte sized items, like samples, pixels, lines, items in a collection, ... and use **size**, when referring to the size in bytes either of single items or of sequences of such items.

@KimonHoffmann  that's my understanding of how the length size and count terms are used in C++

lib/src/internal/FlowParser.cpp
jonasohland: I will add a property to the `BraceWrapping` section in the clang-format configuration for this.
This will change some code that is unrelated to this PR, but not so much that it is a problem, I think.

KimonHoffmann:
> Do you want me to change it to `sliceSize`?

Makes sense to me.

lib/include/mxl/flowinfo.h
lapointejp: I get your point.
What would your suggested use be for a fractional rate producer ?
800 samples for both, with the understanding that it might spit out 801 as well ?

Should we call these "hints" ? Given that they express and approximate value ? The comments makes them seem strict.

KimonHoffmann: In that specific case I would specify them as 801 to adhere to the requested property that these values express a maximum, and then just signal after either 800 or 801 samples depending on where in the sequence we are.

lapointejp: Makes sense to me.
Then I would just suggest that we clearly indicate it is a maximum, but a hint on the typical chunk size we would do. But it works for me.

jonasohland: Are we okay with `max{Commit/Sync}BatchSizeHint`?

lib/include/mxl/flowinfo.h
```diff
@@ -138,7 +144,7 @@ extern "C"
         };
 
         /** User data space. */
-        uint8_t userData[3840];
+        uint8_t userData[3832];
```
KimonHoffmann, line 147: Why did this decrease by 8?

lib/include/mxl/flow.h
vt-tv: @mlefebvre1  can we leave this to 1? no compatibility claims until v1.0

lib/include/mxl/flowinfo.h
jonasohland: I have added a check for the sizes of all the structs and that there is no padding here: https://github.com/dmf-mxl/mxl/pull/168/files#diff-ca3b479e1f91ad50a1712a6c7fd67a32e845d9a467cc194732a300df5454fd10

Changing this was necessary to bring this structs size to 4096. I am not sure why... Maybe this was wrong before?

mlefebvre1: Yeah, it is wrong on the main branch, 4 + 4 + 128 + 128 + 3840 = 4104

KimonHoffmann: Perfect, thank you for fixing this, and for adding the necessary checks that I also suggested in another PR, as these have been very much overdue!

vt-tv (approved):

PR #170, opened 2025-10-07
Add group hint tags to the json generated by the looping file source …

vt-tv:


tools/mxl-gst/looping_filesrc.cpp
```diff
@@ -309,7 +307,13 @@ class LoopingFilePlayer
         root["parents"] = picojson::value(picojson::array());
         root["media_type"] = picojson::value("video/v210");
 
-        picojson::object grain_rate;
+        auto tags = picojson::object{};
+        auto groupHint = picojson::array{};
+        groupHint.emplace_back(picojson::value{"Looping Source:Video"});
+        tags["urn:x-nmos:tag:grouphint/v1.0"] = picojson::value(groupHint);
```
jonasohland, line 313: Would it make sense to define some of these keys as constants somewhere in the library?

jonasohland (approved):

PR #185, opened 2025-10-15
Unhide service files

### LINKED ISSUE #132: Should "service files" like `.json`, `.options`, `.access` or similar in the flow directory be hidden?
May I ask what is a reason for this? It just seems to complicate debugging. There is other stuff in the flow directory, which is perfectly visible, so why these files are hidden? User should not be depending on these files anyway.

pac-work:
Closes #132.


vt-tv (approved):

vt-tv:
This needs a minor version bump since it breaks the world.

vt-tv (changes requested):
Update the minor version please in the main CMakeLists.txt

PR #190, opened 2025-10-16
Add v210+alpha support.  Fix ANC test case, improve docs.

vt-tv:


jonasohland (changes requested):
Needs an extension of the discreteFlowInfo.sliceSize field to allow for multiple different slice sizes of the different payload buffers.
Would also provide a handy way to discover if there is alpha or not:

```
if (flowInfo.discrete.sliceSize[1] != 0) {
    // alpha
}
```

Everything else lgtm.

docs/architecture.md
```diff
@@ -85,14 +85,28 @@ The following example demonstrates how a FlowWriter may write a grain as slices
...
-### Audio
+Video grains can be of two different formats: video/v210 for video without transparency and video/v210+alpha for fill and key signals (video with alpha transparency).
+
+### video/v210
+
+The v210 format is an uncompressed buffer format carrying 10 bit 4:2:2 video. A detailed description of the format can be found [here](https://wiki.multimedia.cx/index.php/V210) and [here](https://developer.apple.com/library/archive/technotes/tn2162/_index.html#//apple_ref/doc/uid/DTS40013070-CH1-TNTAG8-V210__4_2_2_COMPRESSION_TYPE).
+
+#### video/v210+alpha
+
+This format contains both fill and key inside a single grain.  The fill part starts at byte 0 of the grain and follows the v210 definition above. The key buffer is found immediately after the fill buffer.  Samples are organized in blocks of 32 bit values in little-endian.  Each block contains 3 luma samples, one each in bits 0 - 9, 10 - 19 and 20 - 29, the remaining two bits are unused.  The start of each line is aligned to a multiple of 4 bytes, where unused blocks are padded with 0.
```
mlefebvre1: One can infer it from "The start of each line is aligned to a multiple of 4 bytes, where unused blocks are padded with 0. ", but I would probably mention that exceptionally the last block of a line might have more padding than just the last 2 bits if the width is not divisible by 3. 

ex: 1280x720 resolution, has padding for bits 20 to 31 on the last block.

lib/internal/src/FlowParser.cpp
```diff
@@ -350,6 +350,29 @@ namespace mxl::lib
...
             }
+            else if (mediaType == "video/v210+alpha")
+            {
+                if (!_interlaced || ((height % 2) == 0))
+                {
+                    // Interlaced media is handled as separate fields.
+                    auto const h = _interlaced ? height / 2 : height;
+
+                    // Fill size
+                    auto fillPayloadSize = static_cast<std::size_t>((width + 47) / 48 * 128) * h;
```
lapointejp: Nitpick: wondering if we should have a small inline function for V210 line size given it's  redo from plain "video/v210".

lib/internal/src/FlowParser.cpp
```diff
@@ -350,6 +350,29 @@ namespace mxl::lib
...
+                    // Key is stored as 10 bits per pixel, 3x10 bit words in a 32 bit word, little-endian.
+                    // the last 2 bits of the 32 bit group of 3 pixels are unused.
+                    auto keyPayloadSize = static_cast<std::size_t>(4 * ((width + 2) / 3) * h);
+
+                    // Total payload size is the sum of the fill and key sizes
+                    payloadSize = fillPayloadSize + keyPayloadSize;
+                }
+                else
+                {
+                    auto msg = std::string{"Invalid video height for interlaced v210. Must be even."};
```
lapointejp: Technically this is v210+alpha...

lib/internal/src/FlowParser.cpp
vt-tv: Added a MediaUtils.hpp/cpp file and used it throughout the codel

docs/architecture.md
vt-tv: Comment updated.

lib/include/mxl/flowinfo.h
```diff
@@ -13,6 +13,13 @@
 #include <mxl/dataformat.h>
 #include <mxl/rational.h>
 
+// Maximum number of planes per grain supported for a continuous flow.
+//
+// - video/v210 flow will have only 1 plane out of MXL_MAX_PLANES_PER_GRAIN
+// - video/v210+alpha flow will have 2 planes out of MXL_MAX_PLANES_PER_GRAIN
+//
```
jonasohland, line 22: Suggestion: Add a comment that explains why we chose the number 4 (possible support for planar formats).

lib/internal/include/mxl-internal/MediaUtils.hpp
```diff
@@ -0,0 +1,24 @@
...
+{
+    /**
+     * Length in bytes of a line in the V210 video format.
+     * @param width The width of the video frame in pixels.
+     * @return The line length in bytes for the V210 format, including padding.
+     */
+    std::uint32_t getV210LineLength(std::size_t width);
+
+    /**
+     * Length in bytes of a line in the 10-bit ùlpha video format (3x10-bit samples per 32 bit word).
```
jonasohland:
```suggestion
     * Length in bytes of a line in the 10-bit alpha video format (3x10-bit samples per 32 bit word).
```

lib/internal/include/mxl-internal/MediaUtils.hpp
```diff
@@ -0,0 +1,24 @@
+// SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project.
```
jonasohland, line 1: Tiny nitpick: I think "MediaUtils" could be anything and there are a lot of helper functions that would have to be moved into this file consequently. So my suggestion would be: `VideoFormatUtils` or `VideoFormatHelpers`.

vt-tv: That was actually on purpose.  instead of scattering lots of utils in many files I wanted to regroup them in a single one for the time being. if you are ok with it I'll resolve this conversation.

lib/internal/include/mxl-internal/MediaUtils.hpp
vt-tv: fixed.

lib/include/mxl/flowinfo.h
vt-tv: added comment.

jonasohland (approved):

Issue #186, opened 2025-10-15
[TSC] How do we attach an Requirement to a Feature?


felixpou:
Do we create a Feature issue type?
How do we attach an Requirement to a Feature? (could be just an hyperlink in the description but if there is a way to add a specific that could enable future automation.)

peterbrightwell:
I've been experimenting on my fork, and it looks like we need to remove the SPDX headers from issue templates for them to work. In fact we already have several custom issues, which should be appearing when we click on New Issue but don't because of the headers.

peterbrightwell:
This is now merged in #195. When someone clicks on Create new issue they have to choose from five options.  In this case use "Feature" (not "Feature request" which is about new ideas (maybe we should rename that "Idea"?)

pbwell:
Yes showing the Parent looks very good in table view, so I'll see if that can be automated to work with the Feature template.

felixpou:
the trigger of the workflow is not when we create an issue of type "Feature"? then we have it already, I don't see a need for an extra label, we can filter by type.

felixpou:
so @peterbrightwell are we good to implement that?

When creating a Feature
-new issue type Feature
-add parent to the Requirement
-no need for a label or a custom field

felixpou:
@peterbrightwell can we finalise this?

peterbrightwell:
The action is currently failing most of the time, but not always.  It looks like most feature issues are being created without using the Feature template.

Example of one that used the template and works: #318 (which links to requirement 109)
Example of one didn't and the action failed: #327

felixpou:
**TSC**

- Let's use the type, not label

peterbrightwell:
That didn't work, and I went down some blind alleys trying to debug with a manual workflow_dispatch and on my fork (both of which can't work, but for different reasons).  Still to be fixed, and I recommend creating a new sandbox repo for further testing as this needs to be debugged on `main`.

peterbrightwell:
I have made https://github.com/dmf-mxl/mxl-workflow-sandbox for further work on this and other workflow activities that would cause multiple commits on `main` and even more dummy test issues. This is best done in this organisation so it can use the same apps/secrets/variables. 

It's still not working of course, but at least can be investigated without unnecessary noise and triggering unnecessary builds of MXL!

peterbrightwell:
Ok I think this working in the sandbox. See https://github.com/dmf-mxl/mxl-workflow-sandbox/issues/14

[...]

PR #195, opened 2025-10-17
Add Feature issue template with Requirement Issue link

peterbrightwell:
See #186 -- this adds a new issue template for Feature with a mandatory field to link to the related requirement.

It also moves the SPDX headers to the bottom of the templates as having them at the top means that the templates aren't recoginsed by GitHub.


felixpou (approved):
Let's try this.

PR #199, opened 2025-10-23
Fabrics 1/2: Library setup and C++ Wrappers

### LINKED ISSUE #126: Fabrics integration approach
Hi all,

We are ready to start our first step to integrate our fabrics branch into MXL now (#78). As discussed in one of the last TSC meetings, we will begin with the CMake and project setup. We will probably do this from scratch. The changes in our branch were really just made to get it working.

The approach that we initially went with in our branch was to create a separate shared object for the fabrics implementation to avoid a dependency of the main mxl library on libfabric. This has one big downside though:
We need to link to some of the C++ code from the main library to get access to the raw flow memory that is mapped for a flow reader or writer. Since the mxl library is built with `-fvisibility=hidden`, we need to link those symbols statically (right now we create a static library, but we could also create a single [cmake object library](https://cmake.org/cmake/help/latest/command/add_library.html#object-libraries) for both libraries, as @KimonHoffmann pointed out. 
Now, a broken upgrade for example could have lead to different versions of `libmxl.so` and `libmxl-fabrics.so` being installed. And when we cast a pointer received through the API to its C++ representation, we might receive a layout that is different from what the code that we have statically linked expects.

To fix this, we think there a 5 possible approaches:
1. Do some kind of internal version check thingy between the two libraries (I don't like that).
2. Change visibility of the C++ symbols on `libmxl.so` to public, so we can use them from `libmxl-fabrics.so`.
3. Build everything into one library, accept that we need to link against libfabric.
4. Build everything into one library and do not link libfabric, instead load libfabric symbols with `dlopen` when the user calls `mxlFabricsCreateInstance`
5. Try to link libfabric statically, possibly with vcpkg. Problem is: The current vcpkg libfabric port has no CUDA support, so we would have to fix that first, and I think there is value in that people can build their own version of libfabric with the providers and options they need for their usecases and then use that with mxl.

My favored approach would be (4), and there might be more options that I don't see right now.

### LINKED ISSUE #180: Fabrics: PR Step 1: Libfabric C++ Abstractions
Step 1 in getting the PR merged.

jonasohland:
Note: This PR wants to be merged on a feature branch, but should be treated as if intended to be merged into main. The feature branch just exists so we can split the review work into 2 steps.

This changes some things in the CMake project:

1. Creates a new interface library `mxl-fabrics-headers` intended to be linked by tools or libraries that want to use the internal headers in `lib/internal/include/mxl-internal`. The intention is to avoid exposing these headers to targets that link the new `mxl-common` lib (formerly `mxl-internal`). 
2. Rename `mxl-internal` to `mxl-common` and make it a shared library. This new shared library contains C++ code that will  be used from both `libmxl` and `libmxl-fabrics`.

The libraries built and installed now look like this:
```
ls -l build/install/lib/ 
total 15440
drwxr-xr-x 3 ohlano ohlano     4096 Oct 23 03:18 cmake/
drwxr-xr-x 2 ohlano ohlano     4096 Oct 23 03:18 gstreamer-1.0/
lrwxrwxrwx 1 ohlano ohlano       18 Oct 23 03:18 libmxl-common.so -> libmxl-common.so.0*
lrwxrwxrwx 1 ohlano ohlano       20 Oct 23 03:18 libmxl-common.so.0 -> libmxl-common.so.0.8*
-rwxr-xr-x 1 ohlano ohlano 11105688 Oct 23 03:11 libmxl-common.so.0.8*
lrwxrwxrwx 1 ohlano ohlano       19 Oct 23 03:18 libmxl-fabrics.so -> libmxl-fabrics.so.0*
lrwxrwxrwx 1 ohlano ohlano       21 Oct 23 03:18 libmxl-fabrics.so.0 -> libmxl-fabrics.so.0.8*
-rwxr-xr-x 1 ohlano ohlano  3402032 Oct 23 03:11 libmxl-fabrics.so.0.8*
lrwxrwxrwx 1 ohlano ohlano       11 Oct 23 03:18 libmxl.so -> libmxl.so.0*
lrwxrwxrwx 1 ohlano ohlano       13 Oct 23 03:18 libmxl.so.0 -> libmxl.so.0.8*
-rwxr-xr-x 1 ohlano ohlano  1286016 Oct 23 03:11 libmxl.so.0.8*
drwxr-xr-x 2 ohlano ohlano     4096 Oct 23 03:18 pkgconfig/
```

And targets that link to the `mxl` cmake target will link both `mxl` and `mxl-common`:
```
ldd build/Linux-Clang-Debug/tools/mxl-info/mxl-info
        linux-vdso.so.1 (0x00007f39a2d79000)
        libmxl.so.0 => /home/ohlano/src/riedel/til/mxl/build/Linux-Clang-Debug/lib/libmxl.so.0 (0x00007f39a2c34000)
        libmxl-common.so.0 => /home/ohlano/src/riedel/til/mxl/build/Linux-Clang-Debug/lib/internal/libmxl-common.so.0 (0x00007f39a2800000)
        libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x00007f39a2400000)
        libm.so.6 => /usr/lib/libm.so.6 (0x00007f39a2aeb000)
        libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00007f39a2abe000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007f39a2000000)
        /lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007f39a2d7b000)
```
Fixes #180
Fixes #126

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -74,9 +74,9 @@ target_sources(mxl
 target_link_libraries(mxl
         PUBLIC
             mxl-headers # C Headers in <mxl/*.h>
-            mxl-common  # Common objects for libmxl and libmxl-fabrics
         PRIVATE
             mxl-internal-headers # Internal C++ headers in <mxl-internal/*.hpp>
+            mxl-common  # Common objects for libmxl and libmxl-fabrics
     )
 
 # Alias mxl to mxl::mxl so that this library can be used
--- a/lib/fabrics/include/mxl/fabrics.h
+++ b/lib/fabrics/include/mxl/fabrics.h
@@ -5,9 +5,9 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <mxl/flow.h>
 #include <mxl/mxl.h>
-#include "mxl/flow.h"
-#include "mxl/platform.h"
+#include <mxl/platform.h>
 
 #ifdef __cplusplus
 extern "C"
--- a/lib/internal/include/mxl-internal/FlowData.hpp
+++ b/lib/internal/include/mxl-internal/FlowData.hpp
@@ -4,7 +4,7 @@
 #pragma once
 
 #include <cstddef>
-#include "mxl/platform.h"
+#include <mxl/platform.h>
 #include "Flow.hpp"
 #include "SharedMemory.hpp"
 
--- a/lib/internal/src/DomainWatcher.cpp
+++ b/lib/internal/src/DomainWatcher.cpp
@@ -14,9 +14,9 @@
 #include <fcntl.h>
 #include <unistd.h>
 #include <uuid.h>
+#include <mxl/platform.h>
 #include "mxl-internal/Logging.hpp"
 #include "mxl-internal/PathUtils.hpp"
-#include "mxl/platform.h"
 
 #ifdef __APPLE__
 #   include <sys/event.h>
@@ -27,7 +27,6 @@
 
 namespace mxl::lib
 {
-    MXL_EXPORT
     DomainWatcher::DomainWatcher(std::filesystem::path const& in_domain, Callback in_callback)
         : _domain{in_domain}
         , _callback{std::move(in_callback)}
@@ -109,7 +108,6 @@ namespace mxl::lib
         }
     }
 
-    MXL_EXPORT
     DomainWatcher::~DomainWatcher()
     {
         _running = false;
@@ -143,7 +141,6 @@ namespace mxl::lib
 #endif
     }
 
-    MXL_EXPORT
     int16_t DomainWatcher::addFlow(uuids::uuid const& in_flowId, WatcherType in_type)
     {
         auto id = uuids::to_string(in_flowId);
@@ -201,7 +198,6 @@ namespace mxl::lib
         return useCount;
     }
 
-    MXL_EXPORT
     int16_t DomainWatcher::removeFlow(uuids::uuid const& in_flowId, WatcherType in_type)
     {
         auto id = uuids::to_string(in_flowId);
--- a/tools/mxl-gst/CMakeLists.txt
+++ b/tools/mxl-gst/CMakeLists.txt
@@ -46,6 +46,7 @@ if (gstreamer_FOUND)
             PRIVATE
                 mxl
                 mxl-internal-headers
+                mxl-common
                 CLI11::CLI11
                 PkgConfig::gstreamer
                 PkgConfig::gstreamer-app
@@ -76,6 +77,7 @@ if (gstreamer_FOUND)
             PRIVATE
                 mxl
                 mxl-internal-headers
+                mxl-common
                 CLI11::CLI11
                 spdlog::spdlog
                 PkgConfig::gstreamer
@@ -107,6 +109,7 @@ if (gstreamer_FOUND)
             PRIVATE
                 mxl
                 mxl-internal-headers
+                mxl-common
                 CLI11::CLI11
                 PkgConfig::gstreamer
                 PkgConfig::gstreamer-app
--- a/lib/internal/include/mxl-internal/DomainWatcher.hpp
+++ b/lib/internal/include/mxl-internal/DomainWatcher.hpp
@@ -14,6 +14,7 @@
 #include <unordered_map>
 #include <unistd.h>
 #include <uuid.h>
+#include <mxl/platform.h>
 
 namespace mxl::lib
 {
@@ -57,7 +58,7 @@ namespace mxl::lib
     /// by readers when they read a grain, which will trigger a 'FlowInfo.lastRead` update (performed by the FlowWriter
     /// since the writer is the only one mapping the flow in write-mode).
     ///
-    class DomainWatcher
+    class MXL_EXPORT DomainWatcher
     {
     public:
         using Callback = std::function<void(uuids::uuid const&, WatcherType in_type)>;
--- a/lib/internal/include/mxl-internal/Flow.hpp
+++ b/lib/internal/include/mxl-internal/Flow.hpp
@@ -7,6 +7,7 @@
 #include <cstdint>
 #include <iostream>
 #include <mxl/flow.h>
+#include <mxl/platform.h>
 
 namespace mxl::lib
 {
--- a/lib/internal/include/mxl-internal/FlowManager.hpp
+++ b/lib/internal/include/mxl-internal/FlowManager.hpp
@@ -10,6 +10,7 @@
 #include <string>
 #include <vector>
 #include <uuid.h>
+#include <mxl/platform.h>
 #include "ContinuousFlowData.hpp"
 #include "DiscreteFlowData.hpp"
 
@@ -37,7 +38,7 @@ namespace mxl::lib
     /// LIST
     /// List all the flows found in the domain.
     ///
-    class FlowManager
+    class MXL_EXPORT FlowManager
     {
     public:
         ///
--- a/lib/internal/include/mxl-internal/FlowParser.hpp
+++ b/lib/internal/include/mxl-internal/FlowParser.hpp
@@ -18,7 +18,7 @@ namespace mxl::lib
     /**
      * Parses a NMOS Flow resource and extracts / computes key elements based on the flow attributes.
      */
-    class FlowParser
+    class MXL_EXPORT FlowParser
     {
     public:
         /**
--- a/lib/internal/include/mxl-internal/Instance.hpp
+++ b/lib/internal/include/mxl-internal/Instance.hpp
@@ -13,6 +13,7 @@
 #include <string>
 #include <uuid.h>
 #include <mxl/mxl.h>
+#include <mxl/platform.h>
 #include "DomainWatcher.hpp"
 #include "FlowIoFactory.hpp"
 #include "FlowManager.hpp"
@@ -22,7 +23,7 @@ namespace mxl::lib
     ///
     /// Manages mxl resources allocated by the SDK user.  Represents a single mxfInstance.
     ///
-    class Instance
+    class MXL_EXPORT Instance
     {
     public:
         ///
--- a/lib/internal/include/mxl-internal/SharedMemory.hpp
+++ b/lib/internal/include/mxl-internal/SharedMemory.hpp
@@ -5,6 +5,7 @@
 
 #include <cstddef>
 #include <utility>
+#include <mxl/platform.h>
 
 namespace mxl::lib
 {
@@ -15,7 +16,7 @@ namespace mxl::lib
         CREATE_READ_WRITE
     };
 
-    class SharedMemoryBase
+    class MXL_EXPORT SharedMemoryBase
     {
     public:
         /**
--- a/lib/internal/src/FlowManager.cpp
+++ b/lib/internal/src/FlowManager.cpp
@@ -115,7 +115,6 @@ namespace mxl::lib
 
     }
 
-    MXL_EXPORT
     FlowManager::FlowManager(std::filesystem::path const& in_mxlDomain)
         : _mxlDomain{std::filesystem::canonical(in_mxlDomain)}
     {
@@ -126,7 +125,6 @@ namespace mxl::lib
         }
     }
 
-    MXL_EXPORT
     std::unique_ptr<DiscreteFlowData> FlowManager::createDiscreteFlow(uuids::uuid const& flowId, std::string const& flowDef, mxlDataFormat flowFormat,
         std::size_t grainCount, mxlRational const& grainRate, std::size_t grainPayloadSize, std::size_t grainNumOfSlices,
         std::array<uint32_t, MXL_MAX_PLANES_PER_GRAIN> grainSliceLengths)
@@ -200,7 +198,6 @@ namespace mxl::lib
         }
     }
 
-    MXL_EXPORT
     std::unique_ptr<ContinuousFlowData> FlowManager::createContinuousFlow(uuids::uuid const& flowId, std::string const& flowDef,
         mxlDataFormat flowFormat, mxlRational const& sampleRate, std::size_t channelCount, std::size_t sampleWordSize, std::size_t bufferLength)
     {
@@ -250,7 +247,6 @@ namespace mxl::lib
         }
     }
 
-    MXL_EXPORT
     std::unique_ptr<FlowData> FlowManager::openFlow(uuids::uuid const& in_flowId, AccessMode in_mode)
     {
         if (in_mode == AccessMode::CREATE_READ_WRITE)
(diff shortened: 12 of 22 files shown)


pac-work:
> @pac-work @pedro-alves-ferreira What did you do to get the rust branch to pass CI? We also have to change stuff in the Dockerfile, so the CI will try to push a new version of the container and we run in the same permission errors as you before.

Sorry for late reply - [this is the commit which made things work](https://github.com/dmf-mxl/mxl/pull/29/commits/ca2efede49869a18936bcb8ea454a389e1e587f9). Not sure what problem you are facing, we had trouble with CICD trying to push some containers, which it did not have right to do so.

pedro-alves-ferreira:
> > @pac-work @pedro-alves-ferreira What did you do to get the rust branch to pass CI? We also have to change stuff in the Dockerfile, so the CI will try to push a new version of the container and we run in the same permission errors as you before.
> 
> Sorry for late reply - [this is the commit which made things work](https://github.com/dmf-mxl/mxl/pull/29/commits/ca2efede49869a18936bcb8ea454a389e1e587f9). Not sure what problem you are facing, we had trouble with CICD trying to push some containers, which it did not have right to do so.

I saw the comment marked as resolved and I assumed you had overcome that. 

By the way, did you manage to test it locally, using `apt` or a similar tool? Or are you testing with the actual GitHub CI?

lib/fabrics/include/mxl/fabrics.h
```diff
@@ -0,0 +1,331 @@
...
+        char const* service;
+    } mxlEndpointAddress;
+
+    /** Configuration object required to set up a new target.
+     */
+    typedef struct mxlTargetConfig_t
+    {
+        mxlEndpointAddress endpointAddress; /**< Bind address for the local endpoint. */
+        mxlFabricsProvider provider;        /**< The provider that should be used */
+        mxlRegions regions;                 /**< Th Local memory regions of the flow that grains should be written to. */
```
lapointejp:
```suggestion
        mxlRegions regions;                 /**< Local memory regions of the flow that grains should be written to. */
```

lib/fabrics/include/mxl/fabrics.h
```diff
@@ -0,0 +1,331 @@
...
+        bool deviceSupport;                 /**< Th Require support of transfers involving device memory. */
```
lapointejp:
```suggestion
        bool deviceSupport;                 /**< Require support of transfers involving device memory. */
```

lib/fabrics/ofi/src/internal/Address.cpp
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    FabricAddress FabricAddress::fromBase64(std::string_view data)
+    {
+        return FabricAddress{base64::decode_into<std::vector<std::uint8_t>>(data)};
+
+        auto decoded = base64::decode_into<std::vector<std::uint8_t>>(data);
+        if (decoded.empty())
+        {
+            throw std::runtime_error("Failed to decode base64 data into FabricAddress");
```
lapointejp:
```suggestion
            throw std::runtime_error("Failed to decode base64 data into FabricAddress.");
```

lib/fabrics/ofi/src/internal/Completion.cpp
```diff
@@ -0,0 +1,126 @@
...
+    {}
+
+    Completion::Error Completion::err() const
+    {
+        if (auto error = std::get_if<Error>(&_inner); error)
+        {
+            return *error;
+        }
+
+        throw Exception::invalidState("Failed to unwrap completion queue entry as error");
```
lapointejp:
```suggestion
        throw Exception::invalidState("Failed to unwrap completion queue entry as error.");
```

lib/fabrics/ofi/src/internal/Completion.cpp
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    Completion::Data Completion::data() const
+    {
+        if (auto data = std::get_if<Completion::Data>(&_inner); data)
+        {
+            return *data;
+        }
+
+        throw Exception::invalidState("Failed to unwrap completion queue entry as data entry");
```
lapointejp:
```suggestion
        throw Exception::invalidState("Failed to unwrap completion queue entry as data entry.");
```

lib/fabrics/ofi/src/internal/Endpoint.hpp
```diff
@@ -0,0 +1,283 @@
...
+#include "RemoteRegion.hpp"
+
+namespace mxl::lib::fabrics::ofi
+{
+    /** \brief RAII Wrapper around a libfabric endpoint (`fi_ep`).
+     */
+    class Endpoint
+    {
+    public:
+        /** \brief The miniumum interval at which the event queue will be polled when
```
lapointejp:
```suggestion
        /** \brief The minimum interval at which the event queue will be polled when
```

lib/fabrics/ofi/src/internal/Endpoint.hpp
```diff
@@ -0,0 +1,283 @@
...
+
+        /** \brief Accept an incoming connection.
+         *
+         * Information about the incoming connection must previously have been passed to Endpoint::create().
+         * A successfully accepted connection request will result in the active (connecting) endpoint seeing an Event::Connected event on its
+         * associated event queue. A rejected or failed connection request will generate an error event. The error entry will provide additional
+         * details describing the reason for the failed attempt. An Event::Connected event will also be generated on the passive side for the
+         * accepting endpoint once the connection has been properly established. The fid of the FI_CONNECTED event will be that of the endpoint
+         * passed to fi_accept as opposed to the listening passive endpoint. Outbound data transfers cannot be initiated on a connection-oriented
+         * endpoint until an FI_CONNECTED event has been generated. However, receive buffers may be associated with an endpoint anytime
```
lapointejp:
```suggestion
         * endpoint until an FI_CONNECTED event has been generated. However, receive buffers may be associated with an endpoint anytime.
```

lib/fabrics/ofi/src/internal/Endpoint.hpp
```diff
@@ -0,0 +1,283 @@
...
+         * If an event queue is associated with this endpoint. Get the queue. Throws an exception if no event queue is
+         * associated with this endpoint.
+         */
+        [[nodiscard]]
+        std::shared_ptr<EventQueue> eventQueue() const;
+
+        /** \brief Get the associated addres vector.
+         *
+         * If an address vector is associated with this endpoint. Get the AV. Throws an exception if no address vector is
+         * associated with this endpoint
```
lapointejp:
```suggestion
         * associated with this endpoint.
```

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -0,0 +1,652 @@
...
+        uint8_t* dummyPayload;
+        mxlStatus status;
+
+        while (!g_exit_requested)
+        {
+            status = mxlFabricsTargetWaitForNewGrain(_target, &grainIndex, 200);
+            if (status == MXL_ERR_TIMEOUT)
+            {
+                // No completion before a timeout was triggered, most likely a problem upstream.
+                // MXL_WARN("wait for new grain timeout, most likely there is a problem upstream.");
```
lapointejp: Any reason this was left commented ?

mlefebvre1: I don't believe there's a valid reason for it, so I'll uncomment it. Also, thank you for all the corrections you've done.

lapointejp: My pleasure :)

lib/internal/CMakeLists.txt
```diff
@@ -59,17 +81,20 @@ if (NOT TARGET picojson::picojson)
     find_package(picojson REQUIRED)
 endif()
 
-target_include_directories(mxl-internal
-        PUBLIC
+target_include_directories(mxl-common
```
KimonHoffmann, line 85: This should not be necessary, because the `PRIVATE` dependency to `mxl-internal-headers` should take care of that.

lib/internal/CMakeLists.txt
```diff
@@ -3,21 +3,43 @@
...
             cxx_std_20
     )
-set_target_properties(mxl-internal
+set_target_properties(mxl-common
         PROPERTIES 
-            POSITION_INDEPENDENT_CODE ON      # Needed to bulid shared library later
+            POSITION_INDEPENDENT_CODE ON
             VISIBILITY_INLINES_HIDDEN ON
-            C_VISIBILITY_PRESET       hidden
-            CXX_VISIBILITY_PRESET     hidden
```
KimonHoffmann, line 16: If we turn the object library into a true shared library, we should take control of the symbols exported (instead of indiscriminately exporting everything) by means of annotating all (class) types (i.e. with non-inline methods) and (non-inline) functions available through `mcx-internal-headers` with `MXL_EXPORT` and the restore these two visibility presets.

lib/CMakeLists.txt
```diff
@@ -73,14 +73,20 @@ target_sources(mxl
 
 target_link_libraries(mxl
         PUBLIC
-            mxl-headers
+            mxl-headers # C Headers in <mxl/*.h>
+            mxl-common  # Common objects for libmxl and libmxl-fabrics
```
KimonHoffmann: Shouldn't `mxl-common` be a private dependency?

lib/fabrics/ofi/CMakeLists.txt
```diff
@@ -0,0 +1,104 @@
...
+        @ONLY
+    )
+
+# Install the generated pkg-config file
+install(FILES
+            ${CMAKE_CURRENT_BINARY_DIR}/libmxl-fabrics.pc
+        DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig
+        COMPONENT ${PROJECT_NAME}-dev
+    )
+
```
KimonHoffmann, line 140: This misses the generation and installation of a native cmake config. Unfortunately -- due to the depedency on `libfabric` -- this cmake config can not be quite as bare as the one for the main library.

I can contribute an appropriate config if you want.

.devcontainer/scripts/common/libfabric/install.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
KimonHoffmann: Should nowadays be `/usr/bin/env bash`.

.devcontainer/scripts/common/libfabric/install.sh
```diff
@@ -0,0 +1,23 @@
...
+# SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project.
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+set -x
+
+git clone -b v2.2.0 https://github.com/ofiwg/libfabric.git
+
+pushd libfabric >/dev/null
```
KimonHoffmann: Using `pushd`/`popd` instead of `cd` here is unnecessary.

.devcontainer/scripts/debian/register-clang-version.sh
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env bash
+
+# SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project.
+# SPDX-License-Identifier: Apache-2.0
+#
+#     --slave /usr/bin/$1 $1 /usr/bin/$1-\${version} \\
```
KimonHoffmann: How did those two get in here? Could you please use the opportunity to remove those two lines?

.devcontainer/scripts/common/libfabric/install.sh
```diff
@@ -0,0 +1,23 @@
...
+    --prefix=/usr \
+    --disable-kdreg2 \
+    --disable-memhooks-monitor \
+    --disable-uffd-monitor
+
+make install "-j$(nproc)"
+
+popd >/dev/null
+
+exit 0
```
KimonHoffmann: This is unnecessary and prevents forwarding of the error code returned by `make` (when `popd` is removed as mentioned above) should the scorept ever switch to not using `set -e`.

lib/fabrics/ofi/CMakeLists.txt
mlefebvre1: If you have time, that would be appreciated.

KimonHoffmann: Done!

tools/mxl-gst/CMakeLists.txt
```diff
@@ -44,8 +44,9 @@ if (gstreamer_FOUND)
 
     target_link_libraries(mxl-gst-videotestsrc
             PRIVATE
-                mxl-internal
                 mxl
+                mxl-internal-headers
```
mlefebvre1: Link these 2 to tools applications feels wrong. We should probably think how we can expose these data/functions through the C API.

lib/CMakeLists.txt
mlefebvre1: Actually, I don't think, otherwise other apps that use mxl-common symbols (tools apps, maybe they should not??) and also tests will need to link mxl-common instead of just mxl.

lib/internal/include/mxl-internal/FlowData.hpp
```diff
@@ -4,6 +4,7 @@
 #pragma once
 
 #include <cstddef>
+#include "mxl/platform.h"
```
KimonHoffmann: While both works in this case we have the convention in other files to include headers from `mxl/` with angle brackets.

lib/internal/src/DomainWatcher.cpp
```diff
@@ -108,6 +109,7 @@ namespace mxl::lib
         }
     }
 
+    MXL_EXPORT
```
KimonHoffmann: I don't think adding these per method is necessary, when the class as a whole has an export specification?

KimonHoffmann: Never mind, I just realized that the class as a whole isn't exported.

But in order to keep the effort low I would be fine with doing that in all cases for now to get a good mix of not exporting everything but also not cluttering the code with very granular exports.

Your choice. :)

lib/CMakeLists.txt
KimonHoffmann: You are correct, I was under the assumption that the symbols could still be accessed by the dynamic linker due to the transitive de-facto dependency.

The tools using internal APIs is a sore spot that has been bothering me for a while now, I think we should raise this with the TSC to discuss how we go about that. Designing a library API that is not able to support its own demo applications is really not a good look. 😉 

Regarding the tests: I'm (much) less bothered by this since they're supposed to put the code through the wringer and are IMO free to reach into the internals if the need to to accomplish that goal. For those I'm completely fine if they explicitly link to mxl-common.

mlefebvre1: I agree with tests. In that case, I will make mxl-common private, add it as a dependency everywhere it's needed. We can remove that dependency once the test tools are fixed.

lib/internal/src/DomainWatcher.cpp
mlefebvre1: Okay, I was under the impression that we didn't want to export private functions.

PR #204, opened 2025-10-23
feat: Add grainIndex to grain header

mlefebvre1:
* Add index field in mxlGrainInfo struct. Allows to associate a grainIndex to the ring buffer entry.
* Move payload location and device index to mxlCommonFlowInfo, because this information is not specific to a grain but rather to a flow.

### DIFF
--- a/lib/include/mxl/flow.h
+++ b/lib/include/mxl/flow.h
@@ -118,7 +118,7 @@ extern "C"
         /// Size of the structure
         uint32_t size;
 
-        /// Absolute Grain index used by that ring buffer entry.
+        /// Epoch Grain index used by that ring buffer entry.
         uint64_t index;
 
         /// Grain flags.
--- a/lib/internal/src/PosixDiscreteFlowWriter.cpp
+++ b/lib/internal/src/PosixDiscreteFlowWriter.cpp
@@ -65,6 +65,11 @@ namespace mxl::lib
     {
         if (_flowData)
         {
+            if (mxlGrainInfo.index != _currentIndex)
+            {
+                return MXL_ERR_INVALID_ARG;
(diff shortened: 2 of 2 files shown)


lib/include/mxl/flow.h
```diff
@@ -126,12 +117,12 @@ extern "C"
         uint32_t version;
         /// Size of the structure
         uint32_t size;
+
+        /// Absolute Grain index used by that ring buffer entry.
```
vt-tv: index since the epoch or index within the ringbuffer 'array'?

lib/internal/src/PosixDiscreteFlowWriter.cpp
```diff
@@ -40,6 +38,7 @@ namespace mxl::lib
         {
             auto offset = in_index % _flowData->flowInfo()->discrete.grainCount;
             auto const grain = _flowData->grainAt(offset);
+            grain->header.info.index = in_index; // Set the absolute grain index associated to that ring buffer entry
```
vt-tv, line 41: Should we also validate in the ::commit(mxlGrainInfo const& mxlGrainInfo) function that the index of the mxlGrainInfo argument is still what we expect? the user may have changed it.

lib/include/mxl/flow.h
mlefebvre1: I can re-phrase this, but it is the index since epoch

lib/internal/src/PosixDiscreteFlowWriter.cpp
mlefebvre1: will do

jonasohland (approved):

PR #208, opened 2025-10-24
feat: Complete audio support of mxl tools

mlefebvre1:
* Add audio test source in videotestsrc app. Allows to configure the number of desired channels, first channel starts at a frequency of 100hz and subsequent channels increases by 100hz. Also added control for the grain/sample offset.
* Modify videosink app to be able to select which channel to listen via the -l option. Also added control for grain/sample offset. Also I've made changes so that you can listen to audio only, video only or both streams.

### DIFF
--- a/tools/mxl-gst/videosink.cpp
+++ b/tools/mxl-gst/videosink.cpp
@@ -393,7 +393,8 @@ namespace
                     mxlFlowReaderGetInfo(_reader, &_flowInfo);
                     MXL_WARN("Failed to get grain at index {}: TOO LATE. Last published {}", requestedIndex, _flowInfo.discrete.headIndex);
 
-                    // We are too late. Re-align.. The player will handle missing samples
+                    // Grain expired. Realign to current index. GStreamer repeats the last valid frame for missing data; consuming applications
+                    // should do the same.
                     index = mxlGetCurrentIndex(&rate);
                     continue;
                 }
@@ -417,7 +418,9 @@ namespace
 
                 if (!(grainInfo.flags & MXL_GRAIN_FLAG_INVALID))
                 {
-                    if (timeoutMode)
+                    // We've validated the grain. Invalid grains are skipped rather than pushed to GStreamer. Since we provide PTS values based on MXL
+                    // timestamps, GStreamer automatically handles missing grains by repeating the last valid frame. Consuming applications should
+                    // implement similar logic for invalid grain handling.                    if (timeoutMode)
                     {
                         updateHighestLatency(mxlIndexToTimestamp(&rate, requestedIndex), mxlGetTime());
                     }
@@ -464,9 +467,11 @@ namespace
                 }
                 else if (ret == MXL_ERR_OUT_OF_RANGE_TOO_LATE)
                 {
+                    // Samples expired. Realign to current index. GStreamer will generate silence for missing samples. Consuming applications
+                    // should handle this better by inserting silence with a micro fades to prevent clicks and pops.
                     mxlFlowReaderGetInfo(_reader, &_flowInfo);
                     MXL_WARN("Failed to get samples at index {}: TOO LATE. Last published {}", requestedIndex, _flowInfo.continuous.headIndex);
-                    // We are too late. The player will handle missing samples
+
                     index = mxlGetCurrentIndex(&rate);
                     continue;
                 }


tools/mxl-gst/videotestsrc.cpp
```diff
@@ -199,36 +261,257 @@ class GstreamerPipeline final
...
+    auto videoFlowConfigFileOpt = app.add_option(
+        "-v, --video-config-file", videoFlowConfigFile, "The json file which contains the Video NMOS Flow configuration");
+    videoFlowConfigFileOpt->default_val("");
+
+    std::string audioFlowConfigFile;
+    auto audioFlowConfigFileOpt = app.add_option(
+        "-a, --audio-config-file", audioFlowConfigFile, "The json file which contains the Audio NMOS Flow configuration");
+    audioFlowConfigFileOpt->default_val("");
+
+    uint64_t samplesPerBatch;
```
mlefebvre1, line 665: I've added that, because there's no way right now to set `maxSyncBatchSizeHint` and `maxCommitBatchSizeHint`

vt-tv (commented):
Should we try to align everything using GstClockTime extracted from the GstBuffers?

tools/mxl-gst/videosink.cpp
```diff
@@ -51,301 +39,445 @@ namespace
...
+                {
+                    MXL_ERROR("Unexpected error when reading the grain {} with status '{}'. Exiting.", index, static_cast<int>(ret));
+                    return ret;
+                }
+
+                if (grainInfo.validSlices != grainInfo.totalSlices)
+                {
+                    // Full frame not ready yet
+                    continue;
+                }
```
vt-tv, line 417: We should also verify that the grain itself is valid or invalid. (grainInfo.flags & MXL_GRAIN_FLAG_INVALID).  If the grain is invalid the application must implement an invalid grain strategy such as repeating the previous grain.

tools/mxl-gst/videosink.cpp
```diff
@@ -133,13 +133,13 @@ namespace
...
-            _mxlClockOffset = mxlGetTime() - baseTime;
-            MXL_INFO("Video pipeline base time: {} ns, MXL clock offset: {} ns", baseTime, _mxlClockOffset);
+            _gstBaseTime = mxlGetTime() - baseTime;
+            MXL_INFO("Video pipeline base time: {} ns, MXL clock offset: {} ns", baseTime, _gstBaseTime);
         }
 
         void pushSample(GstBuffer* buffer, std::uint64_t now) final
         {
-            GST_BUFFER_PTS(buffer) = now - _mxlClockOffset;
+            GST_BUFFER_PTS(buffer) = now - _gstBaseTime;
```
vt-tv, line 144: shouldn't the PTS be based on the ringbuffer index / sample offset instead of use the wall clock?

docs/Tools.md
```diff
@@ -99,15 +99,32 @@ A binary that uses the gstreamer element 'videotestsrc' to produce video grains
...
-Usage: ./build/Linux-GCC-Release/tools/mxl-gst/mxl-gst-videotestsrc [OPTIONS]
 
-Options:
-  -h,--help                   Print this help message and exit
-  -f,--flow-config-file TEXT REQUIRED
-                              The json file which contains the NMOS Flow configuration
-  -d,--domain TEXT:DIR REQUIRED
+
+./build/Linux-Clang-Debug/tools/mxl-gst/mxl-gst-videotestsrc [OPTIONS]
+
```
sviengkhou, line 105: I would add an example argument for Audio and Video, Video Only and Audio Only

docs/Tools.md
```diff
@@ -116,11 +133,27 @@ A binary that reads from a MXL Flow and display the flow using the gstreamer ele
...
 mxl-gst-videosink
-Usage: ./build/Linux-GCC-Release/tools/mxl-gst/mxl-gst-videosink [OPTIONS]
 
-Options:
-  -h,--help                   Print this help message and exit
-  -f,--flow-id TEXT REQUIRED  The flow ID
-  -d,--domain TEXT:DIR REQUIRED
+
+./build/Linux-Clang-Debug/tools/mxl-gst/mxl-gst-videosink [OPTIONS]
+
```
sviengkhou, line 162: I would add and example argument for Audio and Video, Video Only and Audio Only

docs/Tools.md
```diff
@@ -99,15 +99,58 @@ A binary that uses the gstreamer element 'videotestsrc' to produce video grains
...
+                              The json file which contains the Video NMOS Flow configuration
+  -a,     --audio-config-file TEXT
+                              The json file which contains the Audio NMOS Flow configuration
+  -s,     --samples-per-batch UINT [48]
+                              Number of audio samples per batch
+          --audio-offset INT [0]
+                              Audio sample offset in number of samples. Positive value means
+                              you are adding a delay (commit in the past).
+          --video-offset INT [0]
+                              Video grain offset in number of grains. Positive value means you
```
lapointejp: This is partly confusing to me. I think the app only works in progressive ?
If so perhaps we should use the word "frame" as a grain of interlaced is a field and not a frame.

I would argue that even if the app supported interlaced we probably wouldn't be accepting odd grain offsets, would we ?

Assuming I am correct and interlaced isn't supported by this app, perhaps we should elegantly exit with an error if the user supplies an interlaced flow definition ?

lapointejp (changes requested):

docs/Tools.md
mlefebvre1: Agreed, I've added throws on both testsrc and sink if we detect a flow with an interlace_mode other than "progressive". Also, I've updated the doc to refer to frames instead of grains.

jonasohland:
Should we rename `mxl-gst-videotestsrc` -> `mxl-gst-testsrc`?

pac-work (changes requested):
Hello, it seems like the timeout parameter in the sink is completely ignored now, so I have trouble making this read properly and it is hard to test more. 

Also just conceptually - I do not think that having 2 completely separate pipelines with 2 completely separate clocks sets a good example to how things should really be synchronized. Previous use of one pipeline felt much cleaner. But I can see how this can be subjective.

I think the timeout needs to be fixed, because this now really supports just the "totally linear usecase" and not the "asynchronously read data" usecase, while asynchronicity is in the heart of MXL requirements. I understand the timeout is not available for audio, but dropping it for video as well feels unnecessary.

I see also some strange behavior - I know that my data is definitely available with latency below 800 ms in domain with 3 seconds `history_duration` (latency visible from previous implementation). I specify `--video-read-delay 30` (> 1 s on the video test stream from MXL sources) so I should not even need the timeout to work, yet, I see grains being dropped. Something seems off with the timing, or I do not understand the parameters.

Maybe we can have a slack huddle / teams call so I can show you the behavior?

tools/mxl-gst/videosink.cpp
```diff
@@ -51,301 +43,484 @@ namespace
...
-                _autoaudiosink = gst_element_factory_make("autoaudiosink", "audio_sink");
-                if (!_autoaudiosink)
-                {
-                    throw std::runtime_error("Gstreamer: 'autoaudiosink' could not be created.");
-                }
-                gst_bin_add(GST_BIN(_pipeline), _autoaudiosink);
+            _appsrc = gst_bin_get_by_name(GST_BIN(_pipeline), "audioappsrc");
+            if (_appsrc == nullptr)
+            {
+                throw std::runtime_error("Gstreamer: 'appsink' could not be found in the pipeline.");
```
pac-work: This is not `appsink`, but `audioappsrc`.

mlefebvre1:
> Hello, it seems like the timeout parameter in the sink is completely ignored now, so I have trouble making this read properly and it is hard to test more.
> 
> Also just conceptually - I do not think that having 2 completely separate pipelines with 2 completely separate clocks sets a good example to how things should really be synchronized. Previous use of one pipeline felt much cleaner. But I can see how this can be subjective.
> 
> I think the timeout needs to be fixed, because this now really supports just the "totally linear usecase" and not the "asynchronously read data" usecase, while asynchronicity is in the heart of MXL requirements. I understand the timeout is not available for audio, but dropping it for video as well feels unnecessary.
> 
> I see also some strange behavior - I know that my data is definitely available with latency below 800 ms in domain with 3 seconds `history_duration` (latency visible from previous implementation). I specify `--video-read-delay 30` (> 1 s on the video test stream from MXL sources) so I should not even need the timeout to work, yet, I see grains being dropped. Something seems off with the timing, or I do not understand the parameters.
> 
> Maybe we can have a slack huddle / teams call so I can show you the behavior?

We can now use `--video-read-delay` to achieve similar behavior. There's no notification mechanism available for the continuous flow so I just wanted to have uniform options. 

Both pipelines are synchronized with `CLOCK_TAI` which is also what MXL uses by default, so everything should be in sync. Conceptually, they are two separate processing paths that can live independently.

mlefebvre1:
> I think it would be good to add:
> 
> ```
> CMD ["/app/mxl-gst-videotestsrc", "-d", "/domain", "-v", "/app/v210_flow.json", "-a", "/app/audio_flow.json"]
> ```
> 
> To `examples/Dockerfile.writer.txt` as part of this PR.
> 
> Moreover, as this demo supports multiple channels I think it would be more intuitive to set a "channel count": 2 in the `lib/tests/data/audio_flow.json`. But that would require updating tests too.

I can look into that

PR #216, opened 2025-11-06
Added call to be able to be notified of partial grain updates from discrete flows.

lapointejp:
With the current API, mxlFlowReaderGetGrainSlice would always return immediately as soon as there was a commit done on a given grain index. This adds an API call for users wishing to wait on specific slices being available.

I'm thinking maybe the mxlFlowReaderGetGrain may need to morph into the full grain being available, maybe up for a quick discussion.

@mlefebvre1 can you verify this fixes your issue ?

### DIFF
--- a/lib/include/mxl/flow.h
+++ b/lib/include/mxl/flow.h
@@ -205,7 +205,9 @@ extern "C"
     mxlStatus mxlFlowReaderGetInfo(mxlFlowReader reader, mxlFlowInfo* info);
 
     /**
-     * Accessors for a flow grain at a specific index
+     * Accessors for a flow grain at a specific index.
+     * This method is expected to wait until the full grain is available (or the timeout expires). For partial grain access use
+     * mxlFlowReaderGetGrainSlice()
      *
      * \param[in] reader A valid discrete flow reader.
      * \param[in] index The index of the grain to obtain
@@ -220,6 +222,25 @@ extern "C"
      */
     MXL_EXPORT
     mxlStatus mxlFlowReaderGetGrain(mxlFlowReader reader, uint64_t index, uint64_t timeoutNs, mxlGrainInfo* grain, uint8_t** payload);
+
+    /**
+     * Accessors for a flow grain at a specific index, with a minimum number of valid slices.
+     *
+     * \param[in] reader A valid discrete flow reader.
+     * \param[in] index The index of the grain to obtain
+     * \param[in] minValidSlices The minimum number of valid slices required in the returned grain.
+     * \param[in] timeoutNs How long should we wait for the slice (in nanoseconds)
+     * \param[out] grain The requested mxlGrainInfo structure.
+     * \param[out] payload The requested grain payload.
+     * \return The result code. \see mxlStatus
+     * \note Please note that this function can only be called on readers that
+     *      operate on discrete flows. Any attempt to call this function on a
+     *      reader that operates on another type of flow will result in an
+     *      error.
+     */
+    MXL_EXPORT
+    mxlStatus mxlFlowReaderGetGrainSlice(mxlFlowReader reader, uint64_t index, uint16_t minValidSlices, uint64_t timeoutNs, mxlGrainInfo* grain,
+        uint8_t** payload);
 
     /**
      * Non-blocking accessors for a flow grain at a specific index
--- a/lib/internal/include/mxl-internal/FlowManager.hpp
+++ b/lib/internal/include/mxl-internal/FlowManager.hpp
@@ -60,10 +60,14 @@ namespace mxl::lib
         /// \param[in] grainPayloadSize Size of the grain in host memory.  0 if the grain payload lives in device memory.
         /// \param[in] grainNumOfSlices Number of slices per grain.
         /// \param[in] grainSliceLengths Length of each slice in bytes.
+        /// \param[in] maxSyncBatchSizeHintOpt Optional max sync batch size hint.
+        /// \param[in] maxCommitBatchSizeHintOpt Optional max commit batch size hint
         ///
         std::unique_ptr<DiscreteFlowData> createDiscreteFlow(uuids::uuid const& flowId, std::string const& flowDef, mxlDataFormat flowFormat,
             std::size_t grainCount, mxlRational const& grainRate, std::size_t grainPayloadSize, std::size_t grainNumOfSlices,
-            std::array<std::uint32_t, MXL_MAX_PLANES_PER_GRAIN> grainSliceLengths);
+            std::array<std::uint32_t, MXL_MAX_PLANES_PER_GRAIN> grainSliceLengths,
+            std::optional<std::uint32_t> maxSyncBatchSizeHintOpt = std::nullopt,
+            std::optional<std::uint32_t> maxCommitBatchSizeHintOpt = std::nullopt);
 
         ///
         /// Create a new continuous flow together with its associated channel store and open it in read-write mode.
@@ -75,9 +79,13 @@ namespace mxl::lib
         /// \param[in] channelCount The number of channels in the flow.
         /// \param[in] sampleWordSize The size of one sample in bytes.
         /// \param[in] bufferLength The length of each channel buffer in samples.
+        /// \param[in] maxSyncBatchSizeHintOpt Optional max sync batch size hint.
+        /// \param[in] maxCommitBatchSizeHintOpt Optional max commit batch size hint
         ///
         std::unique_ptr<ContinuousFlowData> createContinuousFlow(uuids::uuid const& flowId, std::string const& flowDef, mxlDataFormat flowFormat,
-            mxlRational const& sampleRate, std::size_t channelCount, std::size_t sampleWordSize, std::size_t bufferLength);
+            mxlRational const& sampleRate, std::size_t channelCount, std::size_t sampleWordSize, std::size_t bufferLength,
+            std::optional<std::uint32_t> maxSyncBatchSizeHintOpt = std::nullopt,
+            std::optional<std::uint32_t> maxCommitBatchSizeHintOpt = std::nullopt);
 
         /// Open an existing flow by id.
         ///
--- a/lib/internal/include/mxl-internal/FlowOptionsParser.hpp
+++ b/lib/internal/include/mxl-internal/FlowOptionsParser.hpp
@@ -1,0 +1,64 @@
+// SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project.
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <picojson/picojson.h>
+
+namespace mxl::lib
+{
+    /**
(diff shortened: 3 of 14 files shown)


lib/internal/src/PosixDiscreteFlowReader.cpp
```diff
@@ -93,29 +93,30 @@ namespace mxl::lib
...
-                        continue;
-                    }
-                    else
-                    {
-                        status = MXL_ERR_TIMEOUT;
-                        break;
-                    }
+                    status = MXL_ERR_TIMEOUT;
+                    break;
                 }
```
KimonHoffmann, line 118: I know it was like this before, but wouldn't this be more readable like this:

```cpp
if (!waitUntilChanged(&flowInfo->discrete.syncCounter, previousSyncCounter, deadline))
{
    status = MXL_ERR_TIMEOUT;
    break;
}
```

lib/internal/src/PosixDiscreteFlowReader.cpp
```diff
@@ -93,29 +93,30 @@ namespace mxl::lib
                     {
                         auto const offset = in_index % flowInfo->discrete.grainCount;
                         auto const grain = _flowData->grainAt(offset);
-                        *out_grainInfo = grain->header.info;
-                        *out_payload = reinterpret_cast<std::uint8_t*>(&grain->header + 1);
-
-                        status = MXL_STATUS_OK;
-                        break;
+                        if (grain->header.info.validSlices >= in_minValidSlices)
```
KimonHoffmann: For resilience I would slightly change this check to:

```cpp
if (grain->header.info.validSlices >= std::min(in_minValidSlices, grain->header.info.totalSlices))
{
    // ...
}
```

lib/src/flow.cpp
```diff
@@ -307,14 +307,22 @@ mxlStatus mxlFlowReaderGetInfo(mxlFlowReader reader, mxlFlowInfo* info)
 extern "C"
 MXL_EXPORT
 mxlStatus mxlFlowReaderGetGrain(mxlFlowReader reader, uint64_t index, uint64_t timeoutNs, mxlGrainInfo* grainInfo, uint8_t** payload)
+{
+    return mxlFlowReaderGetGrainSlice(reader, index, 0, timeoutNs, grainInfo, payload);
```
KimonHoffmann: Is `0` the correct default here? With the change suggested above this should rather be something like `-1U`, shouldn't it?

KimonHoffmann (changes requested):
Looks good to me, with just a few small remarks.
Thank you for contributing this!

lib/src/flow.cpp
mlefebvre1: I tend to agree with you, but I'm wondering what's the expect behavior here. Should we return only once the full grain is valid or as soon as there's a commit to that specific grain (no matter how many validSlices).

vt-tv: > I tend to agree with you, but I'm wondering what's the expect behavior here. Should we return only once the full grain is valid or as soon as there's a commit to that specific grain (no matter how many validSlices).

the original intent was to return as soon as somehting new is available (a new slice, etc).

lapointejp: This was a concern for me to (and why this is a draft PR).
Unfortunately calling mxlFlowReaderGetGrain() is bound to be a race condition (if the writer writes data increasing the validSlices before we grab the syncCounter, we will miss an update).

I'm thinking the original call might be worth making it return when all slices are written and this sliced call would be a configured minimum number of slices. What do you think @KimonHoffmann and @vt-tv ?

lib/internal/include/mxl-internal/FlowManager.hpp
```diff
@@ -59,10 +59,14 @@ namespace mxl::lib
         /// \param[in] grainPayloadSize Size of the grain in host memory.  0 if the grain payload lives in device memory.
         /// \param[in] grainNumOfSlices Number of slices per grain.
         /// \param[in] grainSliceLengths Length of each slice in bytes.
+        /// \param[in] maxSyncBatchSizeHintOpt Optional max sync batch size hint.
+        /// \param[in] maxCommitBatchSizeHintOpt Optional max commit batch size hint
         ///
         std::unique_ptr<DiscreteFlowData> createDiscreteFlow(uuids::uuid const& flowId, std::string const& flowDef, mxlDataFormat flowFormat,
             std::size_t grainCount, mxlRational const& grainRate, std::size_t grainPayloadSize, std::size_t grainNumOfSlices,
-            std::array<std::uint32_t, MXL_MAX_PLANES_PER_GRAIN> grainSliceLengths);
+            std::array<std::uint32_t, MXL_MAX_PLANES_PER_GRAIN> grainSliceLengths,
```
KimonHoffmann, line 68: Should this really be passed by value?

lib/internal/include/mxl-internal/Instance.hpp
```diff
@@ -45,10 +45,11 @@ namespace mxl::lib
         /// Create a flow
         ///
         /// \param[in] flowDef The json flow definition according to the NMOS Flow Resource json schema
+        /// \param[in] options Additional options for flow creation
         /// \return The created flow resources in shared memory
         /// \throw std::runtime_error On any error (parse exception, shared memory conflicts, etc)
         ///
-        std::unique_ptr<FlowData> createFlow(std::string const& flowDef);
+        std::unique_ptr<FlowData> createFlow(std::string const& flowDef, std::string const& options = std::string());
```
KimonHoffmann: Could use `= std::string{}` or just plain `= {}`.

lib/internal/src/FlowOptionsParser.cpp
```diff
@@ -0,0 +1,77 @@
...
+            {
+                throw std::invalid_argument{"maxSyncBatchSizeHint must be a number."};
+            }
+
+            auto const v = maxSyncBatchSizeHintIt->second.get<double>();
+            if (v < 1)
+            {
+                throw std::invalid_argument{"maxSyncBatchSizeHint must be greater or equal to 1."};
+            }
+            _maxSyncBatchSizeHint = static_cast<std::uint32_t>(v);
```
KimonHoffmann: One additional validation we should do is:
```cpp
if ((_maxSyncBatchSizeHint % _maxCommitBatchSizeHint) != 0)
{
    throw std::invalid_argument{...};
}
```

For this reason it also makes sense to flip the order in which those two are parsed from the JSON string.

lib/src/flow.cpp
```diff
@@ -17,15 +17,15 @@ using namespace mxl::lib;
...
+mxlStatus mxlCreateFlow(mxlInstance instance, char const* flowDef, char const* options, mxlFlowInfo* flowInfo)
 {
     try
     {
         if (flowDef != nullptr)
         {
             if (auto const cppInstance = to_Instance(instance); cppInstance != nullptr)
             {
-                auto const flowData = cppInstance->createFlow(flowDef);
+                auto const flowData = cppInstance->createFlow(flowDef, options ? options : std::string());
```
KimonHoffmann: Feels a bit off due to the implicit conversion of the true case, maybe `std::string{options ? options : ""}` instead?!

lib/internal/include/mxl-internal/FlowManager.hpp
lapointejp: Sure.

lib/internal/src/FlowOptionsParser.cpp
lapointejp: Good catch.
I'll also add a test case for this.

lib/src/flow.cpp
lapointejp: Sure.

lib/internal/include/mxl-internal/FlowOptionsParser.hpp
```diff
@@ -0,0 +1,82 @@
...
+        /**
+         * Generic accessor for json fields.
+         *
+         * \param in_field The field name.
+         * \return The field value if found.
+         * \throw If the field is not found or T is incompatible
+         */
+        template<typename T>
+        [[nodiscard]]
+        T get(std::string const& field) const;
```
mlefebvre1, line 53: Why provide this method? Every possible options should be accessible from their appropriate method?

lib/internal/include/mxl-internal/FlowManager.hpp
mlefebvre1: I know this is not your doing, but the number of arguments is getting out of control. Maybe we should create a config struct?

lib/internal/include/mxl-internal/FlowOptionsParser.hpp
lapointejp: Leftover from copy-pasting the FlowParser :)
I'll remove it.

lib/src/flow.cpp
KimonHoffmann: > I'm thinking the original call might be worth making it return when all slices are written and this sliced call would be a configured minimum number of slices. 

Yeah my intuitive interpretation of the original call was also that it would return once all data of a grain has become available. This is why I suggested using `-1U` instead of `0U` , because `-1U` would be clamped down to whatever is the specific maximum of a flow.

lib/internal/include/mxl-internal/FlowManager.hpp
KimonHoffmann: Sorry this went in the wrong direction. I was not advocating for the optionals to become "by value" arguments, but rather for the `std::array<...>` to be come a "by reference" argument (unless there's a specific reason not to do that).

lib/include/mxl/flow.h
```diff
@@ -221,6 +221,25 @@ extern "C"
     MXL_EXPORT
     mxlStatus mxlFlowReaderGetGrain(mxlFlowReader reader, uint64_t index, uint64_t timeoutNs, mxlGrainInfo* grain, uint8_t** payload);
 
+    /**
+     * Accessors for a flow grain at a specific index, with a minimum number of valid slices.
+     *
+     * \param[in] reader A valid discrete flow reader.
+     * \param[in] index The index of the grain to obtain
+     * \param[in] minValidSlices The minimum number of valid slices required in the returned grain.
+     * \param[in] timeoutNs How long should we wait for the grain (in nanoseconds)
```
mlefebvre1: nitpick here but, we are actually waiting for the slices of that grain

lib/internal/include/mxl-internal/FlowManager.hpp
lapointejp: As discussed directly, let's file a separate ticket for addressing that type of thing and handle in a separate PR.

lib/include/mxl/flow.h
mlefebvre1: You did not apply the change to the right function

vt-tv:
Technically I can't merge until @KimonHoffmann  resolves his requested changes.

vt-tv:
> I'll do it right now. I'll also rebase the branch before. Could we please agree on rebasing branches on the target branch before merging, instead of merging the target branch to keep the history clean.

The GitHub offers a button to update the branch which I think is doing a merge.  @peterbrightwell is there a way to change the behaviour so it offers a rebase?

PR #219, opened 2025-11-06
Relocate internal fields from FlowInfo to a separate, private C++ type

KimonHoffmann:
`mxlFowInfo` currently serves two similar, but distinct purposes:
1. Describing and aiding operation on shared memory data structures.
2. Providing descriptive information about the data to the library user.

Serving both needs with one and the same data structure seemed logical at the time, since we originally anticipated users to more directly interact with the data in the shared memory segment. Since then the situation has shifted to a point, where interaction with the data in shared memory is performed exclusively (is it?) through API calls and never through direct manipulation of fields in shared memory, which makes exposing certain internals that are historically part of `mxlFlowInfo`, when passing around instances for descriptive purposes (by value!) unnecessary and inefficient, and effectively limits us in evolving the internal parts of the data structures in future iterations as we see fit.

So this PR separates those two concerns into dedicated data structures, one that only contains the descriptive aspects and one that holds the internals necessary for interacting with the data and facilitating synchronization between producers and consumers.

This PR just relocates the two fields in question.
A second PR will be filed after this PR has been integrated that makes use of the freedom afforded by the newly introduced separation by replacing the plain integer synchronization counter with something more elaborate.

--

The second major change implemented by this PR is a separation of the (remaining) `mxlFlowInfo` structure into two separate sub-structures, one that contains configuration metadata that is immutable over the lifetime of a flow and one that contains runtime metadata that reflects the state the flow is currently in.

For this
* `discrete.grainRate` and `continuous.sampleRate` are moved into the common config area as `grainRate`,
* `discrete.headIndex` and `continuous.headIndex` are moved into the runtime area as `headIndex`, and
* `common.lastWriteTime` and `common.lastReadTime` are moved into the runtime area.

Ultimately the C API has been updated to
1. allow querying `mxlFlowConfigInfo` and `mxlFlowRuntimeInfo` separately from a flow reader, and
2. only return a `mxlFlowConfigInfo` from `mxlCreateFlow()`, as the data in `mxlRuntimeInfo` is not yet populated with meaningful values at that point in time.

### DIFF
--- a/rust/mxl/src/grain/reader.rs
+++ b/rust/mxl/src/grain/reader.rs
@@ -4,8 +4,11 @@
 use std::{sync::Arc, time::Duration};
 
 use crate::{
-    Error, GrainData, Result,
-    flow::{FlowInfo, reader::get_flow_info},
+    Error, FlowConfigInfo, GrainData, Result,
+    flow::{
+        FlowInfo,
+        reader::{get_config_info, get_flow_info, get_runtime_info},
+    },
     instance::InstanceContext,
 };
 
@@ -27,8 +30,18 @@ impl GrainReader {
         self.destroy_inner()
     }
 
+    /// The whole FlowInfo is quite a chunk of data. Go for `get_config_info` or `get_runtime_info`
+    /// if they contain what you need.
     pub fn get_info(&self) -> Result<FlowInfo> {
         get_flow_info(&self.context, self.reader)
+    }
+
+    pub fn get_config_info(&self) -> Result<FlowConfigInfo> {
+        get_config_info(&self.context, self.reader)
+    }
+
+    pub fn get_runtime_info(&self) -> Result<mxl_sys::mxlFlowRuntimeInfo> {
+        get_runtime_info(&self.context, self.reader)
     }
 
     pub fn get_complete_grain<'a>(
--- a/rust/mxl/src/samples/reader.rs
+++ b/rust/mxl/src/samples/reader.rs
@@ -5,7 +5,10 @@ use std::sync::Arc;
 
 use crate::{
     Error, Result, SamplesData,
-    flow::{FlowInfo, reader::get_flow_info},
+    flow::{
+        FlowConfigInfo, FlowInfo,
+        reader::{get_config_info, get_flow_info, get_runtime_info},
+    },
     instance::InstanceContext,
 };
 
@@ -27,8 +30,18 @@ impl SamplesReader {
         self.destroy_inner()
     }
 
+    /// The whole FlowInfo is quite a chunk of data. Go for `get_config_info` or `get_runtime_info`
+    /// if they contain what you need.
     pub fn get_info(&self) -> Result<FlowInfo> {
         get_flow_info(&self.context, self.reader)
+    }
+
+    pub fn get_config_info(&self) -> Result<FlowConfigInfo> {
+        get_config_info(&self.context, self.reader)
+    }
+
+    pub fn get_runtime_info(&self) -> Result<mxl_sys::mxlFlowRuntimeInfo> {
+        get_runtime_info(&self.context, self.reader)
     }
 
     pub fn get_samples(&self, index: u64, count: usize) -> Result<SamplesData<'_>> {
--- a/rust/mxl/examples/flow-reader.rs
+++ b/rust/mxl/examples/flow-reader.rs
@@ -37,7 +37,7 @@ fn main() -> Result<(), mxl::Error> {
     let mxl_instance = mxl::MxlInstance::new(mxl_api, &opts.mxl_domain, "")?;
     let reader = mxl_instance.create_flow_reader(&opts.flow_id)?;
     let flow_info = reader.get_info()?;
-    if flow_info.is_discrete_flow() {
+    if flow_info.config.is_discrete_flow() {
         if opts.sample_batch_size.is_some() {
             return Err(mxl::Error::Other(
                 "Sample batch size is only relevant for \"continuous\" flows.".to_owned(),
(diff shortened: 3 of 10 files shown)


lib/include/mxl/flowinfo.h
```diff
@@ -75,18 +73,22 @@ extern "C"
          */
         uint32_t maxSyncBatchSizeHint;
 
-        /*
+        /**
+         * And indication, where the payload memory is located.
+         * \see mxlPayloadLocation
+         */
+        uint32_t payloadLocation;
```
garethsb, line 81: So this is now type-erased for extensibility, but `mxlPayloadLocation` values are still used? Does simply indicating "device memory" work for different vendor devices?

lib/include/mxl/flowinfo.h
```diff
@@ -75,18 +73,22 @@ extern "C"
...
+
+        /**
          * Device index (if payload is in device memory). -1 if on host memory.
          */
         int32_t deviceIndex;
```
garethsb, line 86: If we might use different domains for different host NUMA zones, could this also be used for that?

lib/include/mxl/flowinfo.h
KimonHoffmann: Yes, exactly.
I type erased it to make the signendess and width of the underlying type explicit and to make it consistent with other enumerated types we are already using.

lib/include/mxl/flowinfo.h
KimonHoffmann: You mean instead of keeping it fixed at `-1`?
I don't see a reason why not, but we need to put some thoughts into how to make that information actionable in a reasonable manner.

lib/include/mxl/flowinfo.h
```diff
@@ -75,18 +73,22 @@ extern "C"
          */
         uint32_t maxSyncBatchSizeHint;
 
-        /*
+        /**
+         * And indication, where the payload memory is located.
```
pedro-alves-ferreira, line 78: How about "An indication of where the payload memory is located"?

lib/include/mxl/flowinfo.h
KimonHoffmann: > Does simply indicating "device memory" work for different vendor devices?

Well, I didn't author this and I think it was just an initial sketch by Vincent on how we might go about this in the future. What I could propose is removing all enumerators except for `MXL_PAYLOAD_LOCATION_HOST_MEMORY`, which is the only one in active use at the moment, thereby leaving the complete number space available for a good encoding once we get around to actually implementing this. What do you think?

KimonHoffmann:
@dmf-mxl/mxl-maintainers  & @dmf-mxl/mxl-contributors
Should we take this PR a little further and
1. move `discrete.grainRate` and `continuous.sampleRate` into the common area as `updateRate
2. move `discrete.headIndex` and `continuous.headIndex` into the common area as `headIndex`
3. separate the common area into two structures, where one only contains immutable information about the flow and the other contains all the mutable fields (`lastWriteTime`, `lastReadTime`, a well as the newly moved `updateRate` and `headIndex`)?

For the two structures suggested in point 3 I would suggest something like `mxlFlowConfigInfo` and `mxlFlowRuntimeInfo` which we could allow querying separately in the API to cut down on unnecessary copying.

lib/include/mxl/flowinfo.h
vt-tv: Correct. it was a draft to express the need for location abstraction.  I agree that we could remove the unused enum values but keep the field for future extensibility.

lib/include/mxl/flowinfo.h
garethsb: 👍

lib/include/mxl/flowinfo.h
garethsb: Yes, this change LGTM. 👍

lib/include/mxl/flowinfo.h
```diff
@@ -171,8 +171,11 @@ extern "C"
             mxlContinuousFlowInfo continuous;
         };
 
-        /** Padding. Do not use. */
-        uint8_t reserved[3832];
+        /**
+         * Padding to get the the total size of this structure to 2048 bytes.
```
mlefebvre1, line 198: Any particular reason why we decided to reduce the size to 2048?

KimonHoffmann: Yes, because it is now split in two sub-structures of 2048 bytes each totaling in the original 4096 bytes.

rust/mxl/src/grain/reader.rs
```diff
@@ -27,10 +30,20 @@ impl GrainReader {
         self.destroy_inner()
     }
 
+    /// The whole FlowInfo is quite a chunk of data. Go for `get_config_info` or `get_runtime_info`
+    /// if they contain what you need.
     pub fn get_info(&self) -> Result<FlowInfo> {
```
pedro-alves-ferreira, line 35: Instead of advising the user not to use it, why not remove this altogether? If the user needs both the config and runtime info, they can call both functions separately.

rust/mxl/src/samples/reader.rs
```diff
@@ -27,10 +30,20 @@ impl SamplesReader {
         self.destroy_inner()
     }
 
+    /// The whole FlowInfo is quite a chunk of data. Go for `get_config_info` or `get_runtime_info`
+    /// if they contain what you need.
```
pedro-alves-ferreira, line 34: Same comment as for the writer.

rust/mxl/src/grain/reader.rs
KimonHoffmann: I contemplated the same thing on the C side of things, but resorted to keeping the unified call around for reasons of efficiency. But I totally agree that we should **heavily** discourage the use in favor of the more granular equivalents.

Fun fact: In the end it turned out that one of the tests actually had a more or less valid use case for the unified call that would have been made slightly less efficient by requiring the separated calls.

pedro-alves-ferreira: Good point. I'm OK with keeping it.

Issue #181, opened 2025-10-14
Fabrics: PR Step 2: API implementation


jonasohland:
Step 2 getting the fabrics PR merged. Happens after: #180

felixpou:
TSC
- @mlefebvre1 almost code complete, last details to be solved. 
- Probably open a PR next week

felixpou:
**TSC**
- AWS is ready to test, @jonasohland you can contact @Thomas-video 
- would be good to have Nvidia try it on their RoCEv2 setup. @garethsb ?
- As well as Intel. CBC could share their results, @felixpou will check.

felixpou:
TSC

- With the recent experience, PR has been merged!
- Do we need documentation?

felixpou:
TSC
- Transfer https://github.com/dmf-mxl/mxl/wiki/WIP:-MXL-Inter-host-memory-sharing-%E2%80%90-Design from the Wiki to the project /docs and clean it up
- [x] @jonasohland will do the PR

vt-tv:
TSC - @jonasohland  will perform the task above and close this ticket.

felixpou:
TSC
- PR https://github.com/dmf-mxl/mxl/pull/526 up for review

PR #227, opened 2025-11-07
feat: Add a getGrainInfo function to allow grainInfo inspection without having to open a grain for mutation.

mlefebvre1:
This relates to a specific need in the Fabrics API demo application. On the target side, we receive only the ring buffer index of the modified entry, not the complete grainIndex, because the immediate data is limited to 32 bits. We need to use the index field in mxlGrainInfo to recover the original grain index. Currently, the only available method to access the grain info is to open a grain for mutation (using the incomplete index). However, this approach has an unwanted side effect: it overwrites the index field with the provided index. To address this issue, I propose adding a new function, mxlFlowWriterGetGrainInfo, which would retrieve the grainInfo for a given index (supporting either grainIndex or ring buffer index) without modifying any fields.

This is needed by https://github.com/dmf-mxl/mxl/issues/181

### DIFF
--- a/lib/include/mxl/flow.h
+++ b/lib/include/mxl/flow.h
@@ -289,6 +289,21 @@ extern "C"
     mxlStatus mxlFlowReaderGetGrainNonBlocking(mxlFlowReader reader, uint64_t index, mxlGrainInfo* grain, uint8_t** payload);
 
     /**
+     * Get grain info for a given index. This is used to inspect the grain info without opening the grain for mutation.
+     *
+     * \param[in] writer A valid flow writer
+     * \param[in] index The index of the grain to obtain.
+     * \param[out] mxlGrainInfo The requested mxlGrainInfo structure.
+     * \return The result code. \see mxlStatus
+     * \note Please note that this function can only be called on writers that
+     *      operate on discrete flows. Any attempt to call this function on a
+     *      writer that operates on another type of flow will result in an
+     *      error.
+     */
+    MXL_EXPORT
+    mxlStatus mxlFlowWriterGetGrainInfo(mxlFlowWriter writer, uint64_t index, mxlGrainInfo* mxlGrainInfo);
+
+    /**
      * Open a grain for mutation.  The flow writer will remember which index is currently opened. Before opening a new grain
      * for mutation, the user must either cancel the mutation using mxlFlowWriterCancelGrain or mxlFlowWriterCommitGrain.
      *
--- a/lib/internal/src/PosixDiscreteFlowWriter.cpp
+++ b/lib/internal/src/PosixDiscreteFlowWriter.cpp
@@ -41,6 +41,18 @@ namespace mxl::lib
         throw std::runtime_error("No open flow.");
     }
 
+    mxlStatus PosixDiscreteFlowWriter::getGrainInfo(std::uint64_t in_index, mxlGrainInfo* out_grainInfo)
+    {
+        if (_flowData)
+        {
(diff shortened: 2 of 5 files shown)


lib/include/mxl/flow.h
```diff
@@ -237,6 +237,21 @@ extern "C"
     MXL_EXPORT
     mxlStatus mxlFlowReaderGetGrainNonBlocking(mxlFlowReader reader, uint64_t index, mxlGrainInfo* grain, uint8_t** payload);
 
+    /**
+     * Get grain info for a given index. This is used to inspect the grain info without opening the grain for mutation.
+     *
+     * \param[in] writer A valid flow writer
+     * \param[in] index The index of the grain to obtain.
+     * \param[out] mxlGrainInfo The requested mxlGrainInfo structure.
```
vt-tv: maybe be more explicit that this must not be null and the the field of the structure will be updated upon return? (we have similar comments elsewhere in the same file)

sviengkhou (approved):

lapointejp (approved):
LGTM :).

lib/internal/src/PosixDiscreteFlowWriter.hpp
```diff
@@ -40,6 +40,10 @@ namespace mxl::lib
         ///
         virtual mxlFlowInfo getFlowInfo() override;
 
+        ///
+        /// \see DiscreteFlowWriter::getGrainInfo
+        mxlStatus getGrainInfo(std::uint64_t in_index, mxlGrainInfo* out_grainInfo) final;
```
KimonHoffmann: Should have `virtual` and `override` qualifiers.
I would rather have a final qualifier on the class as whole rather on that specific method.

mlefebvre1: Unless I'm missing something, I would argue that Posix<Continous|Discrete>Flow<Writer|Reader> classes should all be final.

KimonHoffmann: Yes, indeed.

lib/internal/src/PosixDiscreteFlowWriter.hpp
```diff
@@ -33,32 +33,36 @@ namespace mxl::lib
...
         ///
         [[nodiscard]]
-        FlowData const& getFlowData() const final;
+        FlowData const& getFlowData() const override;
 
         ///
         /// \see FlowWriter::getFlowInfo
         ///
-        virtual mxlFlowInfo getFlowInfo() override;
+        mxlFlowInfo getFlowInfo() override;
```
KimonHoffmann: Please do not remove the `virtual` qualifiers *even though* they are redundant, because having a qualifier on the left hand side makes the virtual functions much easier to parse for the reader than scanning the non-aligned right hand side.

Or, to phrase it differently: The redundant qualifiers were put there intentionally.

KimonHoffmann (changes requested):

lib/internal/src/PosixDiscreteFlowWriter.cpp
```diff
@@ -45,7 +45,7 @@ namespace mxl::lib
     {
         if (_flowData)
         {
-            auto offset = in_index % _flowData->flowInfo()->discrete.grainCount;
+            auto offset = in_index % _flowData->flowInfo()->config.discrete.grainCount;
```
KimonHoffmann, line 48: Maybe make this a fixup to f11eb97?

KimonHoffmann (approved):
Looks good to me now, thank you!

PR #229, opened 2025-11-11
Update Rust bindings to reflect the new flow info structure internals

pac-work:


pac-work (commented):
Agreed with @KimonHoffmann to change it in the following sense (from our conversation on Slack):
> Quick Q … wouldn’t it be cleaner to more closely mirror the new structure of the C API.
> I.e.
> 1. move the polymorphic/variant handling that was previously part of the FlowInfo type into the/a new FlowConfigInfo type,
> 2. add the new FlowRuntimeInfo type, and
> 3. change FlowInfo to be an aggregate of FlowConfigInfo and FlowRuntimeInfo?

Will do these changes soon. Please wait with review and merging till when these are done.

rust/mxl/src/flow.rs
```diff
@@ -60,23 +65,87 @@ impl FlowInfo {
...
+    pub fn sample_rate(&self) -> Result<mxl_sys::mxlRational> {
+        let data_format = self.data_format();
+        if data_format != DataFormat::Audio {
+            return Err(Error::Other(format!(
+                "Flow format is {:?}, sample rate is only relevant for continuous flows.",
+                data_format
+            )));
+        }
+        Ok(self.0.grainRate)
+    }
```
KimonHoffmann, line 116: Do you think it makes sense to keep the distinction between `grain_rate` and `sample_rate`, now that we unified them on the C level?

KimonHoffmann (commented):
Except for the question below this looks good to me.
Thanks a lot for helping meow in updating the bindings!

rust/mxl/src/flow.rs
pac-work: I would say yes - the direct access is still available, if one does not want to perform any checking (`.grain_or_sample_rate(...)`). The `.grain_rate(...)` and `.sample_rate(...)` provide additional level of checking, which, if user does not have that in their code elsewhere, may be useful.

But no strong opinion on that. I can delete them if you want.

KimonHoffmann: Neither do I, so feel free to solve it as you see fit.

pac-work: I will keep it in then. These two should require close to 0 maintenance.

PR #230, opened 2025-11-11
Add signaling for continuous flows

KimonHoffmann:
Added futex based signaling for continuous flows according to the configured commit & sync batch sizes.
For symmetry with the discrete flow API the previous `getSamples()` function was renamed to `getSamplesNonBlocking()` and a new, blocking  variant was introduced as `getSamples()`.

The signaling mechanism takes the configured commit & sync batch size hints into account in order to only signal in appropriate intervals and not overshoot.

-- 

Since this PR is a bit a of a grab bag it probably makes sense to review the commits individually.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,7 @@ cmake_policy(SET CMP0135 NEW)
 
 option(MXL_ENABLE_FABRICS_OFI "Enable building the fabrics library with libfabric" OFF)
 
-set(mxl_VERSION 0.9.1)
+set(mxl_VERSION 0.10.0)
 
 if(DEFINED MXL_BUILD_NUMBER)
     string(APPEND mxl_VERSION ".${MXL_BUILD_NUMBER}")


lib/include/mxl/flow.h
```diff
@@ -331,6 +331,30 @@ extern "C"
...
+     * \note No guarantees are made as to how long the caller may
+     *      safely hang on to the returned range of samples without the
+     *      risk of these samples being overwritten.
+     */
+    MXL_EXPORT
+    mxlStatus mxlFlowReaderGetSamples(mxlFlowReader reader, uint64_t index, size_t count, uint64_t timeoutNs,
+        mxlWrappedMultiBufferSlice* payloadBuffersSlices);
+
+    /**
+     * Non-blocking ccessor for a specific set of samples across all channels
```
mlefebvre1: accessor

lib/internal/src/PosixContinuousFlowWriter.cpp
```diff
@@ -13,26 +14,44 @@ namespace mxl::lib
...
+        , _earlySyncThreshold{}
+        , _lastSyncSampleBatch{}
     {
         if (_flowData)
         {
-            return *_flowData;
+            auto const& commonFlowConfigInfo = _flowData->flowInfo()->config.common;
+
+            auto const commitBatchSize = std::max(commonFlowConfigInfo.maxCommitBatchSizeHint, 1U);
+            _syncBatchSize = ((std::max(commonFlowConfigInfo.maxSyncBatchSizeHint, 1U) + commitBatchSize - 1U) / commitBatchSize) * commitBatchSize;
```
mlefebvre1: I'm wondering if these 2 equations are correct.

My understanding is that _earlySyncThreshold is should be the the highest value for which another commit of size would exceed the batchSize. So for example if we have `maxCommitBatchSizeHint=3` and `maxSyncBatchSizeHint=10`: we would signal for any commit that gives us 8, 9 or 10 samples. This would mean that _earlySyncThreshold should be 8.

If, I plug this parametrization I get different values for syncBatchSize (12) and earlySyncThreshold (9). 

It's probably my understanding, but what's the intent here?

lib/internal/src/PosixContinuousFlowWriter.cpp
```diff
@@ -91,6 +116,31 @@ namespace mxl::lib
...
+    bool PosixContinuousFlowWriter::signalCompletedBatch() noexcept
+    {
+        auto const currentSyncSampleBatch = _currentIndex / _syncBatchSize;
+        if (currentSyncSampleBatch < _lastSyncSampleBatch)
+        {
+            return false;
+        }
+        if (currentSyncSampleBatch == _lastSyncSampleBatch)
+        {
+            // Signal now before overshooting the maximum the next time around
```
mlefebvre1, line 128: I'm confused by this comment. "// Signal now before overshooting the maximum the next time around", but then returns false, which does not signal completion. Should the comment be put after this if?

lib/internal/src/PosixContinuousFlowWriter.cpp
```diff
@@ -91,6 +116,31 @@ namespace mxl::lib
...
+            if ((currentSyncSampleBatch % _syncBatchSize) <= _earlySyncThreshold)
```
mlefebvre1: I think it should be `(_currentIndex % _syncBatchSize) <= _earlySyncThreshold`

mlefebvre1 (changes requested):

lib/internal/src/PosixContinuousFlowWriter.cpp
KimonHoffmann: Your general interpretation is correct, but -- to stick with the numbers you provided -- `_earlySyncThreshold` is supposed to be `7` not `8`, because it is used as an exclusive boundary, later on.

With regards to the radically different numbers you're observing:
The problem is that a sync batch size of `10` violates the requirement that `maxSyncBatchSizeHint` must be a multiple of `maxCommitBatchSizeHint`. So the code rounds it up to `12`, which is the nearest multiple. With the comment about `_earlySyncThreshold` above taken into account this then correctly ends up at `9`.

lib/internal/src/PosixContinuousFlowWriter.cpp
KimonHoffmann: Yes, good catch.

lib/internal/src/PosixContinuousFlowWriter.cpp
KimonHoffmann: I can understand the confusion. The comment is left unchanged from when the method looked slightly different and now ended up on the wrong branch of the conditional.

I'll make this more clear.

lib/internal/src/PosixContinuousFlowWriter.cpp
KimonHoffmann: Thinking about this a little bit more I don't think it is really necessary to enforce this multiplicity. So I could also change the code to not round up in those cases.

vt-tv (changes requested):
Thanks for the PR. a few minor comments : 
- minor version bump please.
- it would be great to use the new blocking call in the GST videosink.cpp tool. This would bring consistency between the video reader and audio readers.  If you think this should not part of this PR could you create a separate ticket for this change?

lib/internal/src/PosixContinuousFlowWriter.cpp
KimonHoffmann: I decided to make the change and not force `maxSyncBatchSizeHint` to be a multiple of `maxCommitBatchSizeHint`, as from the perspective of the writer there is no benefit that can be derived from this.

lib/include/mxl/flow.h
```diff
@@ -346,6 +346,30 @@ extern "C"
      *
      * \param[in] index The head index of the samples to obtain.
      * \param[in] count The number of samples to obtain.
+     * \param[in] timeoutNs How long to wait in nanoseconds for the range of
+     *      samples to become available.
+     * \param[out] payloadBuffersSlices A pointer to a wrapped multi buffer
```
garethsb, line 351: (Not specific to this PR)

One `mxlWrappedMultiBufferSlice` represents multiple wrapped buffer slices (where a wrapped buffer slice consists of 1 or 2 buffer slices), right? (And this is wrapped as in wrap-around in a circular buffer rather than wrapped as in a wrapper/abstraction, which is slightly confusing in itself!)

But all over the place a `mxlWrappedMultiBufferSlice` variable/parameter is called `payloadBuffersSlices` (plural buffers, plural slices)? Just reads a bit odd.

lib/internal/src/PosixContinuousFlowWriter.cpp
garethsb: _Is_ this resolved?

lib/internal/src/PosixContinuousFlowReader.hpp
```diff
@@ -28,41 +28,50 @@ namespace mxl::lib
...
+        [[nodiscard]]
         virtual bool isFlowValid() const override;
 
+    private:
+        /**
+         * Implementation of isFlowValid() that can be used by other methods
+         * that have previously asserted that we're operating on a valid flow.
+         */
+        [[nodiscard]]
+        bool isFlowValidImpl() const;
```
garethsb, line 67: this is just avoiding two things, a virtual function call, and a check of `_flowData`, right?
"previously asserted that we're operating on a valid flow" or "on valid flow data"?

lib/internal/src/PosixContinuousFlowReader.hpp
```diff
@@ -28,41 +28,50 @@ namespace mxl::lib
...
+
+        /**
+         * Implementation of the various forms of getSamples() that can also be
+         * used by other methods that habe previously asserted that we're
```
garethsb:
```suggestion
         * used by other methods that have previously asserted that we're
```

lib/internal/src/PosixContinuousFlowWriter.hpp
```diff
@@ -61,5 +71,13 @@ namespace mxl::lib
...
         /** The currently opened sample range head index. MXL_UNDEFINED_INDEX if no range is currently opened. */
         std::uint64_t _currentIndex;
+
+        /** Cached preprocessed copy of mxlCommonFlowInfo::maxSyncBatchSizeHint. */
+        std::uint32_t _syncBatchSize;
+        /** The threshold at which to signal even before _maxSyncBatchSizeHint is reached, in order to not exceed the maximum. */
+        std::uint32_t _earlySyncThreshold;
+
+        /** The last sample batch (as a factor of _maxSyncBatchSizeHint) that has been signaled. */
+        std::uint64_t _lastSyncSampleBatch;
```
garethsb, line 81: is this referring to the cached copy of `mxlCommonFlowInfo::maxSyncBatchSizeHint`?

```suggestion
        /** Cached preprocessed copy of mxlCommonFlowInfo::maxSyncBatchSizeHint. */
        std::uint32_t _syncBatchSize;
        /** The threshold at which to signal even before _syncBatchSize is reached, in order to not exceed the maximum. */
        std::uint32_t _earlySyncThreshold;

        /** The last sample batch (as a factor of _syncBatchSize) that has been signaled. */
        std::uint64_t _lastSyncSampleBatch;
```

garethsb (approved):
:octocat: I made it to the end :octocat: 

I have only read not run the code, but LGTM.

⬇️ Just a few comments on comments and naming!

lib/internal/src/PosixContinuousFlowReader.hpp
KimonHoffmann: Exactly. Some of the methods check for `isFlowValid()` after they already asserted that `_flowData` is valid, so this allows avoiding the redundant check, as well as properly inlining the call in the `.cpp` file.

lib/internal/src/PosixContinuousFlowWriter.hpp
KimonHoffmann: Correct, thank you.

lib/internal/src/PosixContinuousFlowReader.hpp
KimonHoffmann: :+1:

lib/include/mxl/flow.h
KimonHoffmann: Yeah, I'm also not entirely happy with the naming but struggled to find something that expresses what'S going on and combines nicely with or without `Mutable`.

> One mxlWrappedMultiBufferSlice represents multiple wrapped buffer slices (where a wrapped buffer slice consists of 1 or 2 buffer slices), right? (And this is wrapped as in wrap-around in a circular buffer rather than wrapped as in a wrapper/abstraction, which is slightly confusing in itself!)

Yes, the `Wrapped` part is supposed to express that it wraps around and thus each *slice* can consist of up to two "fragments". The `Multi` part (and the plural in `...Slices`) refer to the fact that this addresses a slice *across multiple channels* though, of which there may be arbitrarily many (as expressed by the `count` field IIRC).

Issue #177, opened 2025-10-14
Fabrics: partial-grain transfer support


jonasohland:
Implement the ability to transfer partial grains using the fabrics API. Should work for flows with alpha and asynchronously. Currently blocked by: #168 and maybe also #175 ?

Mammouth2727:
TSC : Needs testing. MAY work already.

felixpou:
TSC: need a reader and writer to be able to test

felixpou:
TSC

- Implemented, not tested yet. @jonasohland plans to work on this

felixpou:
TSC
- @jonasohland still need to test

jonasohland:
Tested, is missing support for sliced alpha, will do a PR.

felixpou:
TSC
- this one closed.
- created #592 for sliced alpha

PR #233, opened 2025-11-12
Add support of slices commit in videotestsrc

mlefebvre1:
This is required for testing this issue : https://github.com/dmf-mxl/mxl/issues/177

* Add an option to load video/Audio Flow option files in `videotestsrc`. 
* Use that option file to set the `maxSyncBatchSizeHint` field.
* Write frame into multiple batches using `maxSyncBatchSizeHint` to determine the number of commits. 
* Fix an issue where gstreamer frames were produced with a random offset that could span from -1/2 frame to 1/2 frame relative to the epoch frame timestamp. To fix this, we simply set the base time to an epoch frame timestamp, it also required the `do-timestamp` property on the testsrc element.

One issue I see is that `maxSyncBatchSizeHint` defaults to 1 when not set, which is very aggressive for audio (it will most likely not work). Could we change this to a higher default value for audio flows and maybe to a full grain for video?

### DIFF
--- a/docs/Tools.md
+++ b/docs/Tools.md
@@ -161,6 +161,7 @@ video_options.json
 
 ```video_options.json
 {
+  "maxCommitBatchSizeHint": 60,
   "maxSyncBatchSizeHint": 60
 }
 ```
@@ -169,6 +170,7 @@ audio_options.json
 
 ```audio_options.json
 {
+  "maxCommitBatchSizeHint": 512,
   "maxSyncBatchSizeHint": 512
 }
 ```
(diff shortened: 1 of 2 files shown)


sviengkhou (approved):

vt-tv (changes requested):
we should not default to a 1 sample batch size.

mlefebvre1:
> maxSync and maxCommit hints issues will be covered in another bug ticket created by @mlefebvre1

https://github.com/dmf-mxl/mxl/issues/244

PR #234, opened 2025-11-13
feat: Add audio support to mxl-gst looping_filesrc application

alanpawlak:
- Added support for audio pads in the `looping_filesrc` mxl-gst tool.
    - The GStreamer pipeline is now dynamic. When `decodebin` detects a pad, it calls a `cb_pad_added` function that adds and links the elements in the pipeline.
    - This currently supports only the first video and audio pad. Additional pads are discarded (we could handle them in the future).
- Added an audio thread and internal offset accounting for `gstBaseTime`, similar to the `videotestsrc` app.
- Added two Docker containers to the compose yaml: one for looping an audio file, and another for looping an audio/video file. The test files come from https://tsduck.io/streams/ and include:
    - Video/Audio: UK Freeview terrestrial network snapshot (Kitchen Nightmares) stream
    - Audio: Eutelsat Hot Bird 13.0°E, Mediaset, BBC World snapshot (Radio, Audio Only)
- Updated mount points in the media-function containers. The previous configuration didn’t work out of the box with `docker compose up --build` because the other containers couldn’t mount to `dev/shm/mxl` as the mxl folder didn't exist yet.
- Fixed typo in Dockerfile.reader.txt and added path for `lib/internal` which is required due to https://github.com/dmf-mxl/mxl/commit/ba5ff53c81b367ac64123cc11d9f0a362619153d

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -13,6 +13,12 @@ This example launches two containers simulating two different media functions ex
     docker exec -it examples-reader-media-function-1 /app/mxl-info -d /domain -f 5fbec3b1-1b0f-417d-9059-8b94a47197ed
     ```
 
+> **NOTE:** Out of the box, the setup works correctly only with docker.io. When using Docker CE, `docker compose up` may fail with:
+>
+> ```
+> invalid mount config for type "bind": bind source path does not exist: /dev/shm/mxl
+> ``` 
+
 ## Building
 
 Compile the whole MXL library and tools using the ```Linux-Clang-Release``` preset by invoking ```ninja all``` in the ```build/Linux-Clang-Release```  directory.  Then, in the "examples" directory, run:


vt-tv:
Thanks for your contribution!

The example docker compose file used an 'init container"' that was configured to run before other containers in the compose file. this allowed the proper path to be created (/dev/shm/domain) to be created.  The reason behind this subfolder is to demonstrate that we don't have to share ALL /dev/shm to make MXL work.  removing the domain path may send the wrong message.   Is this init-container mechanism now broken?

tools/mxl-gst/looping_filesrc.cpp
```diff
@@ -345,29 +587,77 @@ class LoopingFilePlayer
...
         {
             auto sample = gst_app_sink_try_pull_sample(GST_APP_SINK(appSinkVideo), 100'000'000);
             if (sample)
             {
-                uint64_t grainIndex = mxlGetCurrentIndex(&videoGrainRate);
-                if (lastVideoGrainIndex == 0)
+                auto buffer = gst_sample_get_buffer(sample);
+                if (buffer)
                 {
+                    auto pts = GST_BUFFER_PTS(buffer);
```
mlefebvre1: I know it was already like this, but for safety reason you should probably increase the refcount of `buffer` while you are using it. 

https://gstreamer.freedesktop.org/documentation/gstreamer/gstsample.html?gi-language=c#gst_sample_get_buffer


```
The buffer remains valid as long as sample is valid. If you need to hold on to it for longer than that, take a ref to the buffer.
```

tools/mxl-gst/looping_filesrc.cpp
```diff
@@ -345,29 +587,77 @@ class LoopingFilePlayer
...
-                }
 
-                lastVideoGrainIndex = grainIndex;
+                    if (lastVideoGrainIndex == 0)
+                    {
+                        lastVideoGrainIndex = grainIndex;
+                    }
+                    else if (grainIndex != lastVideoGrainIndex + 1)
+                    {
+                        MXL_WARN("Video skipped grain index. Expected {}, got {}", lastVideoGrainIndex + 1, grainIndex);
```
mlefebvre1: To avoid gaps, you should probably generate the missing grains as invalid by setting the `grainInfo` flags field to `MXL_GRAIN_FLAG_INVALID`.

tools/mxl-gst/looping_filesrc.cpp
```diff
@@ -377,37 +667,116 @@ class LoopingFilePlayer
...
-                    if (mxlFlowWriterCommitGrain(flowWriterVideo, &gInfo) != MXL_STATUS_OK)
+                    if (lastAudioGrainIndex == 0)
+                    {
+                        lastAudioGrainIndex = grainIndex;
+                    }
+                    else if (grainIndex != lastAudioGrainIndex + 1)
                     {
-                        MXL_ERROR("Failed to open grain at index '{}'", grainIndex);
-                        break;
+                        MXL_WARN("Audio skipped grain index. Expected {}, got {}", lastAudioGrainIndex + 1, grainIndex);
```
mlefebvre1: Same comment as video but instead generate a silence. I don't think you have to write a fade, but at least there won't be any gaps.

tools/mxl-gst/looping_filesrc.cpp
```diff
@@ -377,37 +667,116 @@ class LoopingFilePlayer
...
-                    uint8_t* mxl_buffer = nullptr;
-
-                    /// Open the grain for writing.
-                    if (mxlFlowWriterOpenGrain(flowWriterVideo, grainIndex, &gInfo, &mxl_buffer) != MXL_STATUS_OK)
+                    auto pts = GST_BUFFER_PTS(buffer);
+                    if (!_internalOffset)
                     {
-                        MXL_ERROR("Failed to open grain at index '{}'", grainIndex);
-                        break;
+                        _internalOffset = mxlGetTime() - (pts + gstBaseTime);
```
mlefebvre1: Did you validate that the internal offset of the audio appsink and video appsink is the same value?

alanpawlak:
> Thanks for your contribution!
> 
> The example docker compose file used an 'init container"' that was configured to run before other containers in the compose file. this allowed the proper path to be created (/dev/shm/domain) to be created. The reason behind this subfolder is to demonstrate that we don't have to share ALL /dev/shm to make MXL work. removing the domain path may send the wrong message. Is this init-container mechanism now broken?

@vt-tv I might be doing something wrong but when I tried using the containers out of the box according to the README.md I faced the following error (just checked on the main branch)

```
docker compose build
```
```
docker compose up
```

Error:
```
❯ docker compose up
[+] Running 2/2
 ✔ init Pulled                                                                                                                                                                            2.1s 
   ✔ 2d35ebdb57d9 Already exists                                                                                                                                                          0.0s 
[+] Running 1/0
 ⠋ Container examples-reader-media-function-1  Recreate                                                                                                                                   0.0s 
 ✔ Container examples-writer-media-function-1  Created                                                                                                                                    0.0s 
Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /dev/shm/mxl
```

It seems to me that the `mxl` folder must already exist before `docker compose up` is run.  However, if I run `docker compose up --build init` followed by `docker compose up` everything works. Perhaps I should update readme file instead?

tools/mxl-gst/looping_filesrc.cpp
alanpawlak: My bad, I overlooked that!

tools/mxl-gst/looping_filesrc.cpp
alanpawlak: Do I really need to ref the buffer in this case? I am only using it within the scope of the sample. I understand that I would need to increase the reference count if I wanted to keep the buffer beyond the sample’s lifetime, but I do not use it after dereferencing the sample.

mlefebvre1: That's true. Resolving this comment.

tools/mxl-gst/looping_filesrc.cpp
alanpawlak: I resolved this in my latest commit. I can implement a fade in the future. Do we have any plans for how we want to approach it? I have been thinking that we would need to delay processing by the length of the buffer and store a temporary buffer so we can apply a clean fade-out/fade-in.

mlefebvre1:
Looks good.

mlefebvre1 (approved):

examples/Dockerfile.reader.txt
```diff
@@ -10,8 +10,9 @@ RUN apt update && apt-get -y install\
         patchelf && rm -rf /var/lib/apt/lists/*
 
 COPY build/Linux-Clang-Release/lib/*.so* /app/
+COPY build/Linux-Clang-Release/lib/internal/*.so* /app/
 COPY build/Linux-Clang-Release/tools/mxl-info/mxl-info /app/
 
 RUN patchelf --set-rpath /app mxl-info
 
-CMD["sh", "-c", "while true; do /app/mxl-info -d /domain -l || true; sleep 2; done"]
+CMD ["sh", "-c", "while true; do /app/mxl-info -d /shared_memory/mxl -l || true; sleep 2; done"]
```
lapointejp: I am skeptical about this one - I think we should keep using "domain" in a sample as this maps the intended concept within MXL. I worry that using the name "shared_memory" could become confusing to a newcomer. This would also impact the other Dockerfiles.

tools/mxl-gst/looping_filesrc.cpp
```diff
@@ -65,18 +174,34 @@ class LoopingFilePlayer
 
     ~LoopingFilePlayer()
     {
-        // Join threads if they were created
+        // Join video threads if they were created
```
lapointejp: Only one thread, so should be singular.
But also I would be an advocate of removing these comments given the intent is quite clear from the code.

tools/mxl-gst/looping_filesrc.cpp
```diff
@@ -345,73 +586,336 @@ class LoopingFilePlayer
...
+                    else
+                    {
+                        GstMapInfo map_info;
+                        if (gst_buffer_map(buffer, &map_info, GST_MAP_READ))
+                        {
+                            /// Open the grain.
+                            mxlGrainInfo gInfo;
+                            uint8_t* mxl_buffer = nullptr;
+
+                            /// Open the grain for writing.
```
lapointejp: Seems a redundant comment with line 720.

tools/mxl-gst/looping_filesrc.cpp
```diff
@@ -345,73 +586,336 @@ class LoopingFilePlayer
...
+            }
+            else
+            {
+                if (gst_app_sink_is_eos(GST_APP_SINK(appSinkVideo)))
+                {
+                    MXL_WARN("appSinkVideo reached EOS");
                 }
+                else
+                {
+                    MXL_WARN("No sample received from appSinkVideo within timeout");
```
lapointejp, line 756: sample -> grain ?

tools/mxl-gst/looping_filesrc.cpp
```diff
@@ -345,73 +586,336 @@ class LoopingFilePlayer
...
+                    if (!sampleIndex)
                     {
-                        MXL_ERROR("Failed to open grain at index '{}'", grainIndex);
-                        break;
+                        sampleIndex = gstSampleIndex;
+                        MXL_INFO("audioThread: Set initial sample index to {} (GST_BUFFER_PTS={} ns)", *sampleIndex, adjGstBufferPts);
                     }
 
-                    gst_buffer_unmap(buffer, &map);
+                    // Verify that we didnt miss any samples
```
lapointejp:
```suggestion
                    // Verify that we didn't miss any samples
```

examples/Dockerfile.reader.txt
alanpawlak: I am happy with leaving it as domain, however, as I mentioned in my comment above the current docker setup does not work out of the box. Do you suggest I should update the documentation instead?

tools/mxl-gst/looping_filesrc.cpp
alanpawlak: This should stay as sample because it refers to GStreamer sample, not the MXL sample/grain

vt-tv (approved):
LGTM. Thanks!

PR #238, opened 2025-11-13
Fix #237 by extending the API and synchronizing the semantics of the various methods

### LINKED ISSUE #237: Blocking and non-blocking grain access functions have gotten out of sync
### Description

With the recent introduction of the ability to wait for partial grains the interpretation of when a grain is considered being available has gotten out of sync between `mxlFlowReaderGetGrain` and `mxlFlowReaderGetGrainNonBlocking`.

The former now considers a grain to be available, when *all slices* are available while the latter is left with the previous interpretation of a grain being considered available, when *anything* is available.

### Expected behavior

The API should be updated for constancy, which would entail:
1. Changing the availability interpretation of `mxlFlowReaderGetGrainNonBlocking` to match that of `mxlFlowReaderGetGrain`.
2. Introducing `mxlFlowReaderGetGrainSliceNonBlocking` with semantics matching `mxlFlowReaderGetGrainSlice`

Ideally the underlying implementations of `mxlFlowReaderGetGrainSliceNonBlocking` and `mxlFlowReaderGetGrainNonBlocking` should be structured analogous to the implementations of their blocking counterparts, where the whole grain function defers to the partial grain implementation.

KimonHoffmann:


vt-tv (approved):
looks good to me. thanks

lib/internal/src/PosixDiscreteFlowReader.hpp
```diff
@@ -49,13 +49,30 @@ namespace mxl::lib
...
+        /**
+         * Implementation of isFlowValid() that can be used by other methods
+         * that have previously asserted that we're operating on a valid flow.
+         */
+        [[nodiscard]]
+        bool isFlowValidImpl() const;
+
+        /**
+         * Implementation of the various forms of getGrain() that can also be
+         * used by other methods that habe previously asserted that we're
```
garethsb:
```suggestion
         * used by other methods that have previously asserted that we're
```

lib/internal/src/PosixDiscreteFlowReader.hpp
```diff
@@ -49,13 +49,30 @@ namespace mxl::lib
...
         [[nodiscard]]
         virtual bool isFlowValid() const override;
 
+    private:
+        /**
+         * Implementation of isFlowValid() that can be used by other methods
+         * that have previously asserted that we're operating on a valid flow.
+         */
+        [[nodiscard]]
+        bool isFlowValidImpl() const;
```
garethsb, line 67: ⬇️ same query as https://github.com/dmf-mxl/mxl/pull/230/files#r2545657551

garethsb (approved):
LGTM, just ⬇️ comments on the comments copied from _PosixContinuousFlowReader.hpp_ to  _PosixDiscreteFlowReader.hpp_

PR #247, opened 2025-11-20
Modernize and clean up the mxl-info and GStreamer tools

KimonHoffmann:


docs/Tools.md
```diff
@@ -201,28 +201,15 @@ OPTIONS:
...
-                              How far in the past/future to read (in video frames). Positive
-                              value means you are delaying the read
-          --timeout-mode [0]  In timeout mode, the sink uses blocking methods with timeout
-                              instead of reading with a defined delay. Currently only video
-                              supports this mode and --video-offset is used for both the
-                              timeout value and playback offset.
-
+                              Audio channels to listen.
+          --av-delay INT [0]  The time in nanoseconds, by which to delay the audio relative to
+                              video. A positive value means you are delaying audio, a negative
```
lapointejp, line 214: When/if we add ANC, expectation would be to keep Video and ANC affected by the same offset value ?

tools/mxl-gst/videosink.cpp
```diff
@@ -32,661 +33,688 @@
...
+                throw std::runtime_error{"Gstreamer pipeline could not be created."};
             }
 
-            _appsrc = gst_bin_get_by_name(GST_BIN(_pipeline), "videoappsrc");
-            if (_appsrc == nullptr)
+            _appSource = ::gst_bin_get_by_name(GST_BIN(_pipeline), "appsource");
+            if (_appSource == nullptr)
             {
-                throw std::runtime_error("Gstreamer: 'videoappsrc' could not be found in the pipeline.");
+                throw std::runtime_error{"Wellknown application source element could not be found in the gstreamer pipeline."};
```
lapointejp:
```suggestion
                throw std::runtime_error{"Well-known application source element could not be found in the gstreamer pipeline."};
```

tools/mxl-gst/videosink.cpp
```diff
@@ -32,661 +33,688 @@
...
             MXL_INFO("Starting discrete flow reading at rate {}/{} and slices per batch {}", rate.numerator, rate.denominator, slicesPerBatch);
-            auto timeoutNs = discreteFlowTimeout(timeoutMode, gstPipeline._config.offset);
 
-            auto index = mxlGetCurrentIndex(&rate);
-            mxlStatus ret;
-            std::uint16_t validSlices = slicesPerBatch;
+            // The index that corresponds to the current time. We are reading
+            // frames that correspond to the index that is readOffset
+            // nanoseconds in the past and deliver them as the next frame to the
+            // gesteremaer pipeline.
```
lapointejp:
```suggestion
            // gstreamer pipeline.
```

tools/mxl-gst/videosink.cpp
```diff
@@ -32,661 +33,688 @@
...
         }
 
     private:
-        std::uint64_t discreteFlowTimeout(bool timeoutMode, std::int64_t offset)
+        /**
+         * This default constructor is private, because the only reason why it
+         * exists is that it can be used by the publicly accessible constructor
+         * to delegate to.
+         * This delegation is beneficial, because it implies that the destructor
+         * will be invoked when the delgating constructor throws and exception,
```
lapointejp:
```suggestion
         * will be invoked when the delegating constructor throws an exception,
```

tools/mxl-gst/videotestsrc.cpp
```diff
@@ -17,800 +19,874 @@
...
-    MxlWriter(std::string const& domain, std::string const& flow_descriptor, std::string const& flowOptions)
-    {
-        _instance = mxlCreateInstance(domain.c_str(), "");
-        if (_instance == nullptr)
-        {
-            throw std::runtime_error("Failed to create MXL instance");
+            _appSink = ::gst_bin_get_by_name(GST_BIN(_pipeline), "appsink");
+            if (_appSink == nullptr)
+            {
+                throw std::runtime_error{"Wellknown application sink element could not be found in the gstreamer pipeline."};
```
lapointejp:
```suggestion
                throw std::runtime_error{"Well-known application sink element could not be found in the gstreamer pipeline."};
```

tools/mxl-gst/videotestsrc.cpp
```diff
@@ -17,800 +19,874 @@
...
+                                MXL_ERROR("Failed to commit invalid grain at index '{}'", actualGrainIndex);
+                                break;
+                            }
 
-                    GstMapInfo map_info;
-                    if (gst_buffer_map(gstBuffer, &map_info, GST_MAP_READ))
+                            grainIndex += 1U;
+                        }
+                    }
+                    else // gsteamer index matches expected index
```
lapointejp:
```suggestion
                    else // gstreamer index matches expected index
```

tools/mxl-gst/videotestsrc.cpp
```diff
@@ -17,800 +19,874 @@
...
+                    if (sampleIndex == MXL_UNDEFINED_INDEX)
                     {
-                        MXL_ERROR("Failed to open samples at index '{}'", actualSampleIndex);
-                        break;
+                        sampleIndex = gstSampleIndex;
+                        MXL_INFO("ContinuousFlow: Set initial sample index to {} (bufferTs={} ns)", sampleIndex, bufferTs);
                     }
-
-                    for (uint64_t chan = 0; chan < payloadBuffersSlices.count; ++chan)
+                    // Verify that we didnt miss any samples
```
lapointejp:
```suggestion
                    // Verify that we didn't miss any samples
```

tools/mxl-gst/videotestsrc.cpp
```diff
@@ -17,800 +19,874 @@
...
-                                    auto src = map_info.data + addrOffset;
-                                    ::memcpy(dst, src, fragment.size);
-                                    addrOffset += fragment.size;
+                                    auto const dst = static_cast<std::uint8_t*>(fragment.pointer) + (chan * payloadBuffersSlices.stride);
+                                    ::memset(dst, 0, fragment.size); // fill with silence
                                 }
                             }
                         }
+                    }
+                    else // gsteamer index matches expected index
```
lapointejp:
```suggestion
                    else // gstreamer index matches expected index
```

docs/Tools.md
KimonHoffmann: TBH I have not considered that, yet. Do you think this makes sense or would you rather have a separate offset for ANC then?

mlefebvre1: I would rather have a separate offset for each essence. It's more flexible and you can achieve the same workflow.

tools/mxl-gst/sink.cpp
```diff
@@ -32,661 +33,688 @@
...
         {
             if (_pipeline)
             {
-                gst_element_set_state(_pipeline, GST_STATE_NULL);
-                gst_object_unref(_pipeline);
+                ::gst_element_set_state(_pipeline, GST_STATE_NULL);
+                ::gst_object_unref(_pipeline);
             }
-            if (_appsrc)
+            if (_appSource)
```
mlefebvre1, line 116: Why this renaming? Why not follow the name of the gstreamer element?

tools/mxl-gst/sink.cpp
```diff
@@ -32,661 +33,688 @@
...
+                        auto pipeline = VideoPipeline{videoConfig};
+                        reader.run(pipeline, readDelay);
 
-                    MXL_INFO("Video pipeline finished");
-                    return 0;
+                        MXL_INFO("Video pipeline finished");
+                    }
+                    catch (std::exception const& e)
+                    {
+                        MXL_ERROR("Error while processing audio pipeline: {}", e.what());
```
mlefebvre1: video

docs/Tools.md
KimonHoffmann: My proposal would be that instead of separating the two as it was before, we'd introduce an additional `--playback-delay` parameter, instead that applies equally to all essences.

That way it is easy to add a delay to the pipeline without having to take special care to keep the audio and video essences in sync, while still retaining the ability to offset them against each other if that should be necessary.

KimonHoffmann:
The macOS issue has been solved in the most recent push.
Turnet out to be much simpler to address, as expected.

tools/mxl-gst/sink.cpp
```diff
@@ -0,0 +1,782 @@
...
+
+            _appSource = ::gst_bin_get_by_name(GST_BIN(_pipeline), "appsource");
+            if (_appSource == nullptr)
+            {
+                throw std::runtime_error{"Well-known application source element could not be found in the gstreamer pipeline."};
+            }
+
+            ::g_object_set(G_OBJECT(_appSource), "format", GST_FORMAT_TIME, nullptr);
+
+            if (auto const clock = GST_CLOCK(::g_object_new(GST_TYPE_SYSTEM_CLOCK, "name", "mxl-tai-clock", nullptr)); clock != nullptr)
```
mlefebvre1, line 151: Can we have a comment explaining why we can't directly extract the clock from the pipeline (using `gst_pipeline_get_clock(GST_PIPELINE(_pipeline))`) and modify that one instead of creating a new one?

tools/mxl-gst/sink.cpp
```diff
@@ -0,0 +1,782 @@
...
+
+    class AudioPipeline : public GstreamerPipeline
+    {
+    public:
+        AudioPipeline(AudioPipelineConfig const& config)
+            : GstreamerPipeline{}
+            , _config{config}
+            , _audioInfo{}
+        {
+            MXL_INFO("Creating audio piepleine with config: {}", _config.display());
```
mlefebvre1: pipeline

tools/mxl-gst/sink.cpp
```diff
@@ -0,0 +1,782 @@
...
+            , _config{config}
+        {
+            MXL_INFO("Creating video pipeline with config: {}", _config.display());
+
+            auto pipelineDesc = fmt::format(
+                "appsrc name=appsource is-live=true ! "
+                "video/x-raw,format=v210,width={},height={},framerate={}/{} ! "
+                "videoconvert ! "
+                "videoscale ! "
+                "queue ! "
```
mlefebvre1: I know it was there before, but I think we can get rid of the queue

tools/mxl-gst/sink.cpp
```diff
@@ -0,0 +1,782 @@
...
+                throw std::runtime_error{"Attempt to start uninitialized pipeline."};
+            }
+
+            // Start playing
+            ::gst_element_set_state(_pipeline, GST_STATE_PLAYING);
+            _mxlBaseTime = ::mxlGetTime();
+            MXL_INFO("Staring pipeline with base time: {} ns", _mxlBaseTime);
+        }
+
+        void pushSample(GstBuffer* buffer, std::uint64_t now) noexcept
```
mlefebvre1: Additional clean-up, could be renamed `pushBuffer`

tools/mxl-gst/sink.cpp
```diff
@@ -0,0 +1,782 @@
...
+            auto const windowSize = std::max(_configInfo.common.maxSyncBatchSizeHint, 16U);
+
+            MXL_INFO("Starting continuous flow reading at rate {}/{} and batch size {}", rate.numerator, rate.denominator, windowSize);
+
+            // The index that corresponds to the current time. We are reading
+            // frames that correspond to the index that is readOffset
+            // nanoseconds in the past and deliver them as the next frame to the
+            // gesteremaer pipeline.
+            auto const startTime = ::mxlGetTime();
+            auto const readDelayGrains = durationInGrains(rate, readDelay);
```
mlefebvre1: 1 grain = 1 sample?

tools/mxl-gst/sink.cpp
```diff
@@ -0,0 +1,782 @@
...
+        {
+            if (_configInfo.common.format != MXL_DATA_FORMAT_AUDIO)
+            {
+                throw std::domain_error{"Attempt to feed a gstreamer audio pipeline from a non-audio MXL flow."};
+            }
+
+            gstPipeline.start();
+
+            auto const rate = _configInfo.common.grainRate;
+            auto const windowSize = std::max(_configInfo.common.maxSyncBatchSizeHint, 16U);
```
mlefebvre1: For educational purposes, it could be worth explaining why a minimum of 16 samples is used here.

tools/mxl-gst/sink.cpp
```diff
@@ -0,0 +1,782 @@
...
+
+                        MXL_INFO("Video pipeline finished");
+                    }
+                    catch (std::exception const& e)
+                    {
+                        MXL_ERROR("Error while processing audio pipeline: {}", e.what());
+                    }
+                    catch (...)
+                    {
+                        MXL_ERROR("Encountered unknown error while processing audio pipeline.");
```
mlefebvre1: video

tools/mxl-gst/sink.cpp
KimonHoffmann: Because of our `camelCase` naming convention, which makes less sense with an abbreviated identifier.

tools/mxl-gst/sink.cpp
KimonHoffmann: This is an intentional change as explained in the commit message of [this](../commit/a1eca6b8efeae9ac11dc55a9c3a8ea31b0b14b52) commit.

The reasoning goes like this:
The clock returned by `gst_pipeline_get_clock()` is not guaranteed to be of type `GstSystemClock`, which would make setting the `clock-type` a noop. So we create a new clock with the necessary type and make the pipeline use that.

tools/mxl-gst/sink.cpp
KimonHoffmann: I'll have a look, thanks for the hint.

tools/mxl-gst/sink.cpp
KimonHoffmann: Good idea. :+1:

tools/mxl-gst/sink.cpp
KimonHoffmann: Yes. This has been adopted as the universal nomenclature across MXL during the recent unification of the different `FlowInfo` variants.

tools/mxl-gst/sink.cpp
KimonHoffmann: But to your question: Yeah, I'll add a comment explaining this 😄

tools/mxl-gst/sink.cpp
KimonHoffmann: I adapted the comment to make it more clear that we are dealing with a range of samples.

tools/mxl-gst/sink.cpp
KimonHoffmann: My tests all went fine without it so I'll remove it for now. :+1:

docs/Tools.md
KimonHoffmann: Done.

tools/mxl-gst/sink.cpp
```diff
@@ -0,0 +1,788 @@
...
+        }
+
+        [[nodiscard]]
+        constexpr GstElement* getAppSource() noexcept
+        {
+            return _appSource;
+        }
+
+        [[nodiscard]]
+        constexpr GstElement const* getAppSource() const noexcept
```
pac-work, line 172: This function does not seem to be used now.

tools/mxl-gst/testsrc.cpp
```diff
@@ -0,0 +1,939 @@
...
+            {
+                throw std::runtime_error{fmt::format("Failed to create flow with status code {}", static_cast<int>(ret))};
+            }
+            _flowOpened = true;
+
+            // Create the flow writer
+            auto const flowID = to_string(uuids::uuid{_configInfo.common.id});
+            if (::mxlCreateFlowWriter(_instance, flowID.c_str(), "", &_writer) != MXL_STATUS_OK)
+            {
+                throw std::runtime_error{"Failed to create flow writer"};
```
lapointejp:
```suggestion
                throw std::runtime_error{"Failed to create flow writer."};
```

docs/Tools.md
lapointejp: I'm okay with it as is for now with the av-offset setting, but we might need to split this into three settings in the future when ANC is also in the sample.

vt-tv (approved):
LGTM.  OK to merge?

tools/mxl-gst/sink.cpp
KimonHoffmann: It's used in the derived `AudioPipeline` class.
If you're referring specifically to the const-qualified overload I'd like to keep it around, because I find it good style to provide both in such cases.

KimonHoffmann:
> You also commented in Slack that the gstreamer tools need to use an API that is internal to the SDK. Is this covered by this PR? if not, can you create a new ticket for this issue?

There should already be an issue for that, i.e.
1. Introducing those APIs that expose the necessary utility functionality on the C level
2. Making use of these APIs across all examples and tools instead of the currently used internal APIs.

tools/mxl-gst/testsrc.cpp
KimonHoffmann: Thank you for the suggestion!
Updated this occurrence together with the other two above and below.

tools/mxl-gst/sink.cpp
pac-work: I was referring to the const version, but if that is conscious decision, why not...

KimonHoffmann:
> In my eyes, it is a bit unfortunate that we only demonstrate linear reading as if "it was SDI cable", although MXL can do so much more and fits well paradigm of non-linear processing.

As I said I'm open to re-adding a "auto adjust mode" in a future PR.

> That is why we were originally pushing for signalling of data availability in the first place. 

The signaling and blocking calls still have their raison d'être in the linear scenario, as the calling application can choose its operating point and efficiently sleep for grains to become available (hopefully at a regular rate). Not arguing that they are just as useful for the non-linear case, though.

> The excessive jitter would not be passed to the GStreamer - the example already measures highest latency [...]. That can be used to calculate PTS and will give you fluent playback (after initial 1 or 2 stutters when detecting the latency). 

Isn't that exactly "passing on the excess jitter", if you dynamically increase the latency during playback? 😉 

Long story short: Let me repeat that I have no fundamental objections to demonstrating this use case by reading that operation mode in a future PR.

PR #252, opened 2025-11-20
Add link_requirement.yml workflow

peterbrightwell:
This workflow will link to the "parent" requirement if specified in the issue template.  Tested on https://github.com/peterbrightwell/dev but will need testing here before can be merged.


vt-tv (approved):
LGTM

PR #254, opened 2025-11-21
Adding documentation on how to change the audio channel count using t…

rochonma:
Adding documentation on how to change the audio channel count using the mxl-gst-videotestsrc tool. Also updating the video_flow.json example in the documentation to reflect latest changes.

There is one issue with having up to date .json files example in the tools.md though. It make the license parser of the github action to trip on these 2 lines:

"$copyright": "SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project.",
"$license": "SPDX-License-Identifier: Apache-2.0",

If I remove them, the github action work as intended. Do we want to modify the github actions to parse them correctly, or I just remove them from the markdown?

### DIFF
--- a/docs/Tools.md
+++ b/docs/Tools.md
@@ -101,7 +101,8 @@ A binary that uses the gstreamer 'videotestsrc' and 'audiotestsrc' elements to p
   ]
 }
 ```
-And here is an example of the NMOS FLow json file for an audio flow. To modify the number of audio channels present in a flow, modify the *"channel_count"* parameter to the desired value.
+Below is an example of an **augmented NMOS Flow JSON** for audio.  
+&emsp; **Note:** The *channel_count* property is a mandatory MXL requirement (not found in standard NMOS definition). To adjust the number of audio channels, simply update the *channel_count* value.
 
 ```json
 {


vt-tv (approved):
LGTM

docs/Tools.md
```diff
@@ -96,6 +101,31 @@ A binary that uses the gstreamer element 'videotestsrc' to produce video grains
   ]
 }
 ```
+And here is an example of the NMOS FLow json file for an audio flow. To modify the number of audio channels present in a flow, modify the *"channel_count"* parameter to the desired value.
```
garethsb:
```suggestion
And here is an example of the NMOS Flow json file for an audio flow. To modify the number of audio channels present in a flow, modify the *"channel_count"* parameter to the desired value.
```

(Do we keep json lower case?)

garethsb: Also noting that IS-04 does not define `channel_count` on an audio Flow... 🤔

lapointejp: In a documentation text I would tend to capitalize indeed.

lapointejp: Should we then just explicit that it's NMOS definition with this additional mandatory (for audio) property ?

sviengkhou: nitpick
I would change 
1 - "FLow" to "Flow"
2 - "..example of the NMOS FLow json...." to "...example of augmented NMOS Flow json file.."

KimonHoffmann: > Also noting that IS-04 does not define `channel_count` on an audio Flow... 🤔

Indeed, an omission that always puzzled me, especially given the level of detail the properties on video flows provide.

rochonma:
> @rochonma any updates on this PR?

Fixed the typo and applied suggested change around *channel_count* being an extra mandatory parameter not part of the NMOS flow definition.

vt-tv:
> > @rochonma any updates on this PR?
> 
> Fixed the typo and applied suggested change around _channel_count_ being an extra mandatory parameter not part of the NMOS flow definition.

can you resolve the associated conversations?

docs/Tools.md
rochonma: Change applied.

rochonma: Change applied.

PR #256, opened 2025-11-24
Fix macOS compilation

avonea:
Fixes several issues:
- add header for std::size_t
- add header for mkdtemp (and use a macOS appropriate temp directory for domain during tests)
- remove unsupported linker flag (`--exclude-libs`)
- address [libc++ limitation in using atomic_ref with cv-qualified types](https://github.com/llvm/llvm-project/issues/118378); using a `const_cast` should be safe in this situation, since the original references are non-const.

### DIFF
--- a/lib/internal/src/Sync.cpp
+++ b/lib/internal/src/Sync.cpp
@@ -69,7 +69,12 @@ namespace mxl::lib
     {
         static_assert(valid_specialization_v<T>, "Only 32 bit types with natural alignment are supported.");
 
+#if _LIBCPP_VERSION
+        // libc++ limitation due to: https://github.com/llvm/llvm-project/issues/118378
+        auto syncObject = std::atomic_ref{*const_cast<T*>(in_addr)};
+#else
         auto syncObject = std::atomic_ref{*in_addr};
+#endif
         while (syncObject.load(std::memory_order_acquire) == in_expected)
         {
             auto const now = currentTime(Clock::Realtime);


lib/internal/include/mxl-internal/MediaUtils.hpp
```diff
@@ -4,6 +4,9 @@
 #pragma once
 
 #include <cstdint>
+#ifdef __APPLE__
+#   include <cstddef> // for std::size_t
+#endif
```
KimonHoffmann: This should unconditional i.e. not predicated on `__APPLE__` being defined, as this is the header that defines `std::size_t` and the fact that it is available implicitly on other platforms/standard libraries/compilers is an error.

lib/internal/src/Sync.cpp
```diff
@@ -69,7 +69,12 @@ namespace mxl::lib
     {
         static_assert(valid_specialization_v<T>, "Only 32 bit types with natural alignment are supported.");
 
+#if _LIBCPP_VERSION
+        // libc++ limitation due to: https://github.com/llvm/llvm-project/issues/118378
+        auto syncObject = std::atomic_ref{const_cast<std::add_lvalue_reference_t<std::remove_cv_t<T>>>(*in_addr)};
```
KimonHoffmann: TBH I would find the following workaround a bit cleaner, especially since the call chain above seems incomplete to me[*]:

`... = std::atomic_ref{*const_cast<T*>(in_addr)}`

[*] Should be one of the following two to be complete and accurate:
1. `std::add_lvalue_reference_t<std::remove_cvref_t<T>>` (C++20)
2. `std::add_lvalue_reference_t<std::remove_cv_t<std::remove_reference_t<T>>>` (C++11)

KimonHoffmann (changes requested):
Thanks a lot for fixing the two issues preventing compilation on macOS!
I just have two small nits to pick...

lib/tests/Utils.hpp
```diff
@@ -11,6 +11,7 @@
 
 #ifdef __APPLE__
 #   include <stdexcept>
+#   include <unistd.h> // For mkdtemp
```
KimonHoffmann, line 14: Shouldn't that be
1. `cstdlib` and
2. unconditional, as this is the official header for the declaration of `mkdtemp`?

lib/CMakeLists.txt
```diff
@@ -59,10 +59,13 @@ set_target_properties(mxl
             VERSION                   "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}"
             SOVERSION                 "${PROJECT_VERSION_MAJOR}"
     )
+
+if (NOT APPLE)
 target_link_options(mxl
         PRIVATE
             "-Wl,--exclude-libs,ALL"
     )
+endif()
```
KimonHoffmann, line 68: Could you please indent the `target_link_options`invocation by one level?

KimonHoffmann (changes requested):
Missed something ... I should scroll down in reviews ;)

lib/internal/src/Sync.cpp
avonea: I think your solution is a lot better, I was trying to be overly clever.

lib/tests/Utils.hpp
avonea: `mkdtemp` is not an official part of `cstdlib`, and definitely not on libc++ AFAICT. https://en.cppreference.com/w/cpp/header/cstdlib.html

KimonHoffmann: `cstdlib` is the wrapper around `stdlib.h` and includes its definitions in addition to the C++ specifc standard definitions.

On gilbc systems `mkdtemp` is defined in `stdlib.h`, so `cstdlib` is the correct include here. But you are right about the BSDs, where `mkdtemp` is instead defined in the (much more logical) `unistd.h`.

So maybe make this an either/or conditional include?

avonea: Well, TIL. These distinctions are always a bit of a pain to nail down accurately, tbh.

KimonHoffmann:
If you don't mind I'll mixup the original commits with the review changes and then merge the PR.

lib/tests/Utils.hpp
KimonHoffmann: After closer inspection I reverted it to it's original form, because `cstdlib` is needed unconditionally in order to bring in the declaration of `std::getenv()`.

avonea: It definitely still compiled for me locally, but I may be behind the top of `main`. FWIW, this is code for tests, so there's perhaps no need to be frugal with includes :)

vt-tv (approved):
LGDM

PR #257, opened 2025-11-24
Cmake fix alma build

pkeroulas:
gst demo should optional when building the lib only.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,13 +50,6 @@ enable_testing()
 if(APPLE)
     # Add the official GStreamer framework paths to the PKG_CONFIG_PATH
     set(ENV{PKG_CONFIG_PATH} "/Library/Frameworks/GStreamer.framework/Versions/1.0/lib/pkgconfig:$ENV{PKG_CONFIG_PATH}")
-
-    # Configure threads
-    set(CMAKE_THREAD_LIBS_INIT "-lpthread")
-    set(CMAKE_HAVE_THREADS_LIBRARY 1)
-    set(CMAKE_USE_WIN32_THREADS_INIT 0)
-    set(CMAKE_USE_PTHREADS_INIT 1)
-    set(THREADS_PREFER_PTHREAD_FLAG ON)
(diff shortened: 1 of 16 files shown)


CMakeLists.txt
```diff
@@ -133,6 +132,8 @@ if(EXISTS "/etc/os-release")
     endif()
 endif()
 
+add_subdirectory(utils)
+
```
KimonHoffmann: No need to move that down, could you please revert this part of the change?

pkeroulas: The intention is to build the lib only for integration in external apps and minimize the footprint of the build env.
As currently commented in CMakeList, 
> Not adding gstreamer as a dependency since some rpm based platforms do not provide the required packages

This is the reason for adding `utils/gst-....` after the deb vs rpm detection.

KimonHoffmann: What I'm trying to get at is that the way the detection works is too broad and should not depend on the CPack variables as it does right now.

On RPM based distros this variable is unset and thus the utility will not be built on any of them.

KimonHoffmann: It makes more sense to mark the corresponding `find_package` calls as optional and then depend building the respective tools if their dependencies are available.

pkeroulas: Let's keep the 1st commit  only for this PR and make gstreamer optional in a seperate PR. Do you recommend to add a new option `BUILD_UTILS` ?
```
if(BUILD_UTILS)
    find_package(gstreamer)
    if(gstreamer_FOUND)
        add_subdirectory(utils)
    endif()
endif()
```

KimonHoffmann: Yes, I'll merge this PR and then take care of the optional build of the looping file source.
TBH it should live under tools anyway and thus be implicitly covered by the `BUILD_TOOLS` option.

KimonHoffmann:
You have to sign off the commit to signal your approval to the DCO, before I'm able to merge this PR.
You can do that by `git commit --amend --signoff` and then force pushing.

PR #258, opened 2025-11-25
Add a proc macro to generate mxl dlopen API from parsing bindings.rs.

mlefebvre1:


vt-tv:
Would it be possible to have a linked ticket or a description explaining why this is needed?  Thanks

pedro-alves-ferreira:
Hi @mlefebvre1!

Thanks for this but, if your goal is to have the bindings automatically generated and still load libmxl dynamically, that can do that directly with bindgen.
I had discussed this some time ago with @pac-work but I still haven't taken the time to open a PR for it. If that is the case, then I'll try to do it later today.

PR #259, opened 2025-11-25
feat(rust): avoid generating bindings manually

pedro-alves-ferreira:
In order to allow dynamic loading of libmxl, the bindings for the API functions were being generated manually.
`bindgen` now supports the option `dynamic_library_name`, which allows dynamic loading while still generating the bindings automatically.

This PR enables that option, removes the manually generated function bindings, and updates the `mxl` crate to use the automatically generated ones.

### DIFF
--- a/rust/mxl/src/error.rs
+++ b/rust/mxl/src/error.rs
@@ -27,9 +27,6 @@ pub enum Error {
     #[error("Other error: {0}")]
     Other(String),
 
-    #[error("dlopen: {0}")]
-    DlOpen(#[from] dlopen2::Error),
-
     #[error("Null string: {0}")]
     NulString(#[from] std::ffi::NulError),
 
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -143,29 +143,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "dlopen2"
-version = "0.8.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b54f373ccf864bf587a89e880fb7610f8d73f3045f13580948ccbcaff26febff"
-dependencies = [
- "dlopen2_derive",
- "libc",
- "once_cell",
- "winapi",
-]
-
-[[package]]
-name = "dlopen2_derive"
-version = "0.4.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "788160fb30de9cdd857af31c6a2675904b16ece8fc2737b2c7127ba368c9d0f4"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn",
-]
-
-[[package]]
 name = "either"
 version = "1.15.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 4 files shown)


mlefebvre1 (changes requested):
Looks good, but just a minor comment: I think we can now get rid of `dlopen2` dependency.

rust/mxl/src/flow/reader.rs
```diff
@@ -24,7 +24,7 @@ pub(crate) fn get_flow_info(
 ) -> Result<FlowInfo> {
     let mut flow_info: mxl_sys::mxlFlowInfo = unsafe { std::mem::zeroed() };
     unsafe {
-        Error::from_status(context.api.flow_reader_get_info(reader, &mut flow_info))?;
+        Error::from_status(context.api.mxlFlowReaderGetInfo(reader, &mut flow_info))?;
```
garethsb: Do you want the change from snake case to camel case?

Looks like a bindgen ParseCallbacks could be used to auto generate snake case if not?

pedro-alves-ferreira: Good idea. I didn't bother too much because it's only on the sys crate, so not visible to the regular end user, and I didn't want to manually annotate fields, because the whole purpose of the PR is to automate the process. If it can be automated, I'm all for it. I'll give it a try later today. Thanks!

rust/mxl/src/error.rs
```diff
@@ -32,6 +32,9 @@ pub enum Error {
 
     #[error("Null string: {0}")]
     NulString(#[from] std::ffi::NulError),
+
```
pac-work, line 32: If we are also removing the `dlopen2` as suggested by somebody else, we should probably remove the error as well (above). But I guess compiler will shout on that once that activity happens. And I second the `dlopen2` removal, if we are switching to `libloading`.

pedro-alves-ferreira: Done, thanks.

rust/mxl/src/flow/reader.rs
pedro-alves-ferreira: Done.

pedro-alves-ferreira:
Updated as discussed. There is a breaking change noted in the second commit.

rust/mxl/src/flow/reader.rs
garethsb: Pretty cool right? Even if it actually makes for a bigger diff?

garethsb (approved):

vt-tv:
@pedro-alves-ferreira please rebase and I'll merge

vt-tv:
We seem to have unreliable builds now that the rust code has merged. look at the automated tests : the 'Build on Ubuntu - xxx' jobs are not even starting or if starting, not completing.

pedro-alves-ferreira:
> > We seem to have unreliable builds now that the rust code has merged. look at the automated tests : the 'Build on Ubuntu - xxx' jobs are not even starting or if starting, not completing.
> 
> Look at the actions tab : we it only ran the rust tasks. not the c++ jobs! https://github.com/dmf-mxl/mxl/actions

I don't think this has anything to do with this or any other PR related to Rust.
[This PR](https://github.com/dmf-mxl/mxl/pull/235) changed the way things are built. We used to have "Build on Ubuntu (etc)", and now we have "Build (C library|Rust bindings) on Ubuntu (etc)". Maybe GitHub got confused and is still looking for the old checks?

PR #268, opened 2025-12-04
Fix segfault in mxl-info

jonasohland:
The ostream operator was always calling itself, and the compiler was warning about that. Turns out mxl-info was in fact not working properly. This commit fixes that.


jonasohland:
I think Kimon is already fixing that in #247

PR #273, opened 2025-12-05
updating build instructions

scenarnick:
Modifying building.md to be more clear about the "mount" instructions, make text better readable by adding paragraphs and removing mentions of NVIDIA toolkit requirement


KimonHoffmann:
No, absolutely, the mounts should stay in place, just the orphaned environment variables in the `runArgs`.

PR #279, opened 2025-12-06
Add the Addressability.md file

vt-tv:
Add the Addressability.md file. 
Unrelated : add an ubuntu 26.04 devcontainer file.

### DIFF
--- a/docs/Addressability.md
+++ b/docs/Addressability.md
@@ -45,7 +45,7 @@ mxl:///domain/path?id=<flowId1>&id=<flowId2>
 <domain-path> = "/" <segment> *( "/" <segment> )
 ```
 
-The ```domain-path``` part of the URI is required and refers to that on disk *in the context of the media function* of the MXL domain.  The domain path may need to be translated by media functions or orchestration layer(s). For example, an MXL domain may live on ```/dev/shm/domain1``` on a host but is mapped inside a container to ```/dev/shm/mxl```.  This translation is outside the scope of the MXL project.  
+The ```domain-path``` part of the URI is required and refers to the file system path *in the context of the media function* of the MXL domain.  The domain path may need to be translated by media functions or orchestration layer(s). For example, an MXL domain may live on ```/dev/shm/domain1``` on a host but is mapped inside a container to ```/dev/shm/mxl```.  This translation is outside the scope of the MXL project.  
 
 
 


docs/Addressability.md
```diff
@@ -0,0 +1,96 @@
...
+# MXL Addressability
+
+The following addressing scheme allows applications to refer to one or multiple flows hosted in an MXL domain on a local host or on a remote server.  
+
+## URI 
+
+An MXL flow or group of flows can be addressed using this RFC 3986 compatible format:  
+
+```
+mxl://<authority>/<domain-path>?id=<flow1>&<flow2>
```
KimonHoffmann: Should repeat `id=` for the second flow.

jonasohland: I was not sure if repeating a query field names is allowed, but at least the golang stdlib parser is fine with it: https://go.dev/play/p/D4CyceGmt9B

docs/Addressability.md
```diff
@@ -0,0 +1,96 @@
...
+mxl:///domain/path?id=<flowId1>&id=<flowId2>
+```
+
+### Domain Path (Required)
+
+```
+<domain-path> = "/" <segment> *( "/" <segment> )
+```
+
+The ```domain-path``` part of the URI is required and refers to that on disk *in the context of the media function* of the MXL domain.  The domain path may need to be translated by media functions or orchestration layer(s). For example, an MXL domain may live on ```/dev/shm/domain1``` on a host but is mapped inside a container to ```/dev/shm/mxl```.  This translation is outside the scope of the MXL project.
```
jonasohland:
```suggestion
The ```domain-path``` part of the URI is required and refers to the file system path *in the context of the media function* of the MXL domain.  The domain path may need to be translated by media functions or orchestration layer(s). For example, an MXL domain may live on ```/dev/shm/domain1``` on a host but is mapped inside a container to ```/dev/shm/mxl```.  This translation is outside the scope of the MXL project.  
```

docs/Addressability.md
```diff
@@ -0,0 +1,96 @@
...
+
+If authority is absent, the syntax becomes:
+```
+mxl:///domain/path?id=<flowId1>&id=<flowId2>
+```
+
+### Domain Path (Required)
+
+```
+<domain-path> = "/" <segment> *( "/" <segment> )
```
jonasohland, line 45: I think we should state that the domain path is always absolute.

vt-tv: I'm not so sure about that.  Maybe an orchestration layer will want to use abstracted names such as : mxl://host1.local/MORNING_SHOW?id=<id> and implement translation from abstract names to absolute paths valid in the context of the media function running inside a container.

jonasohland: Further down we refer to the domain-path element as something "***on-disk*** *in the context of the media function*", so maybe we should remove that part and just say its a string that could refer to some filesystem node, but could also be just a name that must be translated to an actual filesystem node.

jonasohland: But ok, I get where you are coming from. I am fine with leaving it as-is.

jonasohland (approved):

PR #285, opened 2025-12-11
Users/vt tv/253 mxl gstvideo sink does not handle MXL_ERR_FLOW_INVALID

vt-tv:
1. Handle invalid source flow. 
2. Reconnect and resync when the flows come back.  
3. Properly set the grouphint tag when testsrc generates multiple flows.

### DIFF
--- a/docs/Tools.md
+++ b/docs/Tools.md
@@ -127,6 +127,8 @@ OPTIONS:
                               https://gstreamer.freedesktop.org/documentation/videotestsrc/index.html?gi-language=c#GstVideoTestSrcPattern
   -t,     --overlay-text TEXT [EBU DMF MXL]
                               Change the text overlay of the test source
+  -g,     --group-hint TEXT [mxl-gst-testsrc-group]  
+                              The group-hint value to use in the flow json definition                               
 ```
 
 Example to run with video only:


tools/mxl-gst/sink.cpp
```diff
@@ -551,33 +617,106 @@ namespace
...
+                        MXL_INFO("Reconnected to sound flow: {}.", flowId);
+                        // Get the flow config info again
+                        if (auto const ret = mxlFlowReaderGetConfigInfo(_reader, &_configInfo); ret != MXL_STATUS_OK)
+                        {
+                            // Something is very wrong. we cannot recover from this.
+                            MXL_ERROR("Failed to get flow config info with status code {}. Exiting.", static_cast<int>(ret));
+                            return;
+                        }
+
+                        // Realign to current index.
```
mlefebvre1, line 602: These 4 lines below are also the same as the init. I'm sure we can move that logic to a function. If we ever have to change that logic, we only need to change it in 1 place.

Same comment applies to video.

tools/mxl-gst/sink.cpp
```diff
@@ -551,33 +617,106 @@ namespace
...
                 }
                 else
                 {
-                    MXL_ERROR("Unexpected error when reading samples at index {} with status {}. Exiting.", requestedIndex, static_cast<int>(ret));
-                    return;
+                    // On any other error, verify if the _reader is nullptr (which indicates that the flow may have been invalidated).
+                    // In that case, try to recreate it. Otherwise, exit.
+                    if (_reader == nullptr)
+                    {
+                        // Create a new reader
```
mlefebvre1, line 638: We pretty much have 4 copy paste of the same logic to recover from an invalid flow. I would move that logic to a function.

mlefebvre1 (changes requested):

tools/mxl-gst/sink.cpp
```diff
@@ -551,33 +595,84 @@ namespace
...
+                        auto const flowId = uuids::to_string(_configInfo.common.id);
```
mlefebvre1, line 639: It's not clear to me why we can't just use `handleInvalidFlow` here. Apart from the sleep they are essentially the same. I don't think including the sleep in that function would cause any issues, though 500ms might be excessive. Same comment for video.

mlefebvre1 (changes requested):

tools/mxl-gst/sink.cpp
vt-tv: Fixed

tools/mxl-gst/sink.cpp
vt-tv: Fixed

tools/mxl-gst/sink.cpp
lapointejp: I concur here. I had the same thought as Michael.

tools/mxl-gst/testsrc.cpp
```diff
@@ -834,6 +863,10 @@ int main(int argc, char** argv)
     auto textOverlayOpt = app.add_option("-t,--overlay-text", textOverlay, "Change the text overlay of the test source");
     textOverlayOpt->default_val("EBU DMF MXL");
 
+    auto groupHint = std::string{};
```
lapointejp, line 866: Should update `docs/Tools.md`.

vt-tv: Done

Issue #265, opened 2025-12-03
[Task] Add C-Level API to create, maintain and wait on synchronization groups


KimonHoffmann:
### Description

The API must be designed in a way that enables a source compatible extension to wait on multiple synchronization groups at once ("any of").

PR #286, opened 2025-12-12
Add basic multi-flow synchronization

### LINKED ISSUE #264: [Task] Add a "synchronization group" facility to the internal library
### Description

Add a "synchronization group" facility to the internal library that allows waiting on (full or partial) grains accross multiple flows.

This facility should employ strategies that attempt to minimize the number of syscall+wake pairs executed as part of the wait.

KimonHoffmann:
This PR adds the ability to synchronize on data availability across a group of flows with single API call.

At this point in time this PR is intended to gather feedback on the API as well as the implementation. It has not yet been tested and still misses automated tests at this moment.

Closes #264 and #265.

### DIFF
--- a/lib/internal/include/mxl-internal/Instance.hpp
+++ b/lib/internal/include/mxl-internal/Instance.hpp
@@ -194,11 +194,14 @@ namespace mxl::lib
     /// Utility function to convert from a C mxlInstance handle to a C++ Instance class
     Instance* to_Instance(mxlInstance instance) noexcept;
 
-    /// Utility function to convert from a C mxlFlowReader handle to a C++ FlowReaders instance.
+    /// Utility function to convert from a C mxlFlowReader handle to a C++ FlowReader instance.
     FlowReader* to_FlowReader(mxlFlowReader reader) noexcept;
 
     /// Utility function to convert from a C mxlFlowWriter handle to a C++ FlowWriter instance.
     FlowWriter* to_FlowWriter(mxlFlowWriter writer) noexcept;
+
+    /// Utility function to convert from a C mxlFlowSynchronizationGroup handle to a C++ FlowSynchronizationGroup instance.
+    FlowSynchronizationGroup* to_FlowSynchronizationGroup(mxlFlowSynchronizationGroup group) noexcept;
 
     /**************************************************************************/
     /* Inline implementation.                                                 */
@@ -254,6 +257,11 @@ namespace mxl::lib
     inline FlowWriter* to_FlowWriter(mxlFlowWriter writer) noexcept
     {
         return reinterpret_cast<FlowWriter*>(writer);
+    }
+
+    inline FlowSynchronizationGroup* to_FlowSynchronizationGroup(mxlFlowSynchronizationGroup group) noexcept
+    {
+        return reinterpret_cast<FlowSynchronizationGroup*>(group);
     }
 
 } // namespace mxl::lib
--- a/lib/include/mxl/flow.h
+++ b/lib/include/mxl/flow.h
@@ -150,6 +150,8 @@ extern "C"
     typedef struct mxlFlowReader_t* mxlFlowReader;
     typedef struct mxlFlowWriter_t* mxlFlowWriter;
 
+    typedef struct mxlFlowSynchronizationGroup_t* mxlFlowSynchronizationGroup;
+
     /**
      * Attempt to create a flow writer for a given flow definition. If the flow does not exist already, it is created and 'created' will be set to
      * true. If the flow exists, it will be opened and 'created' will be set to false. If the flow exists already it is not guaranteed that the flow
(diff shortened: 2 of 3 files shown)


lib/src/time.cpp
```diff
@@ -29,27 +30,19 @@ extern "C"
...
+    return (editRate != nullptr) ? mxl::lib::timestampToIndex(*editRate, mxl::lib::Timepoint(timestamp)) : MXL_UNDEFINED_INDEX;
 }
 
 extern "C"
 MXL_EXPORT
 uint64_t mxlIndexToTimestamp(mxlRational const* editRate, uint64_t index)
 {
-    // Validate the edit rate
-    if ((editRate == nullptr) || (editRate->denominator == 0) || (editRate->numerator == 0))
+    if (editRate != nullptr)
```
pedro-alves-ferreira, line 40: How about reversing the condition and doing an early return?

```
if (editRate == nullptr)
{
    return MXL_UNDEFINED_INDEX;
}
```

lib/internal/include/mxl-internal/FlowSynchronizationGroup.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+        public:
+            explicit ListEntry(DiscreteFlowReader const& reader, std::uint16_t minValidSlices);
+            explicit ListEntry(ContinuousFlowReader const& reader);
+
+        private:
+            explicit ListEntry(FlowReader const& reader, Variant variant);
+        };
+
+        using ReaderList = std::forward_list<ListEntry>;
```
pedro-alves-ferreira, line 76: I wonder if it is more efficient to use a `std::vector`. These will never be huge and we'd get much better locality.

lib/internal/include/mxl-internal/Instance.hpp
```diff
@@ -168,6 +187,9 @@ namespace mxl::lib
         /// Protects the maps
         std::mutex _mutex;
 
+        /// The set of active flow synchronization groups
+        std::forward_list<FlowSynchronizationGroup> _syncGroups;
```
pedro-alves-ferreira, line 181: Same comment here about the use of a list vs `std::vector`.

lib/internal/src/FlowSynchronizationGroup.cpp
```diff
@@ -0,0 +1,129 @@
...
+                    case Variant::Continuous:
+                        result = static_cast<ContinuousFlowReader const*>(current->reader)->waitForSamples(expectedIndex, deadline);
+                        break;
+                }
+
+                if (result == MXL_STATUS_OK)
+                {
+                    // If the current source delay of this flow exceeds any previosuly observed source delay of
+                    // this flow we update the cached maximum and if this new maximum turns out to be bigger than
+                    // the maximum source delay observed for the flow at the head of the lis, we move this flow
```
pedro-alves-ferreira: (...) at the head of the list (...)

lib/internal/src/Instance.cpp
```diff
@@ -433,4 +438,22 @@ namespace mxl::lib
...
     }
 
+    FlowSynchronizationGroup* Instance::createFlowSynchronizationGroup()
+    {
+        return &_syncGroups.emplace_front();
+    }
+
+    void Instance::releaseFlowSynchronizationGroup(FlowSynchronizationGroup const* group)
+    {
+        auto prev = _syncGroups.before_begin();
```
pedro-alves-ferreira, line 491: Consider replacing the loop with an algorithm, e.g.:

```
const auto it = std::find_if(_syncGroups.begin(), _syncGroups.end(), [](const auto& current) &current == group);
if(it != _syncGroups.end())
{
    _syncGroups.erase(it);
}
```

lib/internal/src/FlowSynchronizationGroup.cpp
```diff
@@ -0,0 +1,129 @@
...
+            prev = current;
+        }
+
+        _readers.emplace_after(prev, reader, minValidSlices);
+    }
+
+    void FlowSynchronizationGroup::addReader(ContinuousFlowReader const& reader)
+    {
+        auto prev = _readers.before_begin();
+        for (auto current = std::next(prev); current != _readers.end(); ++current)
```
pedro-alves-ferreira, line 51: Consider replacing the loop with an algorithm.

lib/internal/src/FlowSynchronizationGroup.cpp
```diff
@@ -0,0 +1,129 @@
...
+            prev = current;
+        }
+
+        _readers.emplace_after(prev, reader);
+    }
+
+    void FlowSynchronizationGroup::removeReader(FlowReader const& reader)
+    {
+        auto prev = _readers.before_begin();
+        for (auto current = std::next(prev); current != _readers.end(); ++current)
```
pedro-alves-ferreira, line 66: Consider replacing the loop with an algorithm.

lib/internal/src/FlowSynchronizationGroup.cpp
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    FlowSynchronizationGroup::ListEntry::ListEntry(ContinuousFlowReader const& reader)
+        : ListEntry{reader, Variant::Continuous}
+    {}
+
+    void FlowSynchronizationGroup::addReader(DiscreteFlowReader const& reader, std::uint16_t minValidSlices)
+    {
+        auto prev = _readers.before_begin();
+        for (auto current = std::next(prev); current != _readers.end(); ++current)
```
pedro-alves-ferreira, line 35: Consider replacing the loop with an algorithm.

lib/src/time.cpp
KimonHoffmann: I try to follow the scheme where the more likely/expected path corresponds to the `true` branch of the conditional, except for cases where it makes the code cumbersome to write and follow.

lib/internal/src/Instance.cpp
KimonHoffmann: Does not work because of the peculiarities of `forward_list<>` that only offers `erase_after()`.

lib/internal/src/FlowSynchronizationGroup.cpp
KimonHoffmann: Does not work, because of the peculiarities of `forward_list<>`. We need to maintain `prev` to enable `emplace_after()` later on.

lib/internal/src/FlowSynchronizationGroup.cpp
KimonHoffmann: Does not work, because of the peculiarities of `forward_list<>`. We need to maintain `prev` to enable `emplace_after()` later on.

lib/internal/src/FlowSynchronizationGroup.cpp
KimonHoffmann: Does not work because of the peculiarities of `forward_list<>` that only offers `erase_after()`.

lib/internal/include/mxl-internal/FlowSynchronizationGroup.hpp
KimonHoffmann: You are of course correct WRT to locality and the potentially improved performance, but I doubt it makes a serious difference here (especially when we may ultimately do one or more syscalls along the way).

I opted for a linked list here to make the "move to front" mechanism more well defined and easy to follow.

I don't have any strong opinion either way here, though.

lib/internal/src/Instance.cpp
pedro-alves-ferreira: I should have said that I wrote this assuming the use of a vector.  I just find it cleaner and easier to maintain if we use algorithms as often as possible. Otherwise, it's fine.

lib/internal/include/mxl-internal/FlowSynchronizationGroup.hpp
```diff
@@ -0,0 +1,81 @@
...
+#include "Timing.hpp"
+
+namespace mxl::lib
+{
+    class ContinuousFlowReader;
+    class DiscreteFlowReader;
+    class FlowReader;
+
+    /**
+     * A set of weak references to a flow readers that that can be used to check
```
mlefebvre1: too many "that"

lib/internal/src/FlowSynchronizationGroup.cpp
```diff
@@ -0,0 +1,129 @@
...
+        , maxObservedSourceDelay{}
+    {
+        auto const configInfo = reader.getFlowConfigInfo();
+        grainRate = configInfo.common.grainRate;
+    }
+
+    FlowSynchronizationGroup::ListEntry::ListEntry(DiscreteFlowReader const& reader, std::uint16_t minValidSlices)
+        : ListEntry{reader, Variant::Discrete}
+    {
+        minValidSlices = minValidSlices;
```
mlefebvre1: I don't think it correctly sets the member attribute, because the function argument shadows it. Should use `this->minValidSlices = minValidSlices;` or rename the argument.

KimonHoffmann: You're right, good catch. I renamed the member after the fact and didn't catch the naming clash.

pedro-alves-ferreira: Isn't there a warning we can enable to detect shadowing?

KimonHoffmann: Definitely, and I'm a bit surprised that we haven#t enabled all of these yet.
I'll submit a PR that enables a sensible set of warnings once I'm back.

KimonHoffmann: PR #310 enables warnings that should detect such issues in the future.

pedro-alves-ferreira (approved):
LGTM

PR #288, opened 2025-12-15
Enable static builds with pkg-config integration.

jptrainor:
This PR fixes static build correctness across CMake and `pkg-config`.

Static build issues surfaced while implementing static linking using `pkg-config` for an ffmpeg/MXL integration. MXL supported shared-library builds only, ignored `BUILD_SHARED_LIBS`, and provided incomplete `pkg-config` metadata for static linkage.

Key build-related changes:
- Correct handling of `BUILD_SHARED_LIBS=ON|OFF` across mxl targets
- Added `MXL_ENABLE_PIC` (default ON) to support static builds used in shared contexts
- Fixed pkg-config metadata for static linking
- Corrected CMake export-set dependency relationships

This change adds a C consumer link test to prevent regressions.

Tested with:
- Linux (GCC, Clang)
- macOS (Clang/x86_64)
- Shared and static builds (`BUILD_SHARED_LIBS=ON|OFF`)
- Debug and release builds
- Static ffmpeg build as downstream consumer on Linux

The ffmpeg integration lives in a downstream fork and is still being
finalized, but is included here for reference:
https://github.com/cbcrc/FFmpeg/tree/dmf-mxl/master

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,6 +84,20 @@ if(APPLE)
     set(ENV{PKG_CONFIG_PATH} "/Library/Frameworks/GStreamer.framework/Versions/1.0/lib/pkgconfig:$ENV{PKG_CONFIG_PATH}")
 endif()
 
+# Default build is shared libraries.
+# Use '-DBUILD_SHARED_LIBS=OFF' for static build.
+option(BUILD_SHARED_LIBS "Build using shared libraries" ON)
+
+# Default PIC ON for static builds.
+# Use -DMXL_ENABLE_PIC=OFF to disable PIC
+option(MXL_ENABLE_PIC "Enable PIC for static libraries" ON)
+
+# Policy: shared always PIC, static set to MXL_ENABLE_PIC.
+set(MXL_POSITION_INDEPENDENT_CODE ON)
+if(NOT BUILD_SHARED_LIBS)
+  set(MXL_POSITION_INDEPENDENT_CODE ${MXL_ENABLE_PIC})
+endif()
+
 add_subdirectory(lib)
 add_subdirectory(utils)
 if (BUILD_TOOLS)
--- a/docs/Building.md
+++ b/docs/Building.md
@@ -55,6 +55,24 @@ cmake .. --preset Linux-Clang-Debug
 cmake --build build/Linux-Clang-Debug --target all
 ```
 
+## Static build notes
+
+By default, the CMake presets build MXL as shared libraries. To build
+MXL as static libraries instead, use the CMake option
+`-DBUILD_SHARED_LIBS=OFF`.
+
+For example:
+
+```
+mkdir build
+cd build
+cmake .. --preset Linux-GCC-Debug -DBUILD_SHARED_LIBS=OFF
+cmake --build build/Linux-GCC-Debug --target all
+```
+
+PIC is enabled by default and can be disabled with
+`-DMXL_ENABLE_PIC=OFF`.
+
 ## macOS notes
 
 1. Install the [Homebrew](https://brew.sh) package manager
--- a/lib/tests/test_pkgconfig_c_link.sh
+++ b/lib/tests/test_pkgconfig_c_link.sh
@@ -1,0 +1,81 @@
+#!/bin/sh
+
+# SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project.
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+usage() {
+    echo "usage: $0 <build_dir> <triplet> <BUILD_SHARED_LIBS: ON|OFF> <MXL_LINKER_FLAG=<linker-selection-flag>> <c_compiler>" >&2
+    echo "  build_dir         Path to the CMake build directory." >&2
+    echo "  triplet           vcpkg target triplet used for the build." >&2
+    echo "  BUILD_SHARED_LIBS Indicates whether the build uses shared libraries (ON or OFF)." >&2
+    echo "  MXL_LINKER_FLAG   Optional compiler linker-selection flag (e.g. -fuse-ld=lld); empty means use default linker." >&2
+    echo "  c_compiler        Path to the C compiler executable." >&2
+}
+
+if [ "$#" -ne 5 ]; then
+    usage
+    exit 1
+fi
+
+BUILD_DIR="$1"
+VCPKG_TRIPLET="$2"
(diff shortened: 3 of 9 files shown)


vt-tv:
You are mising the SPDX headers/comments in these files : 

lib/tests/test_pkgconfig_c_link.cpp
lib/tests/test_pkgconfig_c_link.sh

CMakeLists.txt
```diff
@@ -52,6 +52,14 @@ if(APPLE)
     set(ENV{PKG_CONFIG_PATH} "/Library/Frameworks/GStreamer.framework/Versions/1.0/lib/pkgconfig:$ENV{PKG_CONFIG_PATH}")
 endif()
 
+# Build with PIC (only affects static builds).
+option(MXL_ENABLE_PIC "Enable position-independent code" ON)
+set(CMAKE_POSITION_INDEPENDENT_CODE ${MXL_ENABLE_PIC})
```
jonasohland: I would prefer setting this as a target property only on the exported libraries instead of setting it globally for all targets.

@KimonHoffmann what do you think?

docs/Building.md
```diff
@@ -55,6 +55,24 @@ cmake .. --preset Linux-Clang-Debug
 cmake --build build/Linux-Clang-Debug --target all
 ```
 
+## Static build notes
+
+By default, the CMake presets build MXL as shared libraries. To build
+MXL as static libraries instead, use the CMake option
+`-DBUILD_SHARED_LIBS=OFF`.
```
jonasohland:
```suggestion
`-DMXL_BUILD_SHARED_LIBS=OFF`.
```

CMakeLists.txt
KimonHoffmann: This variable is slightly different to other global Cmake variables in that it only applies as a default for library targets that aren't shared libraries.

What I find a bit unfortunate though is that this also affects the behavior of `try_compile` so I would prefer if we'd:

Define a `MXL_POSITION_INDEPENDENT_CODE` variable whose value depends on `BUILD_SHARED_LIBS` that we then use as the explicit value for a property on each target.

lib/tests/test_pkgconfig_c_link.sh
```diff
@@ -0,0 +1,78 @@
...
+    RPATH_FLAGS="-Wl,-rpath,$BUILD_DIR/lib"
+else
+    usage
+    exit 2
+fi
+
+EXTRA_C_OPTS=
+if [ "$BUILD_SHARED_LIBS" = "OFF" ] && [ "$C_COMPILER_ID" = "Clang" ]; then
+    # turn on LTO for clang static build
+    EXTRA_C_OPTS="-flto -fuse-ld=lld"
```
KimonHoffmann: There is a canonical way to enable IPO/LTO in Cmake projects for which I will provide a separate PR.

For this test specifically I don't see any real value in explicitly enabling IPO for the test fragment build and would thus remove this section from the script.

KimonHoffmann (commented):
Thank you very much for this contribution and filling in the missing bits for the static library build. That has been long on my TODO list and I very much appreciate being able to tick this off now.

Regarding the pkg-config compile check I'm a bit conflicted. I think it's a good idea, but having it as part of the tests on the side feels wrong to me. Could you please transform this into a post-build step that is independent of the other tests?

Also regarding the C++ parts of this PR please try to match the style present in other tests with regards to pointer types, right hand types and explicit qualification of entities in the global namespace.

CMakeLists.txt
jptrainor: > This variable is slightly different to other global Cmake variables in that it only applies as a default for library targets that aren't shared libraries.
> 
> What I find a bit unfortunate though is that this also affects the behavior of `try_compile` so I would prefer if we'd:
> 
> Define a `MXL_POSITION_INDEPENDENT_CODE` variable whose value depends on `BUILD_SHARED_LIBS` that we then use as the explicit value for a property on each target.

Resolved: MXL_POSITION_INDEPENDENT_CODE is always ON for shared builds, is set to MXL_ENABLE_PIC for static builds (default ON), and is then applied per target.

jptrainor: > I would prefer setting this as a target property only on the exported libraries instead of setting it globally for all targets.
> 
> @KimonHoffmann what do you think?

Resolve: it is now set as a property on each target, including the lib/fabrics/ofi targets.

lib/tests/test_pkgconfig_c_link.sh
jptrainor: The -flto option is not necessary and I removed it. However the LTO capable linker option (-fuse-ld=lld) is necessary to link Clang release builds in the test. I changed the test to pass the value of CMAKE_INTERPROCEDURAL_OPTIMIZATION_RELEASE to the script rather than inferring it from BUILD_SHARED_LIBS. That is more accurate, and should make it easier to adapt to future canonicalization of IPO/LTO.

KimonHoffmann: Tying this to IPO and not predicating `-fuse-ld=lld` on the compiler ID still feels wrong to me.

I *think* (i.e. haven't tried it yet) the correct way to tackle the linker issue you're observing is by passing `CMAKE_C_COMPILER_LINKER` to the script and use that for linking the test binary.

jptrainor: I'm giving this a try. Standby for feedback.

KimonHoffmann: Awesome please be aware that I introduced #292 since you created this PR and am about to merge it.
So rebasing on `main` after the fact and enabling IPO should prove itself as a good test to see whether this approach works as expected.

jptrainor: I found that CMAKE_C_COMPILER_LINKER was set to /usr/bin/ld, and that variable is not configurable. Also separating the compile and link command in the test script was potentially brittle because of the necessity for the link command to reproduce all the dependencies that are otherwise automatically staged by the compiler's link step (e.g. crt1.o).  Further, Cmake intends that to be derived not set by the user so changing it isn't an option.

I subsequently determined that cmake intends the user to set CMAKE_LINKER_TYPE to describe the linker. I added that to the clang preset and set it to "LLD".  With that set, then CMAKE_C_USING_LINKER_LLD is automatically set to "-fuse-ld=lld" and that can be passed to the test_pkgconfig_c_link.sh script and added to the script's compile command. That removes the awkard logic that was formerly in the test script to deduce the same compile option.

PR #290, opened 2025-12-17
Flow creation/destruction API simplification

### LINKED ISSUE #144: Force correct flow writer behaviour with an API change
In the TSC call right before IBC, it was mentioned that there is no plan to support, and no use-case for multi-writer flows. It was also mentioned that some of the reference applications are not well-behaved when it comes to creating their flows, and if these flows should be cleaned up when they exit. 

Based on this, I personally think that with a little API makeover we could make wrong behavior of this kind impossible and with the fix in #134 make it easier for implementers to build applications that behave correctly in all cases.

My idea is to get rid of the function that creates the flow without opening it, and make it so that the only Way to obtain a Writer is to create a flow.

```
(1) mxlFabricsCreateFlowWriter(string flowDescription) -> FlowWriter

(2) mxlFabricsCreateFlowReader(string flowId) -> FlowReader
```

The function to create the flow writer (1) should behave like this:
Succeeds if:
 - The flow does not exist. A new flow is created.
 - The flow exists, but has no writer (is stale). The old flow is cleaned up and the new flow is created.

Fails if:
 - The flow exists, and has an active writer.

When a flow writer is destroyed, its flow is also destroyed.

The function to create the flow reader (2) should behave like this:
Succeeds if:
 - The flow exists.

Fails if:
 - The flow does not exist.

Additionally: reading from a flow that has no writer should fail with an error that signals to the reader that it should try to re-open the flow.

I know, it would be kind of a pain to change this now, but I at least would like to have brought this up.

jonasohland:
This is my proposal for #144.

Flows are created inside the "mxlCreateFlowWriter()" and released in the "mxlReleaseFlowWriter()" functions now. 
This is just a draft for now. Before merging this I would like to improve some of the tests and fix the comments and docs. I will also fix the rust API.

The diff is pretty hard to read, I suggest checking out the branch locally to review.

Fixes #144

### DIFF
--- a/lib/internal/src/Deferred.hpp
+++ b/lib/internal/src/Deferred.hpp
@@ -20,7 +20,7 @@ namespace mxl::lib
         Deferred& operator=(Deferred<F> const&) = delete;
         Deferred& operator=(Deferred<F>&&) = delete;
 
-        /** \brief Calles the deferred function
+        /** \brief Calls the deferred function
          */
         constexpr ~Deferred() noexcept(std::is_nothrow_invocable_v<F>)
         {
--- a/lib/internal/src/Instance.cpp
+++ b/lib/internal/src/Instance.cpp
@@ -203,7 +203,7 @@ namespace mxl::lib
                     {
                         removeFlowWatch = (dynamic_cast<ContinuousFlowWriter*>((*pos).second.get()) == nullptr);
 
-                        // Delete if flow if we are the last writer.
+                        // Delete the flow if we are the last writer.
                         if (writer->isExclusive() || writer->makeExclusive())
                         {
                             _flowManager.deleteFlow(id);
--- a/lib/internal/tests/test_flowmanager.cpp
+++ b/lib/internal/tests/test_flowmanager.cpp
@@ -577,7 +577,7 @@ TEST_CASE_PERSISTENT_FIXTURE(mxl::tests::mxlDomainFixture, "FlowManager: paralle
             }
 
             {
-                // Create or open the flow. This is where the race happenes. Only one of the workers should be able to create the flow (created ==
+                // Create or open the flow. This is where the race happens. Only one of the workers should be able to create the flow (created ==
                 // true). The others should still receive an opened flow that is writable, but created should be false.
                 auto [created,
                     flow] = worker->mgr->createOrOpenDiscreteFlow(id, def, MXL_DATA_FORMAT_VIDEO, 1, mxlRational{50, 1}, payloadSize, 1, sliceSizes);


jonasohland:
> Can you document (here, not in the code for now) what will be the api to 'open Or Create' a flow once we support dual writers? how can we do it without a race ?

I _think_ it should be possible to support that without a change to the API:
In `mxlCreateFlowWriter` we would have to handle EEXIST and then open the existing flow instead of failing. When creating a FlowWriter, we must `flock` with `LOCK_SH` (shared) the flow data file (I think that is already what is happening). When destroying the flow writer, this lock should be released and then the process should try to acquire a `flock` with `LOCK_EX` (exclusive). If then a exclusive lock of the flow data file is held, the flow can be deleted.

lib/src/flow.cpp
```diff
@@ -236,49 +193,99 @@ mxlStatus mxlReleaseFlowReader(mxlInstance instance, mxlFlowReader reader)
...
-mxlStatus mxlCreateFlowWriter(mxlInstance instance, char const* flowId, char const* /*options */, mxlFlowWriter* writer)
+mxlStatus mxlCreateFlowWriter(mxlInstance instance, char const* flowDef, char const* options, mxlFlowWriter* writer, mxlFlowConfigInfo* configInfo)
 {
+    if ((instance == nullptr) || (flowDef == nullptr) || (writer == nullptr))
+    {
+        return MXL_ERR_INVALID_ARG;
+    }
+
+    auto flowData = std::unique_ptr<FlowData>{};
+    auto cppInstance = to_Instance(instance);
```
KimonHoffmann, line 173: The previous code intentionally checked the result of `to_Instance()` for the null pointer instead of checking the input and then assuming `to_Instance()` to always succeed, because we might add additional checking to `to_Instance()` in the future that can result in a null pointer return, even if the input was non-null.

lib/src/flow.cpp
```diff
@@ -236,49 +193,99 @@ mxlStatus mxlReleaseFlowReader(mxlInstance instance, mxlFlowReader reader)
...
+
     try
     {
-        if (writer != nullptr)
+        flowData = cppInstance->createFlow(flowDef, (options == nullptr) ? std::string{} : options);
+        *configInfo = flowData->flowInfo()->config;
+    }
+    catch (std::filesystem::filesystem_error const& e)
```
KimonHoffmann, line 199: Catching `std::system_error` here instead covers a few more error cases.

lib/src/flow.cpp
```diff
@@ -236,49 +193,99 @@ mxlStatus mxlReleaseFlowReader(mxlInstance instance, mxlFlowReader reader)
...
 mxlStatus mxlReleaseFlowWriter(mxlInstance instance, mxlFlowWriter writer)
 {
+    if ((instance == nullptr) || (writer == nullptr))
+    {
+        return MXL_ERR_INVALID_ARG;
+    }
+
+    auto err = MXL_STATUS_OK;
+    auto const cppInstance = to_Instance(instance);
+    auto const cppWriter = to_FlowWriter(writer);
```
KimonHoffmann, line 237: As above, both for `to_Instance()` as well as `to_Writer()`.

KimonHoffmann:
> I think the usecase would be "hitless" upgrades or replacements. I don't know how realistic such a scenario is, but: In your diagram you would still have downtime when you upgrade the Unifier, because you need to do a SDK major version change or have a critical security issue.

Yes, you're right. I misunderstood the requirement yesterday.

lib/include/mxl/flow.h
```diff
@@ -138,23 +138,18 @@ extern "C"
...
-     *     The flow ID is read from the <id> field of this json object.
-     * \param[in] options Additional options (undefined).
-     *     \todo Specify and used the additional options.
-     * \param[out] info A pointer to an mxlFlowConfigInfo structure.
-     *     If not the nullptr, this structure will be updated with the flow
-     *     information after the flow is created.
-     */
     MXL_EXPORT
-    mxlStatus mxlCreateFlow(mxlInstance instance, char const* flowDef, char const* options, mxlFlowConfigInfo* info);
+    mxlStatus mxlCreateFlowWriter(mxlInstance instance, char const* flowDef, char const* options, mxlFlowWriter* writer,
```
vt-tv, line 154: could we get some doxygen documentation back?

lib/internal/include/mxl-internal/SharedMemory.hpp
```diff
@@ -245,6 +278,16 @@ namespace mxl::lib
         return _data;
     }
 
+    constexpr bool SharedMemoryBase::isExclusive() const
+    {
+        if (_lockType == LockType::None)
+        {
+            throw std::runtime_error("");
```
vt-tv: no description?

jonasohland: Oops!

lib/internal/include/mxl-internal/FlowManager.hpp
```diff
@@ -63,10 +63,10 @@ namespace mxl::lib
         /// \param[in] maxSyncBatchSizeHintOpt Optional max sync batch size hint.
         /// \param[in] maxCommitBatchSizeHintOpt Optional max commit batch size hint
         ///
```
pedro-alves-ferreira, line 67: It would be nice to document the return value, especially what the bool means.

lib/internal/include/mxl-internal/FlowManager.hpp
```diff
@@ -81,8 +81,8 @@ namespace mxl::lib
...
```
pedro-alves-ferreira, line 87: Same comment about the return value doc.

lib/internal/include/mxl-internal/SharedMemory.hpp
```diff
@@ -69,7 +71,7 @@ namespace mxl::lib
          * \param payloadSize The minimum expected size of the shared memory
          * \throw If opening or creating the shared memory segment fails.
          */
-        SharedMemoryBase(char const* path, AccessMode mode, std::size_t payloadSize);
+        SharedMemoryBase(char const* path, AccessMode mode, std::size_t payloadSize, bool exclusive);
```
pedro-alves-ferreira: I'd rather using an `enum class` for `exclusive`, here and in the functions that call this.

lib/internal/tests/test_flowmanager.cpp
```diff
@@ -92,20 +96,20 @@ TEST_CASE_PERSISTENT_FIXTURE(mxl::tests::mxlDomainFixture, "Flow Manager : Creat
...
-        }());
-
-    // This should throw since the flow metadata will already exist.
-    REQUIRE_THROWS(
-        [&]()
-        {
-            auto const sampleRate = mxlRational{48000, 1};
-            manager->createContinuousFlow(flowId, flowDef, MXL_DATA_FORMAT_AUDIO, sampleRate, 8, sizeof(float), 8192);
-        }());
+    // // This should throw since the flow metadata will already exist.
```
pedro-alves-ferreira: Do you want to keep this commented out?

lib/internal/tests/test_flowmanager.cpp
```diff
@@ -169,21 +173,21 @@ TEST_CASE_PERSISTENT_FIXTURE(mxl::tests::mxlDomainFixture, "Flow Manager : Creat
     REQUIRE(!exists(grainDir));
 
     // This should throw since the flow metadata will already exist.
-    REQUIRE_THROWS(
-        [&]()
-        {
-            manager->createContinuousFlow(flowId, flowDef, MXL_DATA_FORMAT_AUDIO, sampleRate, 8, sizeof(float), 8192);
-        }());
+    // REQUIRE_THROWS(
```
pedro-alves-ferreira: Same here.

lib/include/mxl/flow.h
```diff
@@ -138,6 +138,18 @@ extern "C"
...
+     * Attempt to create a flow writer for a given flow definition. If the flow does not exist already, it is created and 'created' will be set to
+     * true. If the flow exists, it will be opened and 'created' will be set to false. If the flow exists already it is not guaranteed that the flow
+     * definition of the existing flow is exactly equal to the flow definition supplied in flowDef. The definition of the existing flow can be
+     * retrieved using mxlGetFlowDef().
+     * \param[in] instance The mxl instance created using mxlCreateInstance
+     * \param[in] flowDef The flow definition from which a flow should be created if there is not already a flow with the same flow id.
+     * \param[in] options (optional) Additional options, can be NULL
+     * \param[out] writer A pointer to a memory location where the created flow writer will be written.
+     * \param[out] configInfo (optional) Used to return information about the opened flow. Can be NULL.
+     * \param[out] created (optional) If not NULLL, will be set to true if a new flow was created, false if an existing flow was opened.
```
pedro-alves-ferreira: Extra L in NULL

lib/internal/src/Deferred.hpp
```diff
@@ -0,0 +1,52 @@
...
+    template<typename F>
+    class Deferred
+    {
+    public:
+        Deferred(Deferred<F> const&) = delete;
+        Deferred(Deferred<F>&&) = delete;
+        Deferred& operator=(Deferred<F> const&) = delete;
+        Deferred& operator=(Deferred<F>&&) = delete;
+
+        /** \brief Calles the deferred function
```
mlefebvre1: Calls

lib/internal/src/Instance.cpp
```diff
@@ -215,6 +202,13 @@ namespace mxl::lib
                     if ((*pos).second.releaseReference())
                     {
                         removeFlowWatch = (dynamic_cast<ContinuousFlowWriter*>((*pos).second.get()) == nullptr);
+
+                        // Delete if flow if we are the last writer.
```
mlefebvre1: Delete flow if we are the last writer.

lib/internal/tests/test_flowmanager.cpp
```diff
@@ -530,9 +532,109 @@ TEST_CASE_PERSISTENT_FIXTURE(mxl::tests::mxlDomainFixture, "FlowManager: createF
...
+    {
+        for (auto i = 0; i < numIterations; ++i)
+        {
+            // Wait for the main threads signal to start
+            while (!worker->readySignal->exchange(false))
+            {
+            }
+
+            {
+                // Create or open the flow. This is where the race happenes. Only one of the workers should be able to create the flow (created ==
```
mlefebvre1: happens

PR #291, opened 2025-12-17
Add Rust GStreamer plugins

pedro-alves-ferreira:
This PR adds GStreamer source and sink plugins.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,9 +1,89 @@
 # SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project.
 # SPDX-License-Identifier: Apache-2.0
-
-Checks: "-*,modernize-*,-modernize-use-trailing-return-type"
-HeaderFileExtensions: ["", "h", "hpp"]
-ImplementationFileExtensions: ["cpp"]
+---
+Checks: >-
+    -*,
+    clang-diagnostic-*,
+    bugprone-*,
+    cert-*,
+    misc-*,
+    modernize-*,
+    performance-*,
+    readability-*,
+    android-cloexec-*,
+    cppcoreguidelines-avoid-const-or-ref-data-members,
+    cppcoreguidelines-avoid-goto,
+    cppcoreguidelines-interfaces-global-init,
+    cppcoreguidelines-slicing,
+    hicpp-exception-baseclass,
+    -clang-diagnostic-address-of-packed-member,
+    -clang-diagnostic-unknown-pragmas,
+    -bugprone-easily-swappable-parameters,
+    -bugprone-reserved-identifier,
+    -cert-dcl16-c,
+    -cert-dcl21-cpp,
+    -cert-dcl37-c,
+    -cert-dcl51-cpp,
+    -cert-env33-c,
+    -misc-include-cleaner,
+    -misc-non-private-member-variables-in-classes,
+    -misc-no-recursion,
+    -modernize-avoid-c-arrays,
+    -modernize-concat-nested-namespaces,
+    -modernize-raw-string-literal,
+    -modernize-replace-disallow-copy-and-assign-macro,
+    -modernize-replace-random-shuffle,
+    -modernize-use-default-member-init,
+    -modernize-use-equals-delete,
+    -modernize-use-nodiscard,
+    -modernize-use-override,
+    -modernize-use-trailing-return-type,
+    -performance-avoid-endl,
+    -performance-enum-size,
+    -performance-noexcept-destructor,
+    -performance-noexcept-move-constructor,
+    -performance-noexcept-swap,
+    -readability-const-return-type,
+    -readability-convert-member-functions-to-static,
+    -readability-else-after-return,
+    -readability-identifier-length,
+    -readability-identifier-naming,
+    -readability-implicit-bool-conversion,
+    -readability-magic-numbers,
+    -readability-make-member-function-const,
+    -readability-named-parameter,
+    -readability-qualified-auto,
+    -readability-redundant-access-specifiers,
+    -readability-redundant-member-init,
+    -readability-static-accessed-through-instance,
+    -readability-use-anyofallof
+WarningsAsErrors:      ''
+HeaderFilterRegex:     '/(lib|bin)/'
+HeaderFileExtensions:  ["h", "hpp"]
+ImplementationFileExtensions: ["c", "cpp"]
 CheckOptions:
-  modernize-use-using.IgnoreExternC: "true"
-  modernize-use-using.IgnoreMacros: "true"
+  - key:            bugprone-dangling-handle.HandleClasses
+    value:          'std::basic_string_view;std::experimental::basic_string_view'
+  - key:            cppcoreguidelines-special-member-functions.AllowSoleDefaultDtor
+    value:          true
+  - key:            cppcoreguidelines-special-member-functions.AllowMissingMoveFunctions
+    value:          true
+  - key:            modernize-use-auto.RemoveStars
+    value:          '1'
+  - key:            modernize-use-using.IgnoreExternC
+    value:          true
+  - key:            modernize-use-using.IgnoreMacros
+    value:          true
+  - key:            readability-braces-around-statements.ShortStatementLines
+    value:          '2'
+  - key:            readability-function-cognitive-complexity.Threshold
+    value:          '50'
+  - key:            readability-function-cognitive-complexity.DescribeBasicIncrements
+    value:          false
+  - key:            readability-implicit-bool-conversion.AllowConditionalIntegerCasts
+    value:          '1'
+  - key:            readability-implicit-bool-conversion.AllowConditionalPointerCasts
+    value:          '1'
+  - key:            readability-uppercase-literal-suffix.NewSuffixes
+    value:          'f;L;LL;U;UL,ULL,LL'
+...
--- a/.devcontainer/Dockerfile.debianforky
+++ b/.devcontainer/Dockerfile.debianforky
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project.
 # SPDX-License-Identifier: Apache-2.0
 
-FROM debian:trixie-slim
+FROM debian:forky-slim
 
 ARG DEBIAN_FRONTEND=noninteractive
 
@@ -41,7 +41,7 @@ RUN apt-get update \
     htop \
     librdmacm-dev
 
-ARG CLANG_VERSION=19
+ARG CLANG_VERSION=21
 RUN apt-get install -y --no-install-recommends \
     "clang-${CLANG_VERSION}" \
     "clang-tools-${CLANG_VERSION}" \
--- a/.devcontainer/debianforky/devcontainer.json
+++ b/.devcontainer/debianforky/devcontainer.json
@@ -1,0 +1,53 @@
+// SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project.
+// SPDX-License-Identifier: Apache-2.0
+{
+    "name": "mxl - Debian Forky",
+    "build": {
+        "dockerfile": "../Dockerfile.debianforky",
+        "context": "../",
+        "args": {}
+    },
+    "runArgs": [
+        "--security-opt",
+        "apparmor:unconfined"
+    ],
+    "mounts": [
+        // These 2 mount points are required for GUI applications to work in WSL2. Remove them if you are not using WSL2.
+        "source=/tmp/.X11-unix,target=/tmp/.X11-unix,type=bind,consistency=cached",
+        "source=/mnt/wslg,target=/mnt/wslg,type=bind,consistency=cached"
+    ],
+    "customizations": {
+        "vscode": {
+            "extensions": [
+                "ms-vscode.cpptools-extension-pack",
+                "mhutchie.git-graph",
+                "matepek.vscode-catch2-test-adapter",
+                "llvm-vs-code-extensions.vscode-clangd",
+                "xaver.clang-format",
+                "bierner.markdown-mermaid",
+                "rust-lang.rust-analyzer"
+            ],
+            "settings": {
+                "C_Cpp.intelliSenseEngine": "disabled",
+                "editor.inlayHints.enabled": "off",
+                "cmake.copyCompileCommands": "${workspaceFolder}/build/compile_commands.json",
+                "editor.formatOnSave": true,
+                "[cpp]": {
+                    "editor.defaultFormatter": "xaver.clang-format"
+                },
+                "rust-analyzer.linkedProjects": [
+                    "${workspaceFolder}/rust/Cargo.toml"
+                ]
+            }
+        }
+    },
+    "containerEnv": {
+        "NINJA_STATUS": "[%r processes, %f/%t @ %o/s : %es] ",
+        "CCACHE_DIR": "${containerWorkspaceFolder}/.cache/ccache",
+        "DISPLAY": "${localEnv:DISPLAY}"
+    },
+    "hostRequirements": {
+        "gpu": "optional"
+    },
+    "remoteUser": "devcontainer"
+}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,10 +55,123 @@ project(mxl
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
 
 
+
+# Enable testing for the project
+enable_testing()
+
+# Set macOS specific configuration
+if(APPLE)
+    # Add the official GStreamer framework paths to the PKG_CONFIG_PATH
+    set(ENV{PKG_CONFIG_PATH} "/Library/Frameworks/GStreamer.framework/Versions/1.0/lib/pkgconfig:$ENV{PKG_CONFIG_PATH}")
+endif()
+
+# Default build is shared libraries.
+# Use '-DBUILD_SHARED_LIBS=OFF' for static build.
+option(BUILD_SHARED_LIBS "Build using shared libraries" ON)
+
+# Default PIC ON for static builds.
+# Use -DMXL_ENABLE_PIC=OFF to disable PIC
+option(MXL_ENABLE_PIC "Enable PIC for static libraries" ON)
+
+# Policy: shared always PIC, static set to MXL_ENABLE_PIC.
+if(NOT BUILD_SHARED_LIBS)
+    set(MXL_POSITION_INDEPENDENT_CODE ${MXL_ENABLE_PIC})
+else()
+    set(MXL_POSITION_INDEPENDENT_CODE ON)
+endif()
+
+#-------------------------------------------------------------------------------
+# Helper to conditionally add flags supported by the current compiler
+#-------------------------------------------------------------------------------
+include(CheckCCompilerFlag)
+include(CheckCXXCompilerFlag)
+
+function(mxl_check_and_add_flag flag)
+    string(REGEX REPLACE "^-+" "" var "${flag}")
+    string(MAKE_C_IDENTIFIER "${var}" var)
+    string(TOUPPER "${var}" var)
+
+    set(c_var "MXL_C_COMPILER_SUPPORTS_${var}")
+    set(cxx_var "MXL_CXX_COMPILER_SUPPORTS_${var}")
+
+    set(lang_list "")
+    check_c_compiler_flag("${flag}" ${c_var})
+    if (${c_var})
+        list(APPEND lang_list C)
+    endif()
+
+    check_cxx_compiler_flag("${flag}" ${cxx_var})
+    if (${cxx_var})
+        list(APPEND lang_list CXX)
+    endif()
+
+    list(JOIN lang_list "," lang_list_string)
+    if (NOT "${lang_list_string}" STREQUAL "")
+        add_compile_options($<$<COMPILE_LANGUAGE:${lang_list_string}>:${flag}>)
+    endif()
+endfunction()
+
+#-------------------------------------------------------------------------------
+# Globally add warning flags
+#-------------------------------------------------------------------------------
+mxl_check_and_add_flag("-Wall")
+mxl_check_and_add_flag("-Wextra")
+mxl_check_and_add_flag("-Wunreachable-code")
+mxl_check_and_add_flag("-Wpedantic")
+mxl_check_and_add_flag("-Wno-c++23-attribute-extensions")
+
+
+#-------------------------------------------------------------------------------
+# Try to determine a senisble target architecture that is reasonably modern.
+#-------------------------------------------------------------------------------
+set(MXL_TARGET_ARCH_HELPER)
+if (CMAKE_SYSTEM_PROCESSOR MATCHES "^((x86)|(X86)|(amd)|(AMD))")
+    check_cxx_compiler_flag("-march=x86-64-v3" MXL_CXX_COMPILER_SUPPORTS_PSABI_V3)
+    if (MXL_CXX_COMPILER_SUPPORTS_PSABI_V3)
+        set(MXL_TARGET_ARCH_HELPER x86-64-v3)
+    endif()
+elseif (CMAKE_SYSTEM_PROCESSOR MATCHES "^((arm)|(ARM)|(aarch)|(AARCH))")
+    check_cxx_compiler_flag("-march=armv8.5-a" MXL_CXX_COMPILER_SUPPORTS_ARM_V85A)
+    if (MXL_CXX_COMPILER_SUPPORTS_ARM_V85A)
+        set(MXL_TARGET_ARCH_HELPER armv8.5-a)
+    endif()
+endif()
+
+set(MXL_TARGET_ARCH "${MXL_TARGET_ARCH_HELPER}" CACHE STRING "The target architecture to generate code for.")
+
+if (MXL_TARGET_ARCH)
+    add_compile_options($<$<COMPILE_LANGUAGE:C,CXX>:-march=${MXL_TARGET_ARCH}>)
+endif()
+
+#-------------------------------------------------------------------------------
+# If specified try to add an option specifying the more specific target to tune
+# for.
+#-------------------------------------------------------------------------------
+set(MXL_TUNE_TARGET "" CACHE STRING "The target architecture to tune the generated code for, while keeping it compatible with the base architecture.")
+
+if (MXL_TUNE_TARGET)
+    if (MXL_TARGET_ARCH)
+        check_cxx_compiler_flag("-march=${MXL_TARGET_ARCH} -mtune=generic" MXL_CXX_COMPILER_SUPPORTS_TUNE)
+        if (MXL_CXX_COMPILER_SUPPORTS_TUNE)
+            check_cxx_compiler_flag("-march=${MXL_TARGET_ARCH} -mtune=${MXL_TUNE_TARGET}" MXL_CXX_COMPILER_SUPPORTS_TUNE_TARGET)
+            if (MXL_CXX_COMPILER_SUPPORTS_TUNE_TARGET)
+                add_compile_options($<$<COMPILE_LANGUAGE:C,CXX>:-mtune=${MXL_TUNE_TARGET}>)
+            else()
+                message(ERROR "The specified tune target is not compatible with the target architecture.")
+            endif()
+        else()
+            message(WARNING "The target compile does not seem to support CPU specific tuning.")
+        endif()
+    else()
+        message(WARNING "Can not enable CPU tuning without a target architecture.")
+    endif()
+endif()
+
+
 #-------------------------------------------------------------------------------
 # Determine optional IPO support and allow enabling IPO on a per target basis.
 #-------------------------------------------------------------------------------
-option(MXL_ENABLE_IPO "Enable LTO/IPO on the library targets." OFF)
+set(MXL_ENABLE_IPO OFF CACHE BOOL "Enable LTO/IPO on the library targets.")
 
 set(MXL_IPO_SUPPORTED OFF)
 if(MXL_ENABLE_IPO)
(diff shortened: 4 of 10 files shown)


jonasohland:
Thanks for all the work! I was able to get a few basic pipelines working without issues. Will do a proper review of the rust code later.

rust/mxl/src/flow/flowdef.rs
mlefebvre1, line 1: These could be moved and maintained in the `mxl` crate.

rust/gst-mxl-rs/src/flowdef.rs
```diff
@@ -0,0 +1,61 @@
...
+    pub label: String,
+    pub id: String,
+    pub media_type: String,
+    pub sample_rate: SampleRate,
+    pub channel_count: i32,
+    pub bit_depth: u8,
+    pub parents: Vec<String>,
+}
+
+pub struct FlowDef {
```
mlefebvre1: Why not use an enum instead? A flow can't be both at the same time.

rust/gst-mxl-rs/src/flowdef.rs
```diff
@@ -0,0 +1,61 @@
...
+    pub numerator: i32,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct FlowDefAudio {
+    pub description: String,
+    pub format: String,
+    pub tags: HashMap<String, Vec<String>>,
+    pub label: String,
+    pub id: String,
```
mlefebvre1: Could use the `uuid` crate to add validation.

rust/gst-mxl-rs/src/mxlsrc/imp.rs
```diff
@@ -0,0 +1,393 @@
...
+                "Settings mutex poisoned — property change ignored"
+            );
+        }
+    }
+
+    fn property(&self, _id: usize, pspec: &glib::ParamSpec) -> glib::Value {
+        if let Ok(settings) = self.settings.lock() {
+            match pspec.name() {
+                "video-flow" => settings.video_flow.to_value(),
+                "audio-flow" => settings.video_flow.to_value(),
```
mlefebvre1: Should be `settings.audio_flow.to_value()`

rust/gst-mxl-rs/src/mxlsrc/imp.rs
```diff
@@ -0,0 +1,393 @@
...
+    const NAME: &'static str = "GstRsMxlSrc";
+    type Type = mxlsrc::MxlSrc;
+    type ParentType = gst_base::PushSrc;
+}
+
+impl ObjectImpl for MxlSrc {
+    fn properties() -> &'static [glib::ParamSpec] {
+        static PROPERTIES: LazyLock<Vec<glib::ParamSpec>> = LazyLock::new(|| {
+            vec![
+                glib::ParamSpecString::builder("video-flow")
```
mlefebvre1: Should not the attribute be named `video-flow-id`? 
Same applies to `audio-flow`

rust/gst-mxl-rs/src/mxlsrc/imp.rs
```diff
@@ -0,0 +1,393 @@
...
+            )
+        });
+
+        Some(&*ELEMENT_METADATA)
+    }
+
+    fn pad_templates() -> &'static [gst::PadTemplate] {
+        static PAD_TEMPLATES: LazyLock<Result<Vec<gst::PadTemplate>, glib::BoolError>> =
+            LazyLock::new(|| {
+                let caps = gst::Caps::new_any();
```
mlefebvre1: So far, we can only source video with format=v210 and audio with format=F32LE.

rust/gst-mxl-rs/src/mxlsrc/mxl_helper.rs
```diff
@@ -0,0 +1,317 @@
...
+pub(crate) fn get_flow_def(
+    src: &MxlSrc,
+    serde_json: serde_json::Value,
+) -> Result<FlowDef, gst::LoggableError> {
+    let media_type = serde_json
+        .get("media_type")
+        .and_then(|v| v.as_str())
+        .unwrap_or("unknown");
+    let json = match media_type {
+        "video/v210" => {
```
mlefebvre1: The flow type (video, audio or data) is not defined by the `media_type`, but by `format`.
Furthermore, if FlowDef was an enum you could just pass around a FlowDef and visit it seamlessly instead of passing around a serde_json::Value. It would also clean-up the wierd match patterns of `set_json_caps`.

Maybe this will help: `Internally Tagged` https://serde.rs/enum-representations.html

rust/gst-mxl-rs/src/mxlsrc/mxl_helper.rs
```diff
@@ -0,0 +1,317 @@
...
+
+    let mut context = mxlsrc.context.lock().map_err(|e| {
+        gst::error_msg!(
+            gst::CoreError::Failed,
+            ["Failed to get context mutex: {}", e]
+        )
+    })?;
+
+    let start = Instant::now();
+    let reader;
```
mlefebvre1, line 197: Any reason why we don't use the result of the expression instead?

```
let reader = loop {
        match init_mxl_reader(&settings) {
            Ok(r) => {
                break r;
            }
            Err(e) => {
                if start.elapsed() >= Duration::from_secs(5) {
                    return Err(e);
                }
                eprintln!(" Failed to init reader ({e}), retrying...");
                std::thread::sleep(Duration::from_millis(250));
            }
        }
    }
```

rust/gst-mxl-rs/src/mxlsrc/create_video.rs
```diff
@@ -0,0 +1,149 @@
...
+pub(super) const MXL_GRAIN_FLAG_INVALID: u32 = 0x00000001;
+
+pub(crate) fn create_video(
+    src: &MxlSrc,
+    state: &mut State,
+) -> Result<CreateSuccess, gst::FlowError> {
+    let video_state = state.video.as_mut().ok_or(gst::FlowError::Error)?;
+    let current_index;
+    let rate = video_state.grain_rate;
+    {
```
mlefebvre1: What's the purpose of the scope here?
Can't we just do:  `let  current_index = state.instance.get_current_index(&rate);`

rust/gst-mxl-rs/src/mxlsrc/create_video.rs
```diff
@@ -0,0 +1,149 @@
...
+            mxl_index: current_index,
+            gst_time: ts_gst,
+        };
+        video_state.is_initialized = true;
+    }
+
+    let initial_info = &state.initial_info;
+
+    let mut next_frame_index = initial_info.mxl_index + video_state.frame_counter;
+    let _ = initial_info;
```
mlefebvre1: I'm confused by what's the purpose of dropping the shared reference and then getting the same shared reference again.

rust/gst-mxl-rs/src/mxlsrc/create_video.rs
```diff
@@ -0,0 +1,149 @@
...
+    let initial_info = &state.initial_info;
+    let grain_request_time = Instant::now();
+    let real_time_start = SystemTime::now();
```
mlefebvre1: If i'm not mistaken, SystemTime uses CLOCK_REALTIME. By default, MXL uses CLOCK_TAI instead, this might introduce problems on some systems.

rust/gst-mxl-rs/src/mxlsrc/create_video.rs
```diff
@@ -0,0 +1,149 @@
...
+        pts -= initial_info.gst_time;
+        initial_info.gst_time = initial_info.gst_time + ts_gst - prev_pts;
+        pts += initial_info.gst_time;
+    }
+
+    let mut buffer;
+    {
+        let binding = &video_state.grain_reader;
+        trace!("Getting grain with index: {}", next_frame_index);
+        let grain_data = match binding.get_complete_grain(next_frame_index, GET_GRAIN_TIMEOUT) {
```
mlefebvre1: i'm sure you can just do 
`let grain_data = match video_state.grain_reader.get_complete_grain(next_frame_index, GET_GRAIN_TIMEOUT) {`

rust/gst-mxl-rs/Cargo.toml
```diff
@@ -0,0 +1,40 @@
+# SPDX-FileCopyrightText: 2025 2025 Contributors to the Media eXchange Layer project.
+# SPDX-License-Identifier: Apache-2.0
+
+[package]
+name = "gst-mxl-src"
```
mlefebvre1: Should it be:
`name = "gst-mxl-rs"`

rust/gst-mxl-rs/src/mxlsrc/create_video.rs
```diff
@@ -0,0 +1,149 @@
...
+
+        {
+            let buffer = buffer.get_mut().ok_or(gst::FlowError::Error)?;
+            buffer.set_pts(pts);
+            let mut map = buffer.map_writable().map_err(|_| gst::FlowError::Error)?;
+            map.as_mut_slice().copy_from_slice(grain_data.payload);
+        }
+    }
+
+    trace!("PTS: {:?} GST-CURRENT: {:?}", buffer.pts(), ts_gst);
```
mlefebvre1: As a general comment, using structured logging is better for logs processing.
```
trace!(pts=?buffer.pts(), ts_gst=?ts_gst, buffer=?buffer, "Produced buffer");
```

rust/gst-mxl-rs/src/mxlsrc/create_video.rs
```diff
@@ -0,0 +1,149 @@
...
+    trace!("Produced buffer {:?}", buffer);
+    if video_state.frame_counter == 0 {
```
mlefebvre1: Hmm, can a comment be added here to explain this logic?

rust/gst-mxl-rs/src/mxlsrc/create_audio.rs
```diff
@@ -0,0 +1,317 @@
...
+        .continuous()
+        .map_err(|_| gst::FlowError::Error)?;
+
+    let sample_rate = reader_info
+        .config
+        .common()
+        .sample_rate()
+        .map_err(|_| gst::FlowError::Error)?;
+
+    let batch_size = DEFAULT_BATCH_SIZE.min(continuous_flow_info.bufferLength / 2);
```
mlefebvre1: Do note that `reader_info.config.common().max_commit_batch_size_hint()` indicates the maximum number of samples a writer can commit at once. This `DEFAULT_BATCH_SIZE` could be bigger if left unchecked, thus violating that contract.

rust/gst-mxl-rs/src/mxlsrc/create_audio.rs
```diff
@@ -0,0 +1,317 @@
...
+        &mut state.initial_info,
+        &state.instance,
+        ts_gst,
+        batch,
+        &reader_info,
+        audio_state,
+    );
+
+    let mut head = reader_info.runtime.head_index();
+    wait_for_sample(head, batch, audio_state)?;
```
mlefebvre1, line 57: Any reason why we just don't use  `samples_reader.get_samples()` to wait for samples, use error variants to determine if we are `OutOfRangeTooLate` or `OutOfRangeTooEarly`, and act accordingly.

rust/gst-mxl-rs/src/mxlsrc/create_audio.rs
```diff
@@ -0,0 +1,317 @@
...
+        audio_state,
+        ts_gst,
+        &mut state.initial_info,
+    );
+
+    let is_discont = std::mem::take(&mut audio_state.next_discont);
+
+    let buffer = build_buffer(pts, samples, is_discont, interleaved)?;
+
+    audio_state.batch_counter += 1;
```
mlefebvre1, line 104: Why do we need to keep synchronized a `batch_counter` with the index. Unless I'm missing something, we can infer it from the index.

rust/gst-mxl-rs/src/mxlsrc/create_video.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    trace!(
+        "Grain number: {} | Grain request time: {} µs | Real time start: {} | Real time end: {} | Elapsed wall time: {} ms",
+        next_frame_index,
+        grain_request_time.elapsed().as_micros(),
+        start_hms,
+        end_hms,
+        elapsed_real.as_millis()
+    );
+    let _ = initial_info;
```
mlefebvre1: Could be removed
```
let _ = initial_info;
let initial_info = &state.initial_info;
```

rust/gst-mxl-rs/src/mxlsrc/create_video.rs
```diff
@@ -0,0 +1,149 @@
...
+    let initial_info = &state.initial_info;
+    let pts = (video_state.frame_counter) as u128 * 1_000_000_000u128;
+    let pts = pts * rate.denominator as u128;
+    let pts = pts / rate.numerator as u128;
+
+    let pts = gst::ClockTime::from_nseconds(pts as u64);
+
+    let mut pts = pts + initial_info.gst_time;
+    let _ = initial_info;
```
mlefebvre1: Remove `let _ = initial_info;` 
You are already dropping the shared reference when you create the exclusive reference at line 107.

rust/gst-mxl-rs/src/mxlsink/render_video.rs
```diff
@@ -0,0 +1,102 @@
...
+        &state
+            .flow
+            .as_ref()
+            .ok_or(gst::FlowError::Error)?
+            .common()
+            .grain_rate()
+            .map_err(|_| gst::FlowError::Error)?,
+    );
+    let video_state = state.video.as_mut().ok_or(gst::FlowError::Error)?;
+    let gst_time = mxlsink
```
mlefebvre1, line 26: Can you explain where and the how the clock type is set from this plugin. If it is using `GST_CLOCK_TYPE_MONOTONIC`, in the rare case of a leap of second, there could be an offset between this plugin producer and a consumer that uses `mxl/flow.h` (which uses CLOCK_TAI instead).

rust/gst-mxl-rs/src/mxlsink/render_video.rs
```diff
@@ -0,0 +1,102 @@
...
+                .map_err(|_| gst::FlowError::Error)?
+                + initial_info.index;
+
+            trace!(
+                "PTS {:?} mapped to grain index {}, current index is {} and running time is {} delta= {}",
+                pts,
+                index,
+                current_index,
+                gst_time,
+                if pts > gst_time {
```
mlefebvre1: ` pts.saturating_sub(gst_time)`

rust/gst-mxl-rs/src/mxlsink/render_video.rs
```diff
@@ -0,0 +1,102 @@
...
+    let mut copy_len = std::cmp::min(payload.len(), data.len());
+    let commit_time = Instant::now();
+    payload[..copy_len].copy_from_slice(&data[..copy_len]);
+    if copy_len > access.total_slices() as usize {
+        copy_len = access.total_slices() as usize;
+    }
+    access
+        .commit(copy_len as u16)
+        .map_err(|_| gst::FlowError::Error)?;
+    trace!(
```
mlefebvre1: Just wondering, is this log still relevant? Why would you need to measure the time it takes copy and commit?

rust/gst-mxl-rs/src/mxlsink/render_video.rs
```diff
@@ -0,0 +1,102 @@
...
+    let data = map.as_slice();
+    let mut access = video_state
+        .writer
+        .open_grain(index)
+        .map_err(|_| gst::FlowError::Error)?;
+    let payload = access.payload_mut();
+    let mut copy_len = std::cmp::min(payload.len(), data.len());
+    let commit_time = Instant::now();
+    payload[..copy_len].copy_from_slice(&data[..copy_len]);
+    if copy_len > access.total_slices() as usize {
```
mlefebvre1: We're comparing the payload size (in bytes) vs the number of slices. This will always be true ... 

If you still want to do the check, you will have to obtain the number of bytes per slice which you can get from the field `sliceSizes` of `DiscreteFlowConfigInfo`.

rust/gst-mxl-rs/src/mxlsink/render_audio.rs
```diff
@@ -0,0 +1,204 @@
...
+    let mut write_index = current_index;
+    if let Some(pts) = buffer.pts() {
+        let abs_pts = pts + initial_info.gst_time;
+        write_index = state
+            .instance
+            .timestamp_to_index(abs_pts.nseconds(), &sample_rate)
+            .map_err(|_| gst::FlowError::Error)?
+            + initial_info.index;
+
+        if write_index > current_index + buffer_length {
```
mlefebvre1: use std::cmp::min instead?

rust/gst-mxl-rs/src/mxlsink/render_audio.rs
```diff
@@ -0,0 +1,204 @@
...
+            write_index = current_index + buffer_length - 1;
+        }
+    }
+
+    trace!(
+        "Writing audio batch starting at index {}, sample_rate {}/{}",
+        write_index, sample_rate.numerator, sample_rate.denominator
+    );
+
+    let max_chunk = (buffer_length / 2) as usize;
```
mlefebvre1, line 52: I suppose it's not possible to configure a max size for the gstreamer buffers we receive?

rust/gst-mxl-rs/src/mxlsink/state.rs
```diff
@@ -0,0 +1,229 @@
...
+
+    let writer = instance
+        .create_flow_writer(flow_id.as_str())
+        .map_err(|e| gst::loggable_error!(CAT, "Failed to create flow writer: {}", e))?
+        .to_samples_writer()
+        .map_err(|e| gst::loggable_error!(CAT, "Failed to create grain writer: {}", e))?;
+    state.audio = Some(AudioState {
+        writer,
+        bit_depth,
+        batch_size: (rate as usize / 100),
```
mlefebvre1: Using mxl flow_info you should be able to retrieve the `max_commit_batch_size_hint` instead of arbitrary setting this.

mlefebvre1:
> Thanks, @mlefebvre1! We will review your comments a.s.a.p.

No worries @pedro-alves-ferreira, most of the comments are suggestions aimed at improving code quality and I would be fine with them being addressed after v1.0. You could focus on resolving these comments for v1.0 :
https://github.com/dmf-mxl/mxl/pull/291#discussion_r2640365217
https://github.com/dmf-mxl/mxl/pull/291#discussion_r2641066915
https://github.com/dmf-mxl/mxl/pull/291#discussion_r2640961983
https://github.com/dmf-mxl/mxl/pull/291#discussion_r2640567491
https://github.com/dmf-mxl/mxl/pull/291#discussion_r2640508810
https://github.com/dmf-mxl/mxl/pull/291#discussion_r2640524689

rust/gst-mxl-rs/src/mxlsrc/create_video.rs
SimaoFonseca2: The variable `real_time_start`  is only used for logging purposes, it was only helpful during very early development so it's going to be removed along with the log itself.

rust/gst-mxl-rs/src/mxlsink/render_video.rs
SimaoFonseca2: The plugin does not propose any clock. It uses the pipeline's clock. It is the user's responsibility to set one or to use the default, which is monotonic, as explained in this [documentation](https://gstreamer.freedesktop.org/documentation/gstreamer/gstclock.html?gi-language=c#GstClock).
In regards to the leap of second case, it is to my understanding that monotonic clocks won't cause an offset between the producer and consumer, that case occurs in real-time clocks.

rust/gst-mxl-rs/src/mxlsink/render_audio.rs
SimaoFonseca2: It's not possible, you can't restrict the max size of the buffers you receive in GStreamer.

rust/gst-mxl-rs/src/mxlsink/render_video.rs
SimaoFonseca2: Agreed, this was only useful for an early development bug and probably should’ve been removed sooner.

jonasohland:
> Thanks for the feedback, @jonasohland. There's still some work to be done around reopening the flow, clean shutdown, etc. but we're almost there.

Flow deletion, reopening and those things should be all automatic now that #290 has been merged. In the rust api everything should be RAII now. Hope that helps!

pedro-alves-ferreira:
> > Thanks for the feedback, @jonasohland. There's still some work to be done around reopening the flow, clean shutdown, etc. but we're almost there.
> 
> Flow deletion, reopening and those things should be all automatic now that #290 has been merged. In the rust api everything should be RAII now. Hope that helps!

Yes, that was quite an improvement. Thanks!

rust/gst-mxl-rs/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+[features]
+tracing = []
+
+[lib]
+name = "gstmxl"
+crate-type = ["cdylib", "rlib"]
+path = "src/lib.rs"
+
+[build-dependencies]
+gst-plugin-version-helper = { path = "../version-helper" }
```
pac-work: If possible, I would avoid copy+paste and went for using the [upstream](https://crates.io/crates/gst-plugin-version-helper) directly instead.

rust/gst-mxl-rs/src/mxlsink/render_video.rs
pac-work: I do not know GStreamer enough to be able to say whether plugin can somehow dictate the clock used for the whole pipeline, but there definitely [exists GST_CLOCK_TYPE_TAI](https://gstreamer.freedesktop.org/documentation/gstreamer/gstsystemclock.html?gi-language=c#GST_CLOCK_TYPE_TAI), which would seem like a good fit for this usecase.

rust/Cargo.toml
```diff
@@ -21,7 +22,8 @@ libloading = { version = "0.8.9" }
 thiserror = "2.0.12"
 tracing = { version = "0.1", features = ["log"] }
 tracing-subscriber = { version = "0.3.20", features = ["env-filter", "std"] }
-uuid = { version = "1.17", features = ["v4"] }
+uuid = { version = "1.17", features = ["v4", "serde"] }
+serde = { version = "1.0", features = ["derive"] }
```
pac-work, line 26: The OCD me thinks these should be alphabetically ordered, but feel free to ignore :) If you decide to fix, this holds for all the `Cargo.toml` files.

rust/gst-mxl-rs/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+[package.metadata.cargo-machete]
+ignored = ["regex"]
+
+[dependencies]
+glib = "0.21.5"
+gstreamer = "0.24.4"
+gstreamer-audio = "0.24.4"
+gstreamer-base = "0.24.4"
+mxl = { path = "../mxl" }
+serde = { version = "1.0", features = ["derive"] }
```
pac-work: Can most of the dependencies here point to the top level workspace, to easily keep the versions in sync? The `mxl` and `mxl_sys` crates are doing so.

rust/gst-mxl-rs/README.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+### Create video/audio flows from a single file
+
+```
+export AV_FILE=<path to file with audio and video, e.g. file.mp4>
+
+gst-launch-1.0 filesrc location="$AV_FILE" ! decodebin name=dec ! videoconvert ! queue ! mxlsink flow-id="$VIDEO_FLOW_ID" domain="$MXL_DOMAIN" dec. ! audioresample ! audioconvert ! queue ! mxlsink flow-id="$AUDIO_FLOW_ID" domain="$MXL_DOMAIN"
+```
+
```
pac-work, line 93: Maybe as an alternative, you can add the following command, which does not require any mp4 file. That is what I was testing with:
```
gst-launch-1.0 videotestsrc ! timeoverlay valignment=center ! clockoverlay time-format=\"%F %H:%M:%S %Z\" ! video/x-raw,width=1920,height=1080,framerate=25/1,format=v216 ! videoconvert ! queue ! mxlsink flow-id="$VIDEO_FLOW_ID" domain="$MXL_DOMAIN" audiotestsrc wave=ticks ! audioconvert ! queue ! mxlsink flow-id="$AUDIO_FLOW_ID" domain="$MXL_DOMAIN"
```

rust/mxl/src/flow/flowdef.rs
```diff
@@ -0,0 +1,62 @@
...
+    pub sample_rate: SampleRate,
+    pub channel_count: i32,
+    pub bit_depth: u8,
+    pub parents: Vec<String>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct GrainRate {
+    pub numerator: i32,
+    pub denominator: i32,
```
pac-work, line 75: Maybe instead of having `GrainRate` and `SampleRate`, we could just get away with `Rate` and 

```
    #[serde(default = "default_denominator")]
    pub denominator: i32,
```
and after the struct
```
fn default_denominator() -> i32 {
    1
}
```

rust/mxl/src/flow/flowdef.rs
```diff
@@ -0,0 +1,62 @@
...
+    pub id: String,
+    pub tags: HashMap<String, Vec<String>>,
+    pub format: String,
+    pub label: String,
+    pub parents: Vec<String>,
+    pub media_type: String,
+    pub grain_rate: GrainRate,
+    pub frame_width: i32,
+    pub frame_height: i32,
+    pub interlace_mode: String,
```
pac-work:
```
pub interlace_mode: InterlaceMode,
```
and then 
```
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
pub enum InterlaceMode {
    #[serde(rename = "progressive")]
    Progressive,
    #[serde(rename = "interlaced_tff")]
    InterlacedTff,
    #[serde(rename = "interlaced_bff")]
    InterlacedBff,
}
```

rust/mxl/src/flow/flowdef.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+pub enum FlowDef {
+    Video(FlowDefVideo),
+    Audio(FlowDefAudio),
+}
+
+#[derive(Debug, Serialize, Deserialize)]
```
pac-work:
```
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
```
?

rust/mxl/src/flow/flowdef.rs
```diff
@@ -0,0 +1,62 @@
+// SPDX-FileCopyrightText: 2025 2025 Contributors to the Media eXchange Layer project.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::collections::HashMap;
+
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+pub enum FlowDef {
```
pac-work: For better deserialization, why not go in direction similar to:
```
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
pub struct FlowDef {
    pub id: uuid::Uuid,
    #[serde(flatten)]
    pub details: FlowDefDetails,
}

#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
#[serde(tag = "media_type")]
pub enum FlowDefDetails {
    #[serde(rename = "video/v210")]
    Video(FlowDefVideo),
    // TODO: Add support for "video/v210a".
    #[serde(rename = "audio/float32")]
    Audio(FlowDefAudio),
}
```

rust/mxl/src/flow/flowdef.rs
```diff
@@ -0,0 +1,62 @@
...
+    Video(FlowDefVideo),
+    Audio(FlowDefAudio),
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct FlowDefVideo {
+    pub description: String,
```
pac-work, line 11: Are all these really required? @vt-tv , do we have a JSON schema for this somewhere?

As I see from the code, the only required are:
- id
- format
- label
- group hint tag
- grain_rate
- frame_width
- frame_height
- interlace_mode is not required, but defaults to progressive

Similar-ish for audio:
- id
- format
- label
- group hint tag
- sample_rate
- channel_count is not required, but defaults to 1

Rest seems to be optional.

rust/gst-mxl-rs/src/mxlsink/imp.rs
```diff
@@ -0,0 +1,407 @@
...
+                        gst::Caps::builder("video/x-raw")
+                            .field("format", "v210")
+                            .build(),
+                    );
+                    for ch in 1..64 {
+                        let mask = gst::Bitmask::from((1u64 << ch) - 1);
+                        caps.make_mut().append(
+                            gst::Caps::builder("audio/x-raw")
+                                .field("format", "F32LE")
+                                .field("layout", "interleaved")
```
pac-work, line 208: I haven't yet read the rest of the implementation, but MXL is native `non-interleaved`, so I am a bit surprised that `interleaved` is used on the plugin interface. Using `non-interleaved` and doing simple memory copies feels like faster operation than re-ordering channels. But I am not sure what is customary in the GStreamer world - whether the `interleaved` is the preferred format which more plugins can operate with, or...?

rust/gst-mxl-rs/src/mxlsink/state.rs
```diff
@@ -0,0 +1,231 @@
...
+    let colorimetry = structure
+        .get::<String>("colorimetry")
+        .unwrap_or_else(|_| "BT709".to_string());
+    let mut tags = HashMap::new();
+    tags.insert(
+        "urn:x-nmos:tag:grouphint/v1.0".to_string(),
+        vec!["Media Function XYZ:Video".to_string()],
+    );
+    let flow_def = FlowDefVideo {
+        description: format!(
```
pac-work: This should probably be configurable through the plugin interface? Same as label below...

rust/gst-mxl-rs/src/mxlsink/state.rs
```diff
@@ -0,0 +1,231 @@
...
+    let rate = info.rate() as i32;
+    let bit_depth = info.depth() as u8;
+    let format = info.format().to_string();
+    let mut tags = HashMap::new();
+    tags.insert(
+        "urn:x-nmos:tag:grouphint/v1.0".to_string(),
+        vec!["Media Function XYZ:Audio".to_string()],
+    );
+    let flow_def = FlowDefAudio {
+        description: "MXL Audio Flow".into(),
```
pac-work, line 100: Same as for video - both description and label should be configurable through the plugin interface. Maybe also group hint tag? Or portion of it? Or base the hint on the label somehow?

rust/gst-mxl-rs/src/mxlsrc/mxl_helper.rs
```diff
@@ -0,0 +1,302 @@
...
+            .ok_or(gst::loggable_error!(CAT, "No audio flow id was found"))?
+    };
+    Ok(id)
+}
+
+pub(crate) fn get_mxl_flow_json(
+    domain: &String,
+    flow_id: &String,
+) -> Result<serde_json::Value, gst::LoggableError> {
+    let json_path = format!("{}/{}.mxl-flow/flow_def.json", domain, flow_id);
```
pac-work: This should not got to the disk directly, but should use the API we already have...  `MxlInstance::get_flow_def`

pedro-alves-ferreira:
> Pardon my lack of knowledge about how gstreamer plugins are usually made, but does the `version-helper` need to be copy-pasted in the repo, or is it possible to directly include [the original](https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/tree/main/version-helper) as one of `build-dependencies` and use it from there?

Thanks, Pavel. Good point. I hadn't realized they had pushed that to crates.io. 
@SimaoPFonseca Can you please replace that with a `build-dependency` to `gst-plugin-version-helper`?

rust/gst-mxl-rs/src/mxlsrc/create_audio.rs
```diff
@@ -0,0 +1,324 @@
...
+}
+
+fn wait_for_sample(
+    mut head: u64,
+    batch: u64,
+    audio_state: &AudioState,
+) -> Result<(), gst::FlowError> {
+    let start = Instant::now();
+    while audio_state.index + batch > head {
+        if start.elapsed() > PRODUCER_TIMEOUT {
```
pac-work, line 153: I assume this busy waiting is here, because the code was written before Kimon added the waiting for data / reading with timeout? If so, probably worth updating.

rust/gst-mxl-rs/src/mxlsink/state.rs
pac-work: Based on discussion with @vt-tv : The hint should be unique, even if it is not supplied by the user.

To quote precisely:
> and default could contain the PID of gstlaunch or similar, ensuring uniqueness.  (or mostly ensuring)

rust/mxl/src/flow/flowdef.rs
pac-work: Schemas can be found [here](https://specs.amwa.tv/is-04/releases/v1.3.3/APIs/schemas/). For example [video flow](https://specs.amwa.tv/is-04/releases/v1.3.3/APIs/schemas/with-refs/flow_video.html). However, first glance showed that even the current examples in the MXL repo are not compliant (missing `source_id`, `device_id`, which are [required for every flow](https://specs.amwa.tv/is-04/releases/v1.3.3/APIs/schemas/with-refs/flow_core.html)). Not sure how to approach this. Probably instead of trying to cover full potential of NMOS, use only the parts required by MXL, while staying open to the other things people may want to specify (flattened `other` member for serialization?)...?

rust/gst-mxl-rs/src/mxlsink/imp.rs
SimaoFonseca2: It's because most GStreamer audio elements expect `interleaved`. In the future, `non-interleaved` will also be an option but for now `interleaved` was the easiest.

rust/gst-mxl-rs/src/mxlsink/render_video.rs
pedro-alves-ferreira: @pac-work An element can propose a clock but there could be several elements that can provide a clock in the same pipeline, so we wouldn't control which one would be selected. We decided not to propose a clock and leave it to the application to select a PTP or TAI clock, should it want to, or even running the pipeline within a [clockselect](https://gstreamer.freedesktop.org/documentation/debugutilsbad/clockselect.html?gi-language=c) bin.

rust/gst-mxl-rs/src/mxlsink/imp.rs
pac-work: Well, in that same line of argumentation, one may say that most audio elements do not support `F32LE`, but `S16LE`, but I see your point. In my eyes, because for example of the float format, one will often feed the audio directly to `audioconvert`, which can do both jobs (buffer layout and data type conversion) at the same time, thus most likely make things more efficient. From that point of view, doing as few conversions as possible and preserving original shape of the data is the better strategy. For example, the MXL GST sink test tool feeds `non-interleaved` data directly to the GST pipeline just fine, without the need for manual sample-by-sample processing.

But considering the real-world impact on performance would have to be measured and may be easily questioned without having both implementations available, I guess `interleaved` is fine for now, although it is in my eyes the worse of the two options (requires more code / algorithm logic, slower).

pedro-alves-ferreira: @pac-work I completely agree. Simão just couldn't get it to work immediately but we can try and give it another shot.

pac-work: > @pac-work I completely agree. Simão just couldn't get it to work immediately but we can try and give it another shot.

I see, well, I have never written GStreamer plugin mysel, but I would assume that inside a plugin, one may fill buffers in similar way as outside of them (for `appsink`), so hopefully the code inside the MXL GST sink test tool could serve as an inspiration. If it is too problematic, just feel free to leave it with `interleaved`. As I mentioned above, hard to guess whether there will be measurable impact.

pedro-alves-ferreira: No, I agree we should get that fixed. Filling the buffers was trivial. IIRC, the problem was around caps negotiation.

GOVERNANCE/Release Process.md
```diff
@@ -0,0 +1,69 @@
+<!-- SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project. -->
```
sviengkhou: Should we move to 2026 now?

rust/mxl/src/flow/flowdef.rs
pedro-alves-ferreira: Done

rust/gst-mxl-rs/src/flowdef.rs
pedro-alves-ferreira: Done

rust/gst-mxl-rs/src/mxlsrc/imp.rs
pedro-alves-ferreira: Done

rust/gst-mxl-rs/src/mxlsrc/imp.rs
pedro-alves-ferreira: Done

rust/gst-mxl-rs/src/mxlsrc/imp.rs
pedro-alves-ferreira: Done

rust/gst-mxl-rs/Cargo.toml
pedro-alves-ferreira: Done

rust/gst-mxl-rs/src/mxlsrc/create_audio.rs
pedro-alves-ferreira: Done

rust/gst-mxl-rs/src/mxlsrc/create_video.rs
pedro-alves-ferreira: Done

rust/gst-mxl-rs/src/mxlsrc/create_video.rs
pedro-alves-ferreira: Done

rust/gst-mxl-rs/src/mxlsink/render_video.rs
pedro-alves-ferreira: Done

rust/gst-mxl-rs/src/mxlsink/render_audio.rs
pedro-alves-ferreira: Done

rust/gst-mxl-rs/src/mxlsink/state.rs
pedro-alves-ferreira: Done

rust/gst-mxl-rs/Cargo.toml
pedro-alves-ferreira: Done

rust/Cargo.toml
pedro-alves-ferreira: Done

rust/gst-mxl-rs/Cargo.toml
pedro-alves-ferreira: Done

rust/gst-mxl-rs/README.md
pedro-alves-ferreira: Done

rust/mxl/src/flow/flowdef.rs
pedro-alves-ferreira: Done

rust/mxl/src/flow/flowdef.rs
pedro-alves-ferreira: Done

rust/mxl/src/flow/flowdef.rs
pedro-alves-ferreira: Done

rust/mxl/src/flow/flowdef.rs
pedro-alves-ferreira: Done

rust/mxl/src/flow/flowdef.rs
pedro-alves-ferreira: We implemented the minimum required set of data.

rust/gst-mxl-rs/src/mxlsink/imp.rs
pedro-alves-ferreira: to be done in a subsequent PR.

rust/gst-mxl-rs/src/mxlsrc/mxl_helper.rs
pedro-alves-ferreira: Done

rust/gst-mxl-rs/src/mxlsrc/create_audio.rs
pedro-alves-ferreira: Done

rust/gst-mxl-rs/src/mxlsink/state.rs
pedro-alves-ferreira: Done

rust/gst-mxl-rs/src/mxlsink/state.rs
pedro-alves-ferreira: Done

rust/gst-mxl-rs/src/mxlsink/render_video.rs
pedro-alves-ferreira: Done

rust/gst-mxl-rs/src/flowdef.rs
pedro-alves-ferreira: Done

rust/gst-mxl-rs/src/mxlsrc/create_video.rs
pedro-alves-ferreira: Done

rust/gst-mxl-rs/src/mxlsrc/create_video.rs
pedro-alves-ferreira: Partially done. Some will be updated eventually.

rust/gst-mxl-rs/src/mxlsrc/create_video.rs
pedro-alves-ferreira: Done

rust/gst-mxl-rs/src/mxlsrc/mxl_helper.rs
pedro-alves-ferreira: The implementation was changed here. The error block no longer returns an error. It instead waits for a flow to be created, or if the pipeline is stopped.
This was changed to avoid reading from stale flows.

rust/gst-mxl-rs/src/mxlsrc/create_video.rs
pedro-alves-ferreira: Done

rust/gst-mxl-rs/src/mxlsrc/create_audio.rs
pedro-alves-ferreira: We will add this to a new issue to implement several improvements.

rust/gst-mxl-rs/src/mxlsrc/create_audio.rs
pedro-alves-ferreira: We will add this to a new issue to implement several improvements.

rust/gst-mxl-rs/src/mxlsrc/create_video.rs
pedro-alves-ferreira: Done

rust/gst-mxl-rs/src/mxlsrc/mxl_helper.rs
mlefebvre1: I'm fine with keeping pattern matching instead of letting serde do it, but can we at least use the correct field, which is `media_type` and not `format`?

pedro-alves-ferreira: This will be added to a new issue to be solved eventually.

rust/mxl/src/flow/flowdef.rs
```diff
@@ -0,0 +1,88 @@
...
+    pub format: String,
+    pub label: String,
+    pub parents: Vec<String>,
+    pub media_type: String,
+    #[serde(flatten)]
+    pub details: FlowDefDetails,
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+#[serde(tag = "media_type")]
```
mlefebvre1: Same as my other comment, shouldn't we use `format` instead of `media_type` as the variant tag?

rust/gst-mxl-rs/src/mxlsrc/mxl_helper.rs
mlefebvre1: Feel free to not address this but, loops are also expressions and can return a value via break.
```
let reader = loop {
        match init_mxl_reader(&settings) {
            Ok(r) => {
                break r;
            }
            Err(_) => {
                if !warned {
                    eprintln!("Waiting for flow to be created...");
                    warned = true;
                }
                //Sleep to avoid busy looping
                std::thread::sleep(Duration::from_millis(10));
            }
        }
    };
```

rust/mxl/src/flow/flowdef.rs
SimaoFonseca2: I managed to replicate the TOO LATE warnings issue i will track down the issue and fix it as soon as possible. I also agree that it should be `format` and not `media_type` will change it as well.

SimaoFonseca2: "TOO LATE" issue is now fixed and variant tag was changed.

rust/gst-mxl-rs/src/mxlsrc/mxl_helper.rs
SimaoFonseca2: This has been changed in the latest commit. The format field is used now.

rust/gst-mxl-rs/src/mxlsrc/create_audio.rs
pac-work: Well, this one is not really done, is it... but I assume it will be under https://github.com/dmf-mxl/mxl/issues/330...

pedro-alves-ferreira: Right! I'm sorry for the wrong comment.

PR #299, opened 2025-12-20
Only use public MXL APIs in the tools projects (mxl-info, mxl-gst-*)

vt-tv:
- only use public MXL APIs in the tools projects
- clean up mxl-info and optionally use MXL URIs. Update docs.
- Update vcpkg dependencies
- Clean up 'v210+alpha' left overs.  Renamed everything to v210a

### DIFF
--- a/tools/mxl-gst/looping_filesrc.cpp
+++ b/tools/mxl-gst/looping_filesrc.cpp
@@ -17,7 +17,7 @@
 #include <mxl/flow.h>
 #include <mxl/mxl.h>
 #include <mxl/time.h>
-#include "./utils.hpp"
+#include "utils.hpp"
 
 namespace fs = std::filesystem;
 
--- a/tools/mxl-gst/sink.cpp
+++ b/tools/mxl-gst/sink.cpp
@@ -23,7 +23,7 @@
 #include <mxl/mxl.h>
 #include <mxl/time.h>
 #include "mxl/rational.h"
-#include "./utils.hpp"
+#include "utils.hpp"
 
 #ifdef __APPLE__
 #   include <TargetConditionals.h>
--- a/tools/mxl-gst/utils.hpp
+++ b/tools/mxl-gst/utils.hpp
@@ -37,7 +37,6 @@ namespace media_utils
 
 namespace json_utils
 {
-
     /**
      * Parse a JSON buffer and return the root object.  Throws on error.
      * @param jsonBuffer The JSON buffer to parse.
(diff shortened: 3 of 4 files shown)


tools/mxl-gst/looping_filesrc.cpp
```diff
@@ -17,7 +17,7 @@
 #include <mxl/flow.h>
 #include <mxl/mxl.h>
 #include <mxl/time.h>
-#include "mxl-internal/Logging.hpp"
+#include "./utils.hpp"
```
pedro-alves-ferreira: Do we really need the  `./`?

tools/mxl-gst/sink.cpp
```diff
@@ -17,12 +18,12 @@
...
 #include <gst/gstsystemclock.h>
+#include <picojson/picojson.h>
 #include <mxl/flow.h>
 #include <mxl/mxl.h>
 #include <mxl/time.h>
-#include "mxl-internal/FlowParser.hpp"
-#include "mxl-internal/Logging.hpp"
-#include "mxl-internal/PathUtils.hpp"
+#include "mxl/rational.h"
+#include "./utils.hpp"
```
pedro-alves-ferreira: Same comment about the relative path. If there is any collision, I'd rather changing the file name.

tools/mxl-gst/utils.hpp
```diff
@@ -0,0 +1,171 @@
...
+{
+    std::uint32_t getV210LineLength(std::size_t width)
+    {
+        return static_cast<std::uint32_t>((width + 47) / 48 * 128);
+    }
+}
+
+namespace json_utils
+{
+
```
pedro-alves-ferreira: Extra line here

tools/mxl-gst/looping_filesrc.cpp
vt-tv: Fixed

tools/mxl-gst/utils.hpp
```diff
@@ -0,0 +1,170 @@
+// SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project.
```
mlefebvre1, line 1: This could be re-used by any other tools that are not necessary in `mxl-gst` such as `mxl-fabrics-demo` and probably any user that uses C++. Can we just make it a library?

vt-tv: The goal is to make this lightweight, header-only and not officially supported as part of the SDK.  this is just _convenience_ code.   If the mxl-fabrics-demo tool needs it we could move it up a directory and include it.

PR #300, opened 2025-12-21
Fix SIGBUS crash when there is not enough filesystem space

jonasohland:
Use `posix_fallocate` to resize the SHM files before mapping them.
`ftruncate` will succeed even if there is not enough space on the filesystem, and the program will crash when the memory is first used.

To reproduce:
```
$ sudo mount --mkdir -o size=1K -t tmpfs x /mnt/test
$ build/Linux-GCC-Debug/tools/mxl-gst/mxl-gst-testsrc -v lib/tests/data/v210_flow.json -d /mnt/test
Bus error (core dumped)
```

After the fix:
```
$ sudo mount --mkdir -o size=1K -t tmpfs x /mnt/test
$ build/Linux-GCC-Debug/tools/mxl-gst/mxl-gst-testsrc -v lib/tests/data/v210_flow.json -d /mnt/test
[2025-12-21 23:34:37.980] [console] [error] [flow.cpp:52] Failed to create flow : Shared memory segment does not meet the minimum size requirements.: Cannot allocate memory
[2025-12-21 23:34:38.231] [console] [error] [testsrc.cpp:919] Error while processing video pipeline: Failed to create flow with status code 1.
```

I have only observed this on arm64, but this could also be an issue on amd64.

### DIFF
--- a/lib/internal/src/SharedMemory.cpp
+++ b/lib/internal/src/SharedMemory.cpp
@@ -30,11 +30,11 @@ namespace mxl::lib
         if ((mode == AccessMode::CREATE_READ_WRITE) && ((_fd = ::open(path, OMODE_CREATE, 0664)) != -1))
         {
             // Ensure the file is large enough to hold the shared data
-            if (::posix_fallocate(_fd, 0, payloadSize) == -1)
+            if (auto error = ::posix_fallocate(_fd, 0, payloadSize); error != 0)
             {
                 // No need to close _fd, as the destructor *will* be called,
                 // because we delegated to the default constructor.
-                throw std::system_error(errno, std::generic_category(), "Could not resize shared memory segment.");
+                throw std::system_error(error, std::generic_category(), "Could not resize shared memory segment.");
             }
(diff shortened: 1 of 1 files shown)


lib/internal/src/SharedMemory.cpp
```diff
@@ -30,7 +30,7 @@ namespace mxl::lib
         if ((mode == AccessMode::CREATE_READ_WRITE) && ((_fd = ::open(path, OMODE_CREATE, 0664)) != -1))
         {
             // Ensure the file is large enough to hold the shared data
-            if (::ftruncate(_fd, payloadSize) == -1)
+            if (::posix_fallocate(_fd, 0, payloadSize) == -1)
```
mlefebvre1: From the man pages : 

RETURN VALUE
       posix_fallocate() returns zero on success, or an error number on failure.  **Note that errno is not set.**

So, we need to change `errno` to the returned value in the exception below.

jonasohland: Thanks! I will change that.

PR #302, opened 2025-12-22
Better SemVer support in preparation for first 1.0 RC.

vt-tv:
Add fully rendered version string that conforms to the semver spec (added prerelease suffix and used the + symbol for appending build #).  Add the git hash to the full version, making builds fully traceable.

### DIFF
--- a/lib/include/mxl/mxl.h
+++ b/lib/include/mxl/mxl.h
@@ -53,7 +53,7 @@ extern "C"
         uint16_t minor;
         uint16_t bugfix;
         uint16_t build;
-        char const* fullVersion; /* owned by the library */
+        char const* full; /* owned by the library */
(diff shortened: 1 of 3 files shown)


lib/include/mxl/mxl.h
```diff
@@ -53,6 +53,7 @@ extern "C"
         uint16_t minor;
         uint16_t bugfix;
         uint16_t build;
+        char const* fullVersion; /* owned by the library */
```
jonasohland: All other members don't have the `Version` suffix, so either this should be just `full` or the other should be like: `majorVersion` `minorVersion` `buildNumber`.

But not critical at all, just my personal taste.

vt-tv: changed.

Issue #321, opened 2026-01-13
Generation of position independent code is unintentionally disabled by default


KimonHoffmann:
The way `MXL_POSITION_INDEPENDENT_CODE` is set in the root-level `CMakeLists.txt` makes it default to `OFF`, when `BUILD_SHARED_LIBS` is not specified explicitly on the command line.

**Expected behavior**
When building shared shared libraries (which is the default, when `BUILD_SHARED_LIBS` is left unspecified on the command line) the value of `MXL_POSITION_INDEPENDENT_CODE` should default to `ON`.

PR #322, opened 2026-01-13
Various improvements to the cmake build description

KimonHoffmann:
In addition to fixing #321, this PR adds a few independent features to the cmake build description that also make a few clumsy workarounds previously employed unnecessary.

This PR is ideally reviewed by reviewing each commit individually.

It still needs to be tested, whether newly introduced features work for all build variants, especially when building static libraries.

### DIFF
--- a/.devcontainer/debianforky/devcontainer.json
+++ b/.devcontainer/debianforky/devcontainer.json
@@ -1,0 +1,53 @@
+// SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project.
+// SPDX-License-Identifier: Apache-2.0
+{
+    "name": "mxl - Debian Forky",
+    "build": {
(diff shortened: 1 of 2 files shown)


.devcontainer/debianforky/devcontainer.json
```diff
@@ -0,0 +1,53 @@
+// SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project.
+// SPDX-License-Identifier: Apache-2.0
+{
+    "name": "mxl - Debian Trixie",
```
vt-tv: Change name to "mxl - Debian Forky" ?

KimonHoffmann: Good catch! Fixed.

PR #325, opened 2026-01-15
Add a workflow to automate backports

jonasohland:
This adds a workflow that will be triggered when a pull request is merged that has a label that matches the regex: `^backport/([^ ]+)$`.
The first capture group of that regex will be interpreted as  a target branch for which a new pull request will be created by the bot, which cherry-picks all commits from the original PR on this target branch.

So a label `backport/release/v1.0` will create a PR to `release/v1.0` with the 'contents' of the original PR.

I have tried this on a Toy repository here: https://github.com/jonasohland/mxl-backport-testing

Here is an example of a PR that should be backported: https://github.com/jonasohland/mxl-backport-testing/pull/8
Here is an example or a PR created by the workflow to do the backport: https://github.com/jonasohland/mxl-backport-testing/pull/9

### DIFF
--- a/.github/workflows/backports.yml
+++ b/.github/workflows/backports.yml
@@ -35,8 +35,8 @@ jobs:
       - name: Create backport pull requests
         uses: korthout/backport-action@v4
         with:
-          label_pattern: a^
-          target_branches: ${{ matrix.target-branch }}
+          label_pattern: a^ # We don't need the action to match the labels, we have done that in the prev. job, so we use the pattern 'a^' that never matches anything.
+          target_branches: ${{ matrix.target-branch }} # We pass the branches here instead of relying on the label matching logic of the imported action.
           branch_name: backport/${pull_number}/-/${target_branch}
           copy_assignees: true
           copy_requested_reviewers: true


jonasohland:
> Could we make the `release/` part implicit? I.e. label `backport/v1.1` implies `release/v1.1` as the target branch?

Not natively with the workflow that this PR importing, but I can probably code something around that limitation.

.github/workflows/backports.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Get target branch
+        id: target-branch
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR_NUM: ${{ github.event.number }}
+        run: |
+          labels="$(gh api repos/dmf-mxl/mxl/pulls/${PR_NUM} | jq -r '.labels | .[].name' | grep 'backport/.*')" || true
+          if [[ -z "${labels}" ]]; then echo "No backport labels"; exit 0; fi
+          echo "target-branches=$(sed -nE 's%backport/(.*)%release/\1%p' <<<"${labels}" | tr '\n' ' ' | xargs)" >>$GITHUB_OUTPUT
+          echo "backport=true" >> $GITHUB_OUTPUT
```
KimonHoffmann: Assuming my suggestion from before is applied the workflow would not longer need to decide whether or not it needs to run, but only determine the target branches to merge to.

I instead of querying the labels from the API using an explicit, relative URL the run step can probably just use:

```yaml
run: |
  printf 'target-branches=%s\n' "$(sed -nE 's%backport/(.*)%release/\1%p' <<< "${{ join(github.event.pull_request.labels.*.name, '\n') }}")" >> "${GITHUB_OUTPUT}"
```

When also preparing for the "label-after-merge" case I mentioned in my other comment expression providing the label name needs to be modified to something like:

```
${{ (github.event.action == 'labeled') && github.event.label.name || join(github.event.pull_request.labels.*.name, '\n') }}
```

All written in this edit box without any form of verifications ... so please take this with a grain of salt.

KimonHoffmann: It might make sense to turn `target_branches` into a matrix axis.
Then each matrix job could just focus on a single target branch and also be conditioned on the *absence* of another label, that the jobs adds automatically to indicate the backport has already been performed.

jonasohland: > It might make sense to turn `target_branches` into a matrix axis. Then each matrix job could just focus on a single target branch and also be conditioned on the _absence_ of another label, that the jobs adds automatically to indicate the backport has already been performed.

I can do that, my only fear is that we already regularly run into a rate-limiter when workflow runs are scheduled. We will have to see if this becomes an issue.

.github/workflows/backports.yml
```diff
@@ -0,0 +1,47 @@
...
+    if: ${{ needs.inspect-closed.outputs.target-branches != '[]' }} # Don't run if no backport/* labels were found by the prev. job
+    strategy:
+      matrix:
+        target-branch: ${{ fromJSON(needs.inspect-closed.outputs.target-branches) }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Create backport pull requests
+        uses: korthout/backport-action@v4
+        with:
+          label_pattern: a^
```
KimonHoffmann: I'm sure this is correct in the realm of the action, but a comment would be great to help confused readers such as myself.

PR #328, opened 2026-01-15
test: respect VCPKG_INSTALLED_DIR in pkg-config C consumer test

jptrainor:
The `pkg-config C consumer can link and run` test now uses `VCPKG_INSTALLED_DIR`,
allowing vcpkg installs located outside the build directory to work correctly.

The issue was exposed when specifying a custom `VCPKG_INSTALLED_DIR` during the
configure step, for example:

```sh
cmake -S .../mxl -B .../build_dir ... -DVCPKG_INSTALLED_DIR=.../install
```

When VCPKG_INSTALLED_DIR is not specified, vcpkg defaults to installing within
the build directory. The test previously hard-coded that default path, which
caused failures when a custom install directory was used.

### DIFF
--- a/lib/tests/CMakeLists.txt
+++ b/lib/tests/CMakeLists.txt
@@ -87,6 +87,7 @@ add_test(
   NAME "pkg-config C consumer can link and run"
   COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/test_pkgconfig_c_link.sh
           ${CMAKE_BINARY_DIR}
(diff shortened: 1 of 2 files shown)


jptrainor:
No need to backport this commit. I already submitted a PR to the release/1.0 branch with this change and it was accepted. That's why this backport failed.

Sorry about getting ahead of things with that release/1.0. I wasn't aware of that this automated backport process was happening.

PR #329, opened 2026-01-15
Fix audio reader resync after falling behind the writer.

k0zmo:
This fixes a resync of audio pipeline in `mxl-gst-sink` tool where previously an incorrect time point was use when doing grain index calculation.

### DIFF
--- a/tools/mxl-gst/sink.cpp
+++ b/tools/mxl-gst/sink.cpp
@@ -372,13 +372,9 @@ namespace
             // frames that correspond to the index that is readOffset
             // nanoseconds in the past and deliver them as the next frame to the
             // gstreamer pipeline.
-            auto const startTime = ::mxlGetTime();
-            auto const readDelayGrains = durationInGrains(rate, readDelay);
-            auto currentIndex = ::mxlTimestampToIndex(&rate, startTime);
-            auto requestedIndex = currentIndex - readDelayGrains;
-            auto deliveryDeadline = ::mxlIndexToTimestamp(&rate, currentIndex + 1U);
+            auto cursor = Cursor{rate, 1U, readDelay};
 
-            initializeHighestLatency("Video", rate, currentIndex, requestedIndex);
+            initializeHighestLatency("Video", rate, cursor.currentIndex(), cursor.requestedIndex());
 
             auto expectedSlices = slicesPerBatch;
             while (!g_exit_requested)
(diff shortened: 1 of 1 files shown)


KimonHoffmann:
Good catch, this should not have been `startTime`. But IMO a better fix (and the way I originally intended it) would be to use `iterationStartTime` instead of ad-hoc refetching the current time.

tools/mxl-gst/sink.cpp
```diff
@@ -695,10 +669,68 @@ namespace
...
-        }
+            Cursor(mxlRational const& rate, std::uint32_t windowSize, std::int64_t readDelay)
+                : _rate{rate}
+                , _windowSize{windowSize}
+                , _readDelayGrains{((durationInGrains(readDelay) + _windowSize - 1U) / _windowSize) * _windowSize}
+            {
+                realign(::mxlGetTime());
+            }
+
+            void next()
```
KimonHoffmann: Can be declared `noexcept`.

tools/mxl-gst/sink.cpp
```diff
@@ -695,10 +669,68 @@ namespace
...
+            {
+                ::mxlSleepUntil(_deliveryDeadline);
+
+                _currentIndex += _windowSize;
+                _requestedIndex += _windowSize;
+                _deliveryDeadline = getDeliveryDeadline();
+            }
+
+            void realign(std::uint64_t timeNow)
```
KimonHoffmann: Can also be declared `noexcept`.

tools/mxl-gst/sink.cpp
```diff
@@ -695,10 +669,68 @@ namespace
...
+            }
+
+            [[nodiscard]]
+            std::uint64_t requestedIndex() const noexcept
+            {
+                return _requestedIndex;
+            }
+
+            [[nodiscard]]
+            std::uint64_t deliveryDeadline() const noexcept
```
KimonHoffmann: All three of those can be declared `constexpr`.

tools/mxl-gst/sink.cpp
```diff
@@ -695,10 +669,68 @@ namespace
             , _highestLatencyNs{}
         {}
 
-        static std::int64_t durationInGrains(mxlRational const& rate, std::int64_t duration) noexcept
+        struct Cursor
         {
-            return ::mxlTimestampToIndex(&rate, duration);
-        }
+            Cursor(mxlRational const& rate, std::uint32_t windowSize, std::int64_t readDelay)
```
KimonHoffmann: Can be declared `noexcept`.

tools/mxl-gst/sink.cpp
```diff
@@ -695,10 +669,68 @@ namespace
...
+            }
+
+            std::uint64_t getDeliveryDeadline() const noexcept
+            {
+                return ::mxlIndexToTimestamp(&_rate, _currentIndex + _windowSize);
+            }
+
+            mxlRational const _rate;
+            std::uint32_t const _windowSize;
+            std::int64_t const _readDelayGrains;
```
KimonHoffmann: Please try to avoid `const` qualified fields. And please add a and additional `private` visibility declaration to separate the private fields from the private methods.

PR #331, opened 2026-01-16
Fix two issues with the exported cmake targets.

### LINKED ISSUE #314: Consuming MXL library using CMake file forces C++20
When you consume mxl using:
```cmake
find_package(mxl REQUIRED CONFIG)
target_link_libraries(main PRIVATE mxl::mxl)
```
The `main` target gets applied with C++20 flag. If you check the mxl-targets.cmake from the built .deb package you see this:

```cmake
add_library(mxl::mxl SHARED IMPORTED)

set_target_properties(mxl::mxl PROPERTIES
  INTERFACE_LINK_LIBRARIES "mxl::mxl-headers"
)

set_target_properties(mxl::mxl-headers PROPERTIES
  INTERFACE_COMPILE_FEATURES "cxx_std_20"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
)
set_target_properties(mxl::mxl-common PROPERTIES
  INTERFACE_COMPILE_FEATURES "cxx_std_20"
)
set_target_properties(mxl::mxl-internal-headers PROPERTIES
  INTERFACE_COMPILE_FEATURES "cxx_std_20"
  INTERFACE_LINK_LIBRARIES "mxl::mxl-headers;stduuid;spdlog::spdlog;fmt::fmt;picojson::picojson"
)
```
I think the problem is that we declare `mxl-headers` target to be:
```cmake
add_library(mxl-headers INTERFACE)
target_compile_features(mxl-headers
        INTERFACE
            cxx_std_20
    )
```
and then we use the same target for installation. Either it should be split or we use generator expression:
```cmake
add_library(mxl-headers INTERFACE)
target_compile_features(mxl-headers
        INTERFACE
            $<BUILD_INTERFACE:cxx_std_20>
    )
```

Also, what's the point of installing internal targets such as `mxl-internal-headers` or `mxl-common`. Is there any scenario where consumer of the library wants to reference them?

KimonHoffmann:
The first commit fixes #314, the second fixes an issue discovered by chance along the way.

### DIFF
--- a/lib/fabrics/CMakeLists.txt
+++ b/lib/fabrics/CMakeLists.txt
@@ -11,7 +11,7 @@ target_include_directories(mxl-fabrics-headers
             $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
     )
 
-install(TARGETS mxl-fabrics-headers EXPORT ${PROJECT_NAME}-targets
+install(TARGETS mxl-fabrics-headers EXPORT mxl-fabrics-targets
         COMPONENT ${PROJECT_NAME}-dev
     )
 


vt-tv (approved):
lgtm

Issue #10, opened 2025-06-20
Potential deadlock in DomainWatcher and Instance


pac-work:
While working on https://github.com/dmf-mxl/mxl/pull/9, I encountered the following deadlock, which does not appear to be directly caused by my changes in that PR.

Probably the simplest way of explaining this is by showing stacks:

- Thread 1:
  - futex_wait 0x00007e72c2498f70
  - mxl::lib::DomainWatcher::removeFlow DomainWatcher.cpp:147
    - tries to lock mutex of domain watcher - fails
  - mxl::lib::Instance::releaseWriter Instance.cpp:185
    - holds a mutex of instance
  - mxlReleaseFlowWriter flow.cpp:153
  - CATCH2_INTERNAL_TEST_0 test_flows_timing.cpp:89

- Thread 2:
  - futex_wait 0x00007e72c2498f70
  - mxl::lib::Instance::fileChangedCallback Instance.cpp:74
    - tries to lock mutex of instance - fails
  - $_1::operator()<…>(const uuids::uuid &, mxl::lib::WatcherType) const Instance.cpp:58
  - mxl::lib::DomainWatcher::processEvents DomainWatcher.cpp:267
    - holds the mutex of domain watcher

A bit more verbose description:

- `DomainWatcher::removeFlow` gets stuck on DomainWatcher.cpp:147 (locking `_mutex`)
  - This is entry point from my test.
  - the `removeFlow` got called by `Instance::releaseWriter`, which holds different mutex in Instance.cpp:178
- At the same time, different thread originating in `DomainWatcher::processEvents`:
  - holds the first said `_mutex` in DomainWatcher.cpp:257 while processing callbacks, thus first line of this description cannot lock it
  - the callback it is processing is `Instance::fileChangedCallback`
    - this callback tries to lock the mutex on Instance.cpp:74, which is apparantly the same mutex, which is also held by the `Instance::releaseWriter` above -> deadlock

vt-tv:
Are you able to repro this easily? I am mid flight on my way back home and could try to repro on my macbook

pac-work:
The mentioned PR https://github.com/dmf-mxl/mxl/pull/9 reproduces it with like 1:10 chance on debug build on my PC. Or I could show you sometimes next week...

vt-tv:
I managed to reproduce _something_ by running the test suite in a loop (debug build of your branch) on a macbook. I get, every now and then, this error.  Do you get the same? does it look ike the same thing?
--
 
[...]

PR #333, opened 2026-01-18
bugfix: mxl gst testsrc does not honour command line group flag

vt-tv:
minor fixes
- remove rogue object copy preventing persistance of the nmos grouphint in mxl-gst-testsrc
- unrelated: removing an inotify watch for a file that has been deleted will fail.  test if the file exists before attempting removal of the watch.

### DIFF
--- a/lib/internal/src/DomainWatcher.cpp
+++ b/lib/internal/src/DomainWatcher.cpp
@@ -7,7 +7,6 @@
 #include <filesystem>
 #include <memory>
 #include <mutex>
-#include <stdexcept>
 #include <system_error>
 #include <thread>
 #include <utility>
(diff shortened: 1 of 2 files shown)


jonasohland (approved):
LGTM

lapointejp (approved):
LGTM

lib/internal/src/DomainWatcher.cpp
```diff
@@ -225,7 +224,9 @@ namespace mxl::lib
                         MXL_ERROR("Error closing file descriptor {} for '{}'", wd, rec->fileName);
                     }
 #elif defined __linux__
-                    if (inotify_rm_watch(_inotifyFd, wd) == -1)
+                    auto f = std::filesystem::path{rec->fileName};
+                    // Try to remove the watch only if the file still exists.
+                    if (exists(f) && (inotify_rm_watch(_inotifyFd, wd) == -1))
```
pedro-alves-ferreira, line 229: This is a race condition. Although the odds are very low, can't we determine that it failed because of that, if `errno == EINVAL`?

jonasohland: I am working on a fix for #10 that should make this fix here obsolete.

KimonHoffmann: Yes these existence checks do more harm than good. It's better to just ignore the error in case of failure.

PR #334, opened 2026-01-19
Improved feature requirement workflow

peterbrightwell:
This adds a "Feature" label in the corresponding issue template, and the workflow only runs if that is present so it won't run on Bug issues etc. Also allows shortcuts for linking to mxl-requirement issue - just use the number (some Copilot used for this).

I've tested a version on my personal repo with a PAT, but for this org version we use the github app, which I can't test on my repo, so it will need testing here.

Also removes the "Feature Request" template which was causing some confusion.

### DIFF
--- a/.github/ISSUE_TEMPLATE/small_improvement.yml
+++ b/.github/ISSUE_TEMPLATE/small_improvement.yml
@@ -1,0 +1,28 @@
+name: Small Improvement
+description: Suggest an small improvement to the current feature set, that is NOT derived from an MXL Requirement
+labels:
+  - Small Improvement
+assignees: []
+title: "[Small Improvement] <short description>"
+
+body:
+  - type: textarea
+    id: description
+    attributes:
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -1,23 +0,0 @@
----
```
jonasohland, line 1: I don't think it is a good idea to remove this template. There should be a low-barrier way for people to request small improvements or convenience features from mxl that may not have a requirement (yet).

vt-tv (commented):
Branch has conflicts.

felixpou:
yes we could have an Improvement type for those cases, this would make clear what is the intent.

Task are often used for sub task in a Feature or other main type, or a maintenance or administrative task like this one. but let's document those types so it is clear for everyone.

PR #336, opened 2026-01-21
Fix data race in domain watcher

jonasohland:
Removes the callback function from the domain watcher. The domain watcher keeps a reference to the flow writer and calls the `flowRead()` function directly.

Add/remove to the domain watcher happens inside the constructor/destructor of the writer.

Work in progress because I need to fix some more of the tests.

### DIFF
--- a/lib/internal/include/mxl-internal/DomainWatcher.hpp
+++ b/lib/internal/include/mxl-internal/DomainWatcher.hpp
@@ -47,13 +47,8 @@ namespace mxl::lib
     ///
     /// Monitors flows on disk for changes.
     ///
-    /// A flow reader is looking for changes to the {mxl_domain}/{flow_id}.mxl-flow/data file
-    /// An update to this flow triggers a callback that will notify a condition variable in the relevant FlowReaders.
-    /// if a reader is waiting for the next grain it will be notified that the grain is ready.
-    ///
     /// A flow writer is looking for changes to the {mxl_domain}/{flow_id}.mxl-flow/access file. This file is 'touched'
-    /// by readers when they read a grain, which will trigger a 'FlowInfo.lastRead` update (performed by the FlowWriter
-    /// since the writer is the only one mapping the flow in write-mode).
+    /// by readers when they read a grain, which will trigger a 'FlowInfo.lastRead` update.
     ///
     class MXL_EXPORT DomainWatcher
     {


lib/internal/include/mxl-internal/DomainWatcher.hpp
```diff
@@ -35,15 +30,13 @@ namespace mxl::lib
...
-        WatcherType type;
-        /// use count (many readers and writers can point to the same watch/flow
-        uint16_t count;
 
-        /// records are considered equals if their IDs and Types are equals.
-        bool operator==(DomainWatcherRecord const& other) const
+        DiscreteFlowWriter* fw;
+
+        [[nodiscard]]
+        bool operator==(DomainWatcherRecord const& other) const noexcept
```
KimonHoffmann, line 41: Can be declared `constexpr`.

lib/internal/include/mxl-internal/DomainWatcher.hpp
```diff
@@ -35,15 +30,13 @@ namespace mxl::lib
...
         {
-            return id == other.id && type == other.type;
+            return id == other.id && fw == other.fw;
```
KimonHoffmann: Could use parentheses around the binary sub-expressions.

lib/internal/include/mxl-internal/DomainWatcher.hpp
```diff
@@ -99,14 +79,18 @@ namespace mxl::lib
             _running = false;
         }
 
+        std::size_t count(uuids::uuid) noexcept;
```
KimonHoffmann: * Should be annotated with `[[nodiscard]]`
* The parameters should be named in the header even if they're ultimately unused in the implementation.
* Should be declared `const` with respect to the comment below.

lib/internal/include/mxl-internal/DomainWatcher.hpp
```diff
@@ -99,14 +79,18 @@ namespace mxl::lib
...
+        std::size_t size() noexcept;
```
KimonHoffmann: * Should be annotated with `[[nodiscard]]`
* Should be declared `const` with respect to the comment below.

lib/internal/include/mxl-internal/DomainWatcher.hpp
```diff
@@ -118,7 +102,7 @@ namespace mxl::lib
 #endif
 
         /// Map of watch descriptors to file records.  Multiple records could use the same watchfd
-        std::unordered_multimap<int, DomainWatcherRecord::ptr> _watches;
+        std::unordered_multimap<int, DomainWatcherRecord> _watches;
         /// Prodect maps
         std::mutex _mutex;
```
KimonHoffmann: Should be declared `mutable` in order to not force conceptually `const` methods to be non-`const`.

lib/internal/include/mxl-internal/Instance.hpp
```diff
@@ -37,7 +37,8 @@ namespace mxl::lib
         /// \param[in] options Additional options. \todo Not implemented yet.
         /// \param[in] flowIoFactory A factory used to create flow readers for flows of different types.
         ///
-        Instance(std::filesystem::path const& mxlDomain, std::string const& options, std::unique_ptr<FlowIoFactory>&& flowIoFactory);
+        Instance(std::filesystem::path const& mxlDomain, std::string const& options, std::unique_ptr<FlowIoFactory>&& flowIoFactory,
+            DomainWatcher::ptr);
```
KimonHoffmann: Argument in the header should be named and ideally documented in the comment above.

lib/internal/include/mxl-internal/PosixFlowIoFactory.hpp
```diff
@@ -3,12 +3,15 @@
...
 #pragma once
 
+#include "mxl-internal/DomainWatcher.hpp"
 #include "mxl-internal/FlowIoFactory.hpp"
 
 namespace mxl::lib
 {
     struct MXL_EXPORT PosixFlowIoFactory : FlowIoFactory
     {
+        PosixFlowIoFactory(DomainWatcher::ptr);
```
KimonHoffmann: The parameter should be named in the declaration.

lib/internal/src/DomainWatcher.cpp
```diff
@@ -140,128 +141,111 @@ namespace mxl::lib
...
-        auto id = uuids::to_string(in_flowId);
-        MXL_DEBUG("Add Flow : {}", id);
-
-        // Create a record for this watched file
-        auto record = std::make_shared<DomainWatcherRecord>();
-        record->count = 0;
-        record->id = in_flowId;
-        record->type = in_type;
+        auto lock = std::lock_guard{_mutex};
+        auto it = std::ranges::find_if(_watches, [id](std::pair<int const&, DomainWatcherRecord const&> it) { return it.second.id == id; });
```
KimonHoffmann: The lambda should probably take its argument by (const-qualfied) reference... just `auto const&` if you want to be brief. Also I think `it` as the argument name is misleading here, because it's not an iterator but an element reference.

lib/internal/src/DomainWatcher.cpp
```diff
@@ -140,128 +141,111 @@ namespace mxl::lib
...
+    void DomainWatcher::addFlow(DiscreteFlowWriter* writer, uuids::uuid id)
+    {
+        auto record = DomainWatcherRecord{
+            .id = id,
+            .fileName = makeFlowAccessFilePath(makeFlowDirectoryName(_domain, uuids::to_string(id))),
+            .fw = writer,
+        };
         {
-            if (*rec == *record)
+            std::lock_guard<std::mutex> lock(_mutex);
```
KimonHoffmann: Should use the same syntax as the others.

```cpp
auto lock = std::lock_guard{_mutex};
```

lib/internal/src/DomainWatcher.cpp
```diff
@@ -140,128 +141,111 @@ namespace mxl::lib
...
+            int existingWd = -1;
```
KimonHoffmann: Could use `auto`.

lib/internal/src/DomainWatcher.cpp
```diff
@@ -140,128 +141,111 @@ namespace mxl::lib
...
-        {
-            MXL_DEBUG("Record for {} not found, creating one.", id);
-            record->count = 1;
+            if (existingWd == -1)
+            {
+                MXL_DEBUG("Record for {} not found, creating one.", uuids::to_string(record.id));
 
 #ifdef __APPLE__
-            int wd = open(record->fileName.c_str(), O_EVTONLY);
+                int wd = open(record->fileName.c_str(), O_EVTONLY);
```
KimonHoffmann: Could use `auto const`. What happens to this FD going forward, should this rather be `existingWd`?

Also: Could use a full qualification via `::`.

lib/internal/src/DomainWatcher.cpp
```diff
@@ -140,128 +141,111 @@ namespace mxl::lib
...
 #elif defined __linux__
-            // if not found, add the watch and add it to the maps
-            int wd = inotify_add_watch(_inotifyFd, record->fileName.c_str(), ((in_type == WatcherType::READER) ? IN_MODIFY : IN_ACCESS) | IN_ATTRIB);
+                // if not found, add the watch and add it to the maps
+                existingWd = inotify_add_watch(_inotifyFd, record.fileName.c_str(), IN_ACCESS | IN_ATTRIB);
```
KimonHoffmann: Could probably use a `::` qualification.

lib/internal/src/DomainWatcher.cpp
```diff
@@ -140,128 +141,111 @@ namespace mxl::lib
...
-                throw std::system_error(error, std::generic_category(), "Failed to add watch for file: " + record->fileName);
-            }
-            else
-            {
-                MXL_DEBUG("Added watch {} for file: {}", wd, record->fileName);
+                if (existingWd == -1)
+                {
+                    auto const error = errno;
+                    MXL_ERROR("Failed to add watch for file '{}': {}", record.fileName, std::strerror(error));
+                    throw std::system_error(error, std::generic_category(), "Failed to add watch for file: " + record.fileName);
```
KimonHoffmann: Can probably use curly braces?

lib/internal/src/DomainWatcher.cpp
```diff
@@ -140,128 +141,111 @@ namespace mxl::lib
...
 
-        // Create a record for this watched file
-        auto record = std::make_shared<DomainWatcherRecord>();
-        record->count = 0;
-        record->id = in_flowId;
-        record->type = in_type;
-        record->fileName = _domain / id;
+            // Remove the record for this writer
+            auto it = std::ranges::find_if(_watches,
+                [record](std::pair<int const&, DomainWatcherRecord const&> item) { return item.second == record; });
```
KimonHoffmann: As above.

lib/internal/src/DomainWatcher.cpp
```diff
@@ -140,128 +141,111 @@ namespace mxl::lib
...
+    void DomainWatcher::removeFlow(DiscreteFlowWriter* writer, uuids::uuid id)
     {
-        auto id = uuids::to_string(in_flowId);
+        auto record = DomainWatcherRecord{
+            .id = id,
+            .fileName = makeFlowAccessFilePath(makeFlowDirectoryName(_domain, uuids::to_string(id))),
+            .fw = writer,
+        };
+        {
+            auto lock = std::lock_guard<std::mutex>{_mutex};
```
KimonHoffmann: Can drop the mention of `std::mutex` in favor of CTAD.

lib/internal/src/DomainWatcher.cpp
```diff
@@ -337,7 +321,7 @@ namespace mxl::lib
             }
 
             // We have an inotify event ready
-            ssize_t length = read(_inotifyFd, buffer, sizeof(buffer));
+            ::ssize_t length = ::read(_inotifyFd, buffer, sizeof(buffer));
```
KimonHoffmann: how about `auto` as the type? Also can just use `sizeof buffer`.

lib/internal/src/DomainWatcher.cpp
```diff
@@ -353,19 +337,19 @@ namespace mxl::lib
                 continue; // nothing to read (should not happen if nfds > 0)
             }
 
-            std::lock_guard<std::mutex> lock(_mutex);
+            auto lock = std::lock_guard<std::mutex>{_mutex};
```
KimonHoffmann: As above regarding CTAD.

lib/internal/src/DomainWatcher.cpp
```diff
@@ -353,19 +337,19 @@ namespace mxl::lib
...
             for (char* ptr = buffer; ptr < buffer + length;)
             {
-                struct inotify_event* event = reinterpret_cast<struct inotify_event*>(ptr);
+                auto const* event = reinterpret_cast<struct ::inotify_event const*>(ptr);
```
KimonHoffmann: could use just `auto`.

lib/internal/src/DomainWatcher.cpp
```diff
@@ -353,19 +337,19 @@ namespace mxl::lib
                 continue; // nothing to read (should not happen if nfds > 0)
             }
 
-            std::lock_guard<std::mutex> lock(_mutex);
+            auto lock = std::lock_guard<std::mutex>{_mutex};
             for (char* ptr = buffer; ptr < buffer + length;)
```
KimonHoffmann: Could use `auto ptr = buffer;`.

lib/internal/src/PosixFlowIoFactory.cpp
```diff
@@ -9,6 +9,10 @@
 
 namespace mxl::lib
 {
+    PosixFlowIoFactory::PosixFlowIoFactory(DomainWatcher::ptr watcher)
+        : _watcher(std::move(watcher))
```
KimonHoffmann: Should use curly braces.

lib/internal/src/PosixDiscreteFlowWriter.hpp
```diff
@@ -73,5 +77,10 @@ namespace mxl::lib
         std::unique_ptr<DiscreteFlowData> _flowData;
         /** The currently opened grain index. MXL_UNDEFINED_INDEX if no grain is currently opened. */
         std::uint64_t _currentIndex;
+
+        // The watcher reference needs live in the most derived class of `FlowWriter` because it only synchronizes with the `DomainWatcher` thread
+        // while the destructor runs. If it was inside the `FlowWriter` destructor itself, the domain watcher thread could call `flowRead()` of a
+        // already destructed instance of `PosixDiscreteFlowWriter` which would result in a `pure virtual function called` exception.
+        DomainWatcher::ptr _watcher;
```
KimonHoffmann, line 81: This could also be achieved by having the reference in the base class, and provide a protected `releaseWatcher()` method that the derived destructors call to drop the reference early.

But I don't think there is any real benefit over doing it this way TBH.

lib/internal/tests/test_domainwatcher.cpp
```diff
@@ -6,198 +6,287 @@
...
 namespace fs = std::filesystem;
 
+namespace
+{}
+
+struct MockFlowFiles
+{
+    MockFlowFiles(fs::path domain, uuids::uuid const& id)
+        : _domain(std::move(domain))
+        , _id(id)
```
KimonHoffmann: Can both use curly braces.

lib/internal/tests/test_domainwatcher.cpp
```diff
@@ -6,198 +6,287 @@
...
+        {
+            auto const error = errno;
+            throw std::system_error(error, std::system_category(), "Failed to update flow access time");
+        }
+    }
+
+    fs::path _domain;
+    uuids::uuid _id;
+
+    int _accessFileFd = -1;
```
KimonHoffmann: Should be moved to the constructor to give a concise overview of all initial values in one place.

lib/internal/tests/test_domainwatcher.cpp
```diff
@@ -6,198 +6,287 @@
...
+};
+
+struct MockWriter : mxl::lib::DiscreteFlowWriter
+
+{
+    MockWriter(uuids::uuid id, DomainWatcher::ptr const& watcher)
+        : mxl::lib::DiscreteFlowWriter(id)
+        , _id(id)
+        , _watcher(watcher)
```
KimonHoffmann: Could all use curly braces.

lib/internal/tests/test_domainwatcher.cpp
```diff
@@ -6,198 +6,287 @@
...
+    }
+
+    ~MockWriter()
+    {
+        _watcher->removeFlow(this, _id);
+    }
+
+    FlowData const& getFlowData() const override
+    {
+        throw;
```
KimonHoffmann: Is this intentional? Outside of an active exception handler this is equivalent to `std::terminate()`.

lib/internal/tests/test_domainwatcher.cpp
```diff
@@ -6,198 +6,287 @@
...
+    {
+        _watcher->addFlow(this, _id);
+    }
+
+    ~MockWriter()
+    {
+        _watcher->removeFlow(this, _id);
+    }
+
+    FlowData const& getFlowData() const override
```
KimonHoffmann: Should be declared `virtual` in addition to `override`.

lib/internal/tests/test_domainwatcher.cpp
```diff
@@ -6,198 +6,287 @@
...
+    {
+        throw;
+    }
+
+    ::mxlFlowInfo getFlowInfo() const override
```
KimonHoffmann: Should be declared `virtual` in addition to `override`.

lib/internal/tests/test_domainwatcher.cpp
```diff
@@ -6,198 +6,287 @@
...
+    {
+        return info;
+    }
+
+    ::mxlFlowConfigInfo getFlowConfigInfo() const override
```
KimonHoffmann: Should be declared `virtual` in addition to `override`.

lib/internal/tests/test_domainwatcher.cpp
```diff
@@ -6,198 +6,287 @@
...
+    {
+        return info.config;
+    }
+
+    ::mxlFlowRuntimeInfo getFlowRuntimeInfo() const override
```
KimonHoffmann: Should be declared `virtual` in addition to `override`.

lib/internal/tests/test_domainwatcher.cpp
```diff
@@ -6,198 +6,287 @@
...
+    {
+        return info.runtime;
+    }
+
+    bool isExclusive() const override
```
KimonHoffmann: Should be declared `virtual` in addition to `override`.

lib/internal/tests/test_domainwatcher.cpp
```diff
@@ -6,198 +6,287 @@
...
+    {
+        return false;
+    }
+
+    bool makeExclusive() override
```
KimonHoffmann: Should be declared `virtual` in addition to `override`.

lib/internal/tests/test_domainwatcher.cpp
```diff
@@ -6,198 +6,287 @@
...
+    {
+        return false;
+    }
+
+    /**
+     * Get the grain info for a specific grain index without opening the grain for mutation.
+     */
+    [[nodiscard]]
+    mxlGrainInfo getGrainInfo(std::uint64_t) const override
```
KimonHoffmann: Should be declared `virtual` in addition to `override`.

lib/internal/tests/test_domainwatcher.cpp
```diff
@@ -6,198 +6,287 @@
...
+    {
+        throw;
```
KimonHoffmann: As above regarding the effects outside of exception handlers.

lib/internal/tests/test_domainwatcher.cpp
```diff
@@ -6,198 +6,287 @@
...
+    };
+
+    mxlStatus openGrain(std::uint64_t, mxlGrainInfo*, std::uint8_t**) override
+    {
+        return MXL_STATUS_OK;
+    }
+
+    mxlStatus commit(mxlGrainInfo const&) override
```
KimonHoffmann: Should be declared `virtual` in addition to `override`.

lib/internal/tests/test_domainwatcher.cpp
```diff
@@ -6,198 +6,287 @@
...
+    /**
+     * Get the grain info for a specific grain index without opening the grain for mutation.
+     */
+    [[nodiscard]]
+    mxlGrainInfo getGrainInfo(std::uint64_t) const override
+    {
+        throw;
+    };
+
+    mxlStatus openGrain(std::uint64_t, mxlGrainInfo*, std::uint8_t**) override
```
KimonHoffmann: Should be declared `virtual` in addition to `override`.

lib/internal/tests/test_domainwatcher.cpp
```diff
@@ -6,198 +6,287 @@
...
+    {
+        return MXL_STATUS_OK;
+    }
+
+    mxlStatus commit(mxlGrainInfo const&) override
+    {
+        return MXL_STATUS_OK;
+    }
+
+    mxlStatus cancel() override
```
KimonHoffmann: Should be declared `virtual` in addition to `override`.

lib/internal/tests/test_domainwatcher.cpp
```diff
@@ -6,198 +6,287 @@
...
+    {
+        return MXL_STATUS_OK;
+    }
+
+    void flowRead() override
```
KimonHoffmann: Should be declared `virtual` in addition to `override`.

lib/internal/tests/test_domainwatcher.cpp
```diff
@@ -6,198 +6,287 @@
...
+    {
+        _notNotified.clear();
+    }
+
+    bool checkAndClearNotified()
```
KimonHoffmann: Should be annotated with `[[nodiscard]]`.

lib/internal/tests/test_domainwatcher.cpp
```diff
@@ -6,198 +6,287 @@
...
+    {
+        return !_notNotified.test_and_set();
+    }
+
+    mxlFlowInfo info;
+    uuids::uuid _id;
+    std::atomic_flag _notNotified{true};
```
KimonHoffmann: Should be moved to the constructors list of initializers to have all initial values in one place.

lib/internal/tests/test_domainwatcher.cpp
```diff
@@ -240,208 +328,206 @@ TEST_CASE_PERSISTENT_FIXTURE(mxl::tests::mxlDomainFixture, "DomainWatcher suppor
...
+        std::optional<std::thread> thread;
+    };
+
+    auto watcher = std::make_shared<DomainWatcher>(domain);
+    auto work = [&](Worker* w)
+    {
+        readyCount.fetch_add(1);
+        while (!startSignal.load())
+        {
+        }
```
KimonHoffmann: Can/should probably use `wait(false)` instead.

lib/internal/tests/test_domainwatcher.cpp
```diff
@@ -240,208 +328,206 @@ TEST_CASE_PERSISTENT_FIXTURE(mxl::tests::mxlDomainFixture, "DomainWatcher suppor
...
+//         auto touch = std::ofstream{bad2};
+//         touch << "notadir";
+//     }
+//     REQUIRE(exists(bad2));
+//     REQUIRE(is_regular_file(bad2));
+//     REQUIRE_THROWS_AS((mxl::lib::DomainWatcher{bad2, nullptr}), std::filesystem::filesystem_error);
+//
+//     // Clean up
+//     remove_all(bad2);
+// }
```
KimonHoffmann: This entire section can probably just go, can't it?

KimonHoffmann (commented):
Thank you for this fix! I only have a few minor remarks, most of them syntactic/stylistic.

lib/internal/src/DomainWatcher.cpp
jonasohland: I have no way to compile and test the code inside `#ifdef __APPLE__`.
I was planning on building myself a hackintosh on proxmox for a different project, but I don't know when I will actually do that.

KimonHoffmann: Judging from how it was before it seems to me as if using `existingWd` is the right choice here. I can test it later on macOS.

lib/internal/src/PosixDiscreteFlowWriter.hpp
jonasohland: @vt-tv @lapointejp 
I think one of you is the orig. author of this code. Any opinions?

lib/internal/tests/test_domainwatcher.cpp
jonasohland: The implementation for this function is just not provided by the Mock and this is a way to silence the warning about missing return.
Is there a better way to do that?

lib/internal/tests/test_domainwatcher.cpp
jonasohland: See above

lib/internal/tests/test_domainwatcher.cpp
KimonHoffmann: Maybe calling `std::terminate()` is a little bit more explicit about the intent. It is annotated with `[[noreturn]]` so it *should* also silence any warnings about not returning anything.

lib/internal/tests/test_domainwatcher.cpp
KimonHoffmann: In this specific case it could just do `return {};`

lib/internal/tests/test_domainwatcher.cpp
jonasohland: Not sure I understand what you mean. Where does `wait()` come from?

KimonHoffmann: `std::atomic<>` has a `wait()` method that basically does what you're doing here manually.
See: https://en.cppreference.com/w/cpp/atomic/atomic/wait

jonasohland: Wow
I did not know that!

jonasohland:
> > My thread sanitizer is still reporting an issue with the access to the shared memory data during the call to `flowRead()` from the DomainWatcher thread.
> 
> Could you quote the complete diagnostic here for reference?

[mxl-test-helgrind.txt](https://github.com/user-attachments/files/24795695/mxl-test-helgrind.txt)

If I comment out the implementation of `PosixDiscreteFlowWriter::flowRead()` function the diagnostic no longer appears.

jonasohland:
I believe the offending access is to the vtable of `PosixDiscreteFlowWriter`.

I was wrong before about commenting out the implementation of `flowRead()`. This does not actually fix the issue, it just makes the test abort before the race can happen. 
What actually fixes it: Commenting out the **call** to `flowRead()` inside the `processEvents()` function. This is where a vtable access happens that competes with the vtable access that happens during the destructor call of `PosixDiscreteFlowWriter`.

lib/internal/include/mxl-internal/DomainWatcher.hpp
```diff
@@ -61,35 +58,22 @@ namespace mxl::lib
     class MXL_EXPORT DomainWatcher
```
mlefebvre1, line 53: Will need to update the class documentation.

lib/internal/include/mxl-internal/DomainWatcher.hpp
```diff
@@ -61,35 +58,22 @@ namespace mxl::lib
...
-        ///
-        int16_t addFlow(uuids::uuid const& in_flowId, WatcherType in_type);
-
-        ///
-        /// Removes a flow from being watched.
-        /// \param in_flowId the flow to remove
-        /// \return The current use count of this watched flow + type.  -1 if the watch did not exist.
-        ///
-        int16_t removeFlow(uuids::uuid const& in_flowId, WatcherType in_type);
+        void addFlow(DiscreteFlowWriter* writer, uuids::uuid id);
```
mlefebvre1, line 75: Missing function documentation.

lib/internal/src/DomainWatcher.cpp
```diff
@@ -140,128 +142,117 @@ namespace mxl::lib
...
+                if (::inotify_rm_watch(_inotifyFd, wd) == -1)
                 {
-                    rec->count--;
-                    useCount = rec->count;
+                    auto const error = errno;
+                    if (error != EINVAL || std::filesystem::exists(it->second.fileName))
+                    {
+                        MXL_WARN("Failed to remove inotify watch (wd={}) for '{}': {}", wd, record.fileName, std::strerror(error));
+                    }
+                    // Continue with cleanup despite failure
```
mlefebvre1: I think this comment could be removed, there's nothing else to be done after this point.

lib/internal/src/DomainWatcher.cpp
```diff
@@ -140,128 +142,117 @@ namespace mxl::lib
...
+                    MXL_ERROR("Error closing file descriptor {} for '{}'", wd, rec->fileName);
                 }
-                else
+#elif defined __linux__
+                if (::inotify_rm_watch(_inotifyFd, wd) == -1)
                 {
-                    rec->count--;
-                    useCount = rec->count;
+                    auto const error = errno;
+                    if (error != EINVAL || std::filesystem::exists(it->second.fileName))
```
KimonHoffmann: Please add parentheses around the binary sub-expression.

PR #341, opened 2026-01-29
Basic fix for the macos build

jonasohland:
Does not provide the same guarantees as the linux code, but fixes the build.


jonasohland:
#336 Will break the macos build again, if we merge this first, I will rebase #336 and fix the mac issues there too.

jonasohland:
Fixed the new issues introduced with the DomainWatcher changes. Ready for review again!

jonasohland:
> lgtm. minor comment about auto+curly braces init in DomainWatcher.cpp:338

fixed!

PR #350, opened 2026-01-29
Make the backport workflow use a token generated from github app credentials

jonasohland:
I have no way to test this...


vt-tv (approved):
worth a test.

PR #382, opened 2026-02-05
Silence warning in picojson header for GCC release build

jonasohland:
Not the prettiest fix, but it silences the warning only for that header.


vt-tv (approved):
Lgtm

PR #394, opened 2026-02-12
Add build utils option

alanpawlak:
While working on the vcpkg overlay port for mxl, I noticed that we currently enforce building utils. This isn't necessary for users who only need the library for their projects, especially since building the utilities introduces a dependency on gstreamer.

Similarly we should check for BUILD_TESTS in lib/internal/CMakeLists.txt so we do not require Catch2 when someone is just building the lib.

We should also add core dependencies to package configuration file (https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#creating-a-package-configuration-file).

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -117,6 +117,11 @@ write_basic_package_version_file(
         COMPATIBILITY SameMinorVersion
     )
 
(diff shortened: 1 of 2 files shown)


KimonHoffmann (approved):
Looks good to me, thank you for this!

PR #395, opened 2026-02-12
mxl-sink: Fix video bug causing desync and artifacts with certain files

pedro-alves-ferreira:


rust/gst-mxl-rs/src/mxlsink/imp.rs
```diff
@@ -413,13 +412,17 @@ fn init_mxl_instance(
...
-        mxl::MxlInstance::new(mxl_api, settings.domain.as_str(), "").map_err(|e| {
-            gst::error_msg!(
-                gst::CoreError::Failed,
-                ["Failed to load MXL instance: {}", e]
-            )
-        })?;
+    let mxl_instance = mxl::MxlInstance::new(
+        mxl_api,
+        settings.domain.as_str(),
+        "", /*r#"{"urn:x-mxl:option:history_duration/v1.0": 900000000}"#*/
```
jonasohland: I think this diff is only because of this comment? It can probably be removed. (I have learned now that options are unused 🥲).

rust/gst-mxl-rs/src/mxlsink/render_video.rs
```diff
@@ -1,65 +1,41 @@
...
-            );
-            if index > current_index && index - current_index > video_state.grain_count as u64 {
-                index = current_index + video_state.grain_count as u64 - 1;
-            }
-            video_state.grain_index = index;
-        }
-        None => {
-            video_state.grain_index = current_index;
-        }
+    let buffer_size = video_state.grain_count as u64;
```
jonasohland, line 23: I guess that means the PTS of the incoming buffer is no longer used and we just 'syntonize' to the current grain epoch?

rust/gst-mxl-rs/src/mxlsink/imp.rs
SimaoFonseca2: Comment removed.

rust/gst-mxl-rs/src/mxlsink/render_video.rs
SimaoFonseca2: Yes, that’s correct. We now use the grain epoch as the timing reference instead of mapping from buffer PTS. Audio was already using grain epoch instead of the PTS so this change keeps it consistent.

PR #407, opened 2026-02-15
Fix uncaught exception in mxl-info

jonasohland:
Thrown by constructor of ScopedMxlInstance

### DIFF
--- a/tools/mxl-info/main.cpp
+++ b/tools/mxl-info/main.cpp
@@ -346,6 +346,20 @@ namespace
             return EXIT_FAILURE;
         }
     }
+
+    template<typename F>
+    int tryRun(F&& f)
+    {
+        try
+        {
+            return f();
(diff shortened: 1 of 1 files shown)


tools/mxl-info/main.cpp
```diff
@@ -409,22 +409,35 @@ int main(int argc, char** argv)
         return EXIT_FAILURE;
     }
 
+    auto tryRun = [](auto&& f)
```
pedro-alves-ferreira: How about moving this to the anonymous namespace, to avoid cluttering main?

jonasohland: I like having behavior defined locally, but `tryRun` should be clear enough as a name.

jonasohland:
Checks failing because of vcpkg fails to download the cmake binary.

Issue #420, opened 2026-02-24
Building.md documentation error in sequence


sviengkhou:
**Describe the bug**
In Building.md file the sequence example described in the Option 2: CMake with presets cannot be reproduced in console with build/Linux-Clang-Debug path.
It reads
```
# Generate the build files using the Linux-Clang-Debug preset
mkdir build
cd build
cmake .. --preset Linux-Clang-Debug

# Build everything
cmake --build build/Linux-Clang-Debug --target all
```

but it should be (since the example already brings the user to build directory)
```
# Generate the build files using the Linux-Clang-Debug preset
mkdir build
cd build
cmake .. --preset Linux-Clang-Debug

# Build everything
cmake --build Linux-Clang-Debug --target all
```

jonasohland:
fixed in #423

PR #423, opened 2026-02-25
fix small documentation error

sviengkhou:
as described in this bug:  https://github.com/dmf-mxl/mxl/issues/420


sviengkhou:
@vt-tv I would backport since it is documentation related and we want the best documentation possible.

PR #443, opened 2026-03-26
Fix #438 by accounting for the fact that only half of the buffer of a continuous flow is readable at any one point in time

### LINKED ISSUE #438: History duration issue with continuous ring buffers
The _accessible_ history duration of audio samples  is always _exactly_ half the actual history duration.

The attached patch adds a unit test demonstrating the issue : 

[test_flows_audio_history_half.patch](https://github.com/user-attachments/files/26090188/test_flows_audio_history_half.patch)

### LINKED ISSUE #571: [Task] Backport to v1.0 #438 on half of the buffer of a continuous flow
### Description

Apparently automatic backport of this PR did not hapenend properly.

Let's do this manually. 

Backport #443 to v1.0

KimonHoffmann:
The configured history duration is used to determine the buffer size, but before this change this failed to account for the fact that only half of the buffer is readable at any one point in time, effectively halving the history duration for all continuous flows.


lib/internal/src/Instance.cpp
```diff
@@ -260,7 +260,7 @@ namespace mxl::lib
         // Read the mandatory grain_rate field
         auto const sampleRate = parser.getGrainRate();
         // Compute the grain count based on our configured history duration
-        auto const bufferLength = _historyDuration * sampleRate.numerator / (1'000'000'000ULL * sampleRate.denominator);
+        auto const bufferLength = _historyDuration * sampleRate.numerator / (500'000'000ULL * sampleRate.denominator);
```
pedro-alves-ferreira, line 266: I would add a comment to make it clearer. Something like: "The denominator is multiplied by 500M, rather than 1B, so that the buffer length is twice as large as the history duration."

pedro-alves-ferreira (approved):
Thanks for fixing this. LGTM, with one comment.

lib/internal/src/Instance.cpp
KimonHoffmann: Fair point, I added an explanatory comment.

vt-tv:
@KimonHoffmann  could you add the unit test that was attached to the bug report? that would validate your fix

KimonHoffmann:
> @KimonHoffmann could you add the unit test that was attached to the bug report? that would validate your fix

Unfortunately not, because it lacks a DCO signoff and does not follow the practices of our unit tests (it has intermittent console output, for example). I can try to attach a clean room test instead.

SpungeCake:
Hi, Serge from GV here, I'm the one who reported the problem to Vincent Trussart.  This is not an acceptable fix, I tried it with the unit test I wrote that reproduces the issue, still not fixed.  Doubling the history will make the size match the video and anc duration but for an audio only configuration the returned size is invalid since only half of it is usable.  That makes the test fail.

KimonHoffmann:
> Hi, Serge from GV here, I'm the one who reported the problem to Vincent Trussart. This is not an acceptable fix, I tried it with the unit test I wrote that reproduces the issue, still not fixed. Doubling the history will make the size match the video and anc duration but for an audio only configuration the returned size is invalid since only half of it is usable. That makes the test fail.

You mean that only half of `mxlContinuousFlowConfigInfo::bufferLength` is actually accessible for reading?
While I agree that this does not match the way in which `mxlDiscreteFlowConfigInfo::grainCount` can be used, we claim nowhere that all of `bufferLength` is accessible for reading. On the contrary the documentation in `architecture.md` explicitly states otherwise.

We can consider how to improve the UX on this, I'm not really in favor of having the physical buffer length diverge from `mxlContinuousFlowConfigInfo::bufferLength`, but maybe an easy to use inline function to obtain the usable min-index for a given flow could be a way to help users?

SpungeCake:
My point is mainly that a flow writer and reader user need to know exactly how much data a flow can hold and when it is outside the range.  This fixes the continuous buffer duration not matching the discrete flow, I'm ok with this.  However we still need to provide a way for a continuous user to know the exact history duration.  I just find it awkward that I have to divide by 2 the bufferLength given by the mxlContinuousFlowConfigInfo, we need a cleaner way, maybe inline function will do the work.  The fact that mxlFlowConfigInfo is a union of the two types of config makes you think it returns a similar information about the buffer duration, I'm pretty sure more will do the same mistake and not divide the bufferLength by two.  At least flowinfo.h header should be updated to explain this for mxlContinuousFlowConfigInfo.

KimonHoffmann:
Yes, I agree that the UX as it is right now is error prone.
I think an inline function and a comment in `mxlContinuousFlowConfigInfo` referring to it should do the trick.

vt-tv (approved):
LGTM

KimonHoffmann:
I still wanted to provide the inline function ...

felixpou:
I am not sure this was backported to v1.0, I don't see this was released in v1.0.1 ?

Issue #364, opened 2026-02-02
[Small Improvement]  Create a GitHub Pages for clean documentation


felixpou:
### Description

We could use mxl.github.io to better format the documentation, tracking the versions. https://docs.github.com/en/pages

vt-tv:
- TSC 03-19 : no progress.

peterbrightwell:
I just tried out Zensical, looks promising and was fairly straightforward.  See https://peterbrightwell.github.io/mxl-docs/, which has the main MXL readme as its index page (we'd probably want to change that) and the existing docs/ as other pages.

peterbrightwell:
2026-04-09: updated version shown on TSC call. We should do this within the mxl repo, not a separate docs repo. TODO: work out best approach across released versions.

peterbrightwell:
See https://peterbrightwell.github.io/mxl for a proposed layout using Zensical in the mxl repo

- For now each version has its own directory under `docs` on `main`, rather than using the NMOS-spec approach of checking out multiple branches/tags, then rendeing and combining these (this is significantly more complicated, so we should decide if it's really worth doing it)
- Related docs are grouped into directories within each version (copy-paste as needed for now). This lets us split larger docs such as `Architecture.md` into smaller docs.
- The Zensical config file describes the navigation (as well project settings).  Currently manual.

peterbrightwell:
See https://peterbrightwell.github.io/mxl with a version selector, and https://github.com/dmf-mxl/mxl/pull/455#issuecomment-4297695714 comments on that.

felixpou:
TSC
- First step is to render the current /docs
- With versioning based on the current release list in the CMake
- [ ] Peter will look at this the week after.

felixpou:
TSC 
- @peterbrightwell is progressing

peterbrightwell:
Ok, I think this is now worth checking following #541 and #542:

Go to <https://dmf-mxl.github.io/mxl> (which redirects to "latest") and click on things.

To see difference between v1.0.0 and v1.0.1 go to the Building page and look at the CMake presets

felixpou:
TSC

- Done

PR #455, opened 2026-04-13
Render documentation for GitHub Pages

peterbrightwell:
Addresses #364 

The `.github/workflow/docs.yml` action renders a site with [Zensical](https://github.com/zensical/zensical), checks for broken links with [Lychee](https://github.com/lycheeverse/lychee) and uploads to GitHub pages.

It needs to be run from `main` to use GitHub Pages, which will need enabling in settings once the PR is merged (default name will be dmf-mxl.github.io/mxl). So to look at it before that see https://peterbrightwell.github.io/mxl  -- it's improved a little from what I showed at the TSC, and now adds `v1.0` or `v1.1` to the sidebar selector (only placeholder v1.1 content for now).

For now the navigation is manually defined in the `nav` object of `zensical.toml` (which annoyingly can't be in the docs/ folder -- see https://github.com/zensical/backlog/issues/101) but that can probably be improved later.

For now I have removed the MXL logo from the header as it needs to be in the documentation tree rather than use from mxl-logos, but we can fix this at some point.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,3 +1,6 @@
+<!-- SPDX-FileCopyrightText: 2026 Contributors to the Media eXchange Layer project. -->
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+
 <!-- Before submitting a pull request, please check that:
 - it has a suitably descriptive title
 - it mentions any issue(s) that it addresses


docs/Guide/Building.md
```diff
@@ -23,8 +23,8 @@ This is the preferred option for development on WSL2 or native linux desktop. Th
 
 _Note: the following instructions apply to Ubuntu 22.04._
 
-1. Install all apt packages specified in the [Dockerfile](../.devcontainer/Dockerfile)
-2. Install vcpkg as done in the [Dockerfile](../.devcontainer/Dockerfile)
+1. Install all apt packages specified in the [Dockerfile](https://github.com/dmf-mxl/mxl/blob/main/.devcontainer/Dockerfile)
+2. Install vcpkg as done in the [Dockerfile](https://github.com/dmf-mxl/mxl/blob/main/.devcontainer/Dockerfile)
```
vt-tv: Does this mean that we will have to edit the URLs for each branches?  (/mxl/blob/main) Can't we use relative URLs so we don't have to change them all the time?

peterbrightwell:
Following discussion on TSC it would be good to get a version selector included. See https://peterbrightwell.github.io/mxl/ for an example of this, allowing you to select between v0.1.0 (stable) and v0.1.1 (latest).  (The two trees are the same apart from the comment about versions).   This example uses [mike](https://github.com/jimporter/mike) with zensical, pending a zensical-native versioning system.  It seems to work ok, but unlike the approach in this draft PR, it works by building a tree in the `gh-pages` branch (see [here](https://github.com/peterbrightwell/mxl/tree/gh-pages)) and using the "Classic Pages experience" setting rather than a custom GitHub action. A bit more work will be needed turning this example into a usable automation, but it looks promising.

vt-tv (changes requested):
Would it be possible to avoid hardcoding absolute urls in the markdown files?

docs/Guide/Building.md
peterbrightwell: @vt-tv Unfortunately that requested change won't work as is, as ../.devcontainer/Dockerfile isn't in the docs/ directory, so won't be rendered. So I used the absolute URL to get things going.

We could have a more complete solution by running a pre-build script and use a variable, e.g.

```markdown
[Dockerfile](https://github.com/{{ repo }}/blob/{{ branch }}/.devcontainer/Dockerfile)
```

But that's a bit of a pain for authoring and difficulties arise if we want the links to work both on the github.com repo site and the github.io documentation site.  For the https://specs.amwa.tv/ NMOS rendering we have some complex scripts to rewrite links (and fix various other issues with the rendering tools), but it's all quite obscure and I certainly wouldn't want to repeat that for MXL.  

Probably worth a discussion of what we really want here.

peterbrightwell: Or the script could copy across the non-docs part of the repo, including that Dockerfile, include files etc, which might form part of the "documentation".  This is something we do for NMOS also, copying APIs/, APIs/schemas/ and examples/ into the render tree.

PR #456, opened 2026-04-15
Improve latency printout in mxl-info

pac-work:
- Currently overflows when grain index is in the future.

### DIFF
--- a/tools/mxl-info/main.cpp
+++ b/tools/mxl-info/main.cpp
@@ -84,12 +84,12 @@ namespace
                     color = fmt::color::yellow;
                 }
 
-                os << '\t' << fmt::format(fmt::fg(color), "{: >18}: {}{}, {}{:.6f}", "Latency (grains, ms)", sign, latencyGrains, sign, latencyMs)
+                os << '\t' << fmt::format(fmt::fg(color), "{: >20}: {}{}, {}{:.6f}", "Latency (grains, ms)", sign, latencyGrains, sign, latencyMs)
                    << std::endl;
             }
             else
             {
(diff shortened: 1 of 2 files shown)


jonasohland:
It looks like this PR does not contain the commits it should contain? I only see a commit that is related to rust code.

pac-work:
@jonasohland You are totally, right, sorry, was supposed to be draft, also some old commit got in the way... will fix tomorrow.

pac-work:
@jonasohland Now it should be ready.

tools/mxl-info/main.cpp
```diff
@@ -52,25 +52,44 @@ namespace
             auto const now = ::mxlGetTime();
 
             auto const currentIndex = ::mxlTimestampToIndex(&grainRate, now);
-            auto const latency = currentIndex - headIndex;
+            // The latency should never really be negative (grains in the future), but in case of clock mismatch, incorrect code or similar, this may
```
jonasohland, line 55: For all common rates, a signed `int64_t` will still be fine for the next few hundred years and could simplify this code significantly, but I am not strictly against staying unsigned of course.

jonasohland (approved):

vt-tv (approved):
LGTM

tools/mxl-info/main.cpp
pac-work: Agree that I was little paranoid, but considering one really uses this for early diagnosing of problems, I:
1. Did not want to change ("worsen") the current behavior
2. Wanted to cover most corner cases without resorting to some non-native bit arithmetics.

PR #464, opened 2026-04-21
Use system linker for Darwin presets instead of LLD

### LINKED ISSUE #463: [Bug] Darwin-Clang presets fail: LLD linker not available on macOS
### Description

The `Darwin-Clang-Debug` and `Darwin-Clang-Release` configure presets fail because the shared clang preset sets `CMAKE_LINKER_TYPE` to `LLD`. On macOS, lld is not available, neither Apple's Clang nor Homebrew's `llvm` package ship it.

```clang++: error: invalid linker name in argument '-fuse-ld=lld'```

### Steps to reproduce

1. On macOS (arm64, Apple Clang), run:
  `cmake -S . -B build --preset Darwin-Clang-Debug`
2. CMake fails at compiler detection with the error above.

### Root cause

The clang preset in `CMakePresets.json` unconditionally sets `CMAKE_LINKER_TYPE: LLD`. This works on Linux but not on macOS, where LLD does not support Mach-O.

### Suggested fix

Override `CMAKE_LINKER_TYPE` to `SYSTEM` in the darwin preset so macOS uses the default Apple linker while Linux presets continue using LLD.

aerben:
LLD does not support Mach-O, so the Darwin-Clang-* presets fail with "invalid linker name in argument '-fuse-ld=lld'". Override CMAKE_LINKER_TYPE to SYSTEM in the darwin preset.

Fixes #463

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -57,7 +57,8 @@
                 "rhs": "Darwin"
             },
             "cacheVariables": {
-                "VCPKG_TARGET_TRIPLET": "arm64-osx"
+                "VCPKG_TARGET_TRIPLET": "arm64-osx",
+                "CMAKE_LINKER_TYPE": "SYSTEM"
             },
             "vendor": {
                 "microsoft.com/VisualStudioRemoteSettings/CMake/1.0": {


KimonHoffmann (approved):
Thank you for this, looks good!

vt-tv (approved):
LGTM.  Thanks!

PR #465, opened 2026-04-21
Add mxlIsTmpFs() utility to detect RAM-backed filesystems

### LINKED ISSUE #459: [Small Improvement] Detect if a domain is backed by tmpfs. Error if it is not.
### Description

NAB 2026 interop pain point....
Perform a pre-emptive check in mxlCreateInstance to detect if a domain path is really backed by tempfs or a physical disk.  Realtime uncompressed IO of mulitple flows will never work.  This simple check could prevent lots of headaches. (make this linux only for now)

It could be simple to implement
```
#include <sys/vfs.h>   // or <sys/statfs.h> depending on system
#include <string>

bool isRamDrive(const std::string& path)
{
    struct statfs buf;
    if (statfs(path.c_str(), &buf) != 0)
        return false; // or handle error

    // TMPFS_MAGIC is defined in linux/magic.h
    return buf.f_type == TMPFS_MAGIC;
}
```

### Additional Context

_No response_

aerben:
Adds a public C API function that checks whether a given path resides on a RAM-backed filesystem.

Closes #459

### DIFF
--- a/lib/src/mxl.cpp
+++ b/lib/src/mxl.cpp
@@ -11,6 +11,15 @@
 #include "mxl-internal/Logging.hpp"
 #include "mxl-internal/PosixFlowIoFactory.hpp"
 
+#ifdef __linux__
+#   include <sys/vfs.h>
+#   include <linux/magic.h>
+#elif defined(__APPLE__)
+#   include <cstring>
+#   include <sys/mount.h>
+#   include <sys/param.h>
+#endif
+
 static char const g_mxl_version_string[] = MXL_VERSION_FULL;
 
 extern "C"
@@ -52,6 +61,38 @@ mxlInstance mxlCreateInstance(char const* in_mxlDomain, char const* in_options)
         MXL_ERROR("Failed to create instance");
         return nullptr;
     }
+}
+
+extern "C" MXL_EXPORT
+mxlStatus mxlIsTmpFs(char const* in_path, bool* out_isTmpFs)
+{
+    if ((in_path == nullptr) || (out_isTmpFs == nullptr))
+    {
+        return MXL_ERR_INVALID_ARG;
+    }
+
+#ifdef __linux__
+    using statfs_t = struct statfs;
+    auto buf = statfs_t{};
+    if (::statfs(in_path, &buf) != 0)
+    {
+        return MXL_ERR_UNKNOWN;
+    }
+    *out_isTmpFs = (buf.f_type == TMPFS_MAGIC) || (buf.f_type == RAMFS_MAGIC);
+    return MXL_STATUS_OK;
+#elif defined(__APPLE__)
+    using statfs_t = struct statfs;
+    auto buf = statfs_t{};
+    if (::statfs(in_path, &buf) != 0)
+    {
+        return MXL_ERR_UNKNOWN;
+    }
+    *out_isTmpFs = std::strcmp(buf.f_fstypename, "tmpfs") == 0;
+    return MXL_STATUS_OK;
+#else
+    *out_isTmpFs = false;
+    return MXL_STATUS_OK;
+#endif
 }
 
 extern "C"
--- a/lib/tests/test_instance.cpp
+++ b/lib/tests/test_instance.cpp
@@ -7,6 +7,54 @@
 #include <mxl/mxl.h>
 #include "Utils.hpp"
 
+#ifdef __linux__
+#   include <sys/vfs.h>
+#   include <linux/magic.h>
+
+namespace
+{
+    bool pathIsTmpFs(char const* path)
+    {
+        struct statfs buf;
+        REQUIRE(statfs(path, &buf) == 0);
+        return (buf.f_type == TMPFS_MAGIC) || (buf.f_type == RAMFS_MAGIC);
+    }
+}
+#endif
+
+TEST_CASE("mxlIsTmpFs returns MXL_ERR_INVALID_ARG for NULL path", "[mxlIsTmpFs]")
+{
+    bool isTmpFs = true;
+    REQUIRE(mxlIsTmpFs(nullptr, &isTmpFs) == MXL_ERR_INVALID_ARG);
+}
+
+TEST_CASE("mxlIsTmpFs returns MXL_ERR_INVALID_ARG for NULL out pointer", "[mxlIsTmpFs]")
+{
+    REQUIRE(mxlIsTmpFs("/tmp", nullptr) == MXL_ERR_INVALID_ARG);
+}
+
+TEST_CASE("mxlIsTmpFs returns error for non-existent path", "[mxlIsTmpFs]")
+{
+    bool isTmpFs = true;
+    REQUIRE(mxlIsTmpFs("/this/path/does/not/exist", &isTmpFs) != MXL_STATUS_OK);
+}
+
+#ifdef __linux__
+TEST_CASE("mxlIsTmpFs matches statfs filesystem type on Linux", "[mxlIsTmpFs]")
+{
+    bool isTmpFs = false;
+    REQUIRE(mxlIsTmpFs("/", &isTmpFs) == MXL_STATUS_OK);
+    REQUIRE(isTmpFs == pathIsTmpFs("/"));
+}
+
+TEST_CASE("mxlIsTmpFs detects /dev/shm as tmpfs on Linux", "[mxlIsTmpFs]")
+{
+    bool isTmpFs = false;
+    REQUIRE(mxlIsTmpFs("/dev/shm", &isTmpFs) == MXL_STATUS_OK);
+    REQUIRE(isTmpFs);
+}
+#endif
+
 TEST_CASE_PERSISTENT_FIXTURE(mxl::tests::mxlDomainFixture, "Flow readers / writers caching", "[instance]")
 {
     auto const opts = "{}";
--- a/lib/include/mxl/mxl.h
+++ b/lib/include/mxl/mxl.h
@@ -6,6 +6,7 @@
 #ifdef __cplusplus
 #   include <cstdint>
 #else
+#   include <stdbool.h>
 #   include <stdint.h>
 #endif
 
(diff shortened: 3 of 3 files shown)


lib/src/mxl.cpp
```diff
@@ -54,6 +59,27 @@ mxlInstance mxlCreateInstance(char const* in_mxlDomain, char const* in_options)
...
+extern "C" MXL_EXPORT
+bool mxlIsTmpFs(char const* in_path)
+{
+    if (in_path == nullptr)
+    {
+        return false;
+    }
+
+#ifdef __linux__
+    struct statfs buf;
```
jonasohland:
```suggestion
    auto buf = ::statfs{};
```

lib/src/mxl.cpp
```diff
@@ -54,6 +59,27 @@ mxlInstance mxlCreateInstance(char const* in_mxlDomain, char const* in_options)
     }
 }
 
+extern "C" MXL_EXPORT
+bool mxlIsTmpFs(char const* in_path)
```
jonasohland: I personally would prefer a version that returns an error and the result via an argument like:
```
mxlStatus mxlIsTmpFs(char const* path, isTmpFs *bool) 
```

If the user calling the function does not have permissions to inspect the path for which this function is called, this will incorrectly return `false`.
This will lead to confusing errors and confused users upstream.

lib/src/mxl.cpp
```diff
@@ -54,6 +59,27 @@ mxlInstance mxlCreateInstance(char const* in_mxlDomain, char const* in_options)
...
+{
+    if (in_path == nullptr)
+    {
+        return false;
+    }
+
+#ifdef __linux__
+    struct statfs buf;
+    if (statfs(in_path, &buf) != 0)
```
jonasohland:
```suggestion
    if (::statfs(in_path, &buf) != 0)
```

lib/tests/test_instance.cpp
```diff
@@ -7,6 +7,38 @@
...
+    REQUIRE_FALSE(mxlIsTmpFs("/this/path/does/not/exist"));
+}
+
+#ifdef __linux__
+TEST_CASE("mxlIsTmpFs matches statfs filesystem type on Linux", "[mxlIsTmpFs]")
+{
+    REQUIRE(mxlIsTmpFs("/") == pathIsTmpFs("/"));
+}
+#endif
+
```
jonasohland, line 57: Maybe add a test that explicitly checks for a case where a positive result is expected?
AFAIK, `/dev/shm` is always tmpfs, and I have never seen a system where `/tmp`,  `/run` or `/run/user/$UID/` aka `$XDG_RUNTIME_DIR` are not tmpfs.

jonasohland (changes requested):

lib/src/mxl.cpp
vt-tv: What about something more generic such as : mxlIsRamBasedFs(char const* path, isRamBasedFs *bool);
 
- Verify if the magic is TMPFS or RAMFS.  If OVERLAY_SUPER_MAGIC then it could either accept it blindly or parse /mnt/self/mountinfo then find the upperdir fs type)

vt-tv (changes requested):

lib/src/mxl.cpp
```diff
@@ -54,6 +59,27 @@ mxlInstance mxlCreateInstance(char const* in_mxlDomain, char const* in_options)
...
+    {
+        return false;
+    }
+    return buf.f_type == TMPFS_MAGIC || buf.f_type == RAMFS_MAGIC;
```
KimonHoffmann: Please add parentheses around the binary sub-expressions. I.e. `return (....) || (...);`

lib/tests/test_instance.cpp
```diff
@@ -7,6 +7,38 @@
...
+#   include <linux/magic.h>
+
+namespace
+{
+    bool pathIsTmpFs(char const* path)
+    {
+        struct statfs buf;
+        REQUIRE(statfs(path, &buf) == 0);
+        return buf.f_type == TMPFS_MAGIC || buf.f_type == RAMFS_MAGIC;
+    }
```
KimonHoffmann, line 21: Same comments as in corresponding code above.

KimonHoffmann (changes requested):
Thank you for this contribution.
I would like to also have a proper macOS implementation before merging this.

On macOS you can also use `::statfs()` on any path and then make a decision based on `f_type`, which is supposed to be `28` for `tmpfs`. Apple discourages relying on `f_type` though so the more resilient approach is to check, whether `f_fstypename` equals the string `tmpfs`.

lib/src/mxl.cpp
```diff
@@ -60,23 +60,24 @@ mxlInstance mxlCreateInstance(char const* in_mxlDomain, char const* in_options)
 }
 
 extern "C" MXL_EXPORT
-bool mxlIsTmpFs(char const* in_path)
+mxlStatus mxlIsTmpFs(char const* in_path, bool* out_isTmpFs)
 {
-    if (in_path == nullptr)
+    if (in_path == nullptr || out_isTmpFs == nullptr)
```
KimonHoffmann: Same comment about binary sub-expressions, as before.

lib/src/mxl.cpp
```diff
@@ -60,23 +60,24 @@ mxlInstance mxlCreateInstance(char const* in_mxlDomain, char const* in_options)
...
-    struct statfs buf;
-    if (statfs(in_path, &buf) != 0)
+    auto buf = ::statfs{};
+    if (::statfs(in_path, &buf) != 0)
     {
-        return false;
+        return MXL_ERR_UNKNOWN;
     }
-    return buf.f_type == TMPFS_MAGIC || buf.f_type == RAMFS_MAGIC;
+    *out_isTmpFs = buf.f_type == TMPFS_MAGIC || buf.f_type == RAMFS_MAGIC;
```
KimonHoffmann: And once more...

lib/src/mxl.cpp
```diff
@@ -60,23 +60,24 @@ mxlInstance mxlCreateInstance(char const* in_mxlDomain, char const* in_options)
...
+    if (in_path == nullptr || out_isTmpFs == nullptr)
     {
-        return false;
+        return MXL_ERR_INVALID_ARG;
     }
 
 #ifdef __linux__
-    struct statfs buf;
-    if (statfs(in_path, &buf) != 0)
+    auto buf = ::statfs{};
```
KimonHoffmann: Even though we try to follow AA style in case of macOS (see my header comment in the other review) this unfortunately has to be reverted back to `struct statfs but` or something like this:

```cpp
using statfs_t = struct statfs;
auto buf = statfs_t{};
```

KimonHoffmann (changes requested):
Some more small remarks on the recent changes.

lib/src/mxl.cpp
aerben: Honestly, I think this would be overkill in this case. This seems perfectly readable and unambiguous to me.

lib/src/mxl.cpp
aerben: I addressed this issue now and added macOS-specific implementation. Would you kindly review again?

lib/src/mxl.cpp
aerben: I think overlayfs detection is out of scope for this PR. It adds significant complexity (parsing /proc/self/mountinfo). Happy to revisit in a follow-up if there's demand for it.

lib/src/mxl.cpp
KimonHoffmann: Please add them anyway as this is the style we have been following in the rest of the codebase until now.

lib/src/mxl.cpp
```diff
@@ -11,6 +11,14 @@
 #include "mxl-internal/Logging.hpp"
 #include "mxl-internal/PosixFlowIoFactory.hpp"
 
+#ifdef __linux__
+#   include <sys/vfs.h>
+#   include <linux/magic.h>
+#elif defined(__APPLE__)
+#   include <cstring>
+#   include <sys/mount.h>
```
KimonHoffmann, line 19: According to the man page you also need to include `sys/param.h`.

KimonHoffmann (changes requested):
Looks good, just one small remark regarding the set of system includes.

lib/src/mxl.cpp
aerben: Sure! I added the parens

lib/src/mxl.cpp
aerben: Done

KimonHoffmann (approved):
Looks good now, thank you!

As a final step could you please rebase the branch on `main` and drop the **Update backport workflow to use Node.js >20 actions** commit along the way?

aerben:
@KimonHoffmann Rebased onto main and squashed into a single commit.

jonasohland (approved):
LGTM

lib/src/mxl.cpp
aerben: @vt-tv bump. 
Your opinion on this?

vt-tv: I fear we may be breaking volumes running in containers.  What's your gut feeling? should we 'allow' overlayfs in order to avoid potentially breaking the world?

aerben: I don't think we'd break containers. statfs() sees through to the actual mount type. RAM-backed mounts still report as tmpfs even inside containers.
 
 ```bash
  # /dev/shm — Docker mounts this as tmpfs by default
  docker run --rm alpine stat -f -c 'Path: /dev/shm , Type: 0x%t' /dev/shm
  # Path: /dev/shm , Type: 0x1021994 (TMPFS_MAGIC)

  # /tmp — explicitly mounted as tmpfs via --tmpfs flag
  docker run --rm --tmpfs /tmp alpine stat -f -c 'Path: /tmp , Type: 0x%t' /tmp
  # Path: /tmp , Type: 0x1021994 (TMPFS_MAGIC)

  # / — container root filesystem (overlayfs)
  docker run --rm alpine stat -f -c 'Path: / , Type: 0x%t' /
  # Path: / , Type: 0x794c7630 (OVERLAY_FS)
  ```

PR #467, opened 2026-04-21
Add flow permissions checks when creating flowreaders and flowwriters

vt-tv:
check permissions in flow writers and readers

### DIFF
--- a/lib/internal/include/mxl-internal/FlowWriter.hpp
+++ b/lib/internal/include/mxl-internal/FlowWriter.hpp
@@ -77,8 +77,8 @@ namespace mxl::lib
         [[nodiscard]]
         bool checkPermissions() const;
 
-        explicit FlowWriter(uuids::uuid&& flowId, std::filesystem::path const& domain);
-        explicit FlowWriter(uuids::uuid const& flowId, std::filesystem::path const& domain);
+        FlowWriter(uuids::uuid&& flowId, std::filesystem::path const& domain);
+        FlowWriter(uuids::uuid const& flowId, std::filesystem::path const& domain);
 
     private:
         uuids::uuid _flowId;
--- a/lib/internal/include/mxl-internal/FlowWriter.hpp
+++ b/lib/internal/include/mxl-internal/FlowWriter.hpp
@@ -70,10 +70,18 @@ namespace mxl::lib
         virtual ~FlowWriter();
 
     protected:
-        explicit FlowWriter(uuids::uuid&& flowId);
-        explicit FlowWriter(uuids::uuid const& flowId);
+        /**
+         * Verifies that we can traverse and write into the domain directory.
+         * \return true if the flow domain is writable, false otherwise.
+         */
+        [[nodiscard]]
+        bool checkPermissions() const;
+
+        explicit FlowWriter(uuids::uuid&& flowId, std::filesystem::path const& domain);
+        explicit FlowWriter(uuids::uuid const& flowId, std::filesystem::path const& domain);
 
     private:
         uuids::uuid _flowId;
+        std::filesystem::path _domain;
     };
 }
--- a/lib/internal/src/FlowReader.cpp
+++ b/lib/internal/src/FlowReader.cpp
@@ -3,6 +3,7 @@
 
 #include "mxl-internal/FlowReader.hpp"
 #include <utility>
+#include "mxl-internal/PathUtils.hpp"
 
 namespace mxl::lib
 {
@@ -26,5 +27,11 @@ namespace mxl::lib
     std::filesystem::path const& FlowReader::getDomain() const
     {
         return _domain;
+    }
+
+    bool FlowReader::checkPermissions() const
+    {
+        auto flowDefPath = makeFlowDescriptorFilePath(_domain, uuids::to_string(_flowId));
+        return std::filesystem::exists(flowDefPath) && std::filesystem::is_regular_file(flowDefPath) && (access(flowDefPath.c_str(), R_OK) == 0);
     }
 }
--- a/lib/internal/src/FlowReader.cpp
+++ b/lib/internal/src/FlowReader.cpp
@@ -3,6 +3,7 @@
 
 #include "mxl-internal/FlowReader.hpp"
 #include <utility>
+#include <unistd.h>
 #include "mxl-internal/PathUtils.hpp"
 
 namespace mxl::lib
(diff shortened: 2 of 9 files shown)


lib/internal/include/mxl-internal/FlowWriter.hpp
```diff
@@ -70,10 +70,18 @@ namespace mxl::lib
...
-        explicit FlowWriter(uuids::uuid const& flowId);
+        /**
+         * Verifies that we can traverse and write into the domain directory.
+         * \return true if the flow domain is writable, false otherwise.
+         */
+        [[nodiscard]]
+        bool checkPermissions() const;
+
+        explicit FlowWriter(uuids::uuid&& flowId, std::filesystem::path const& domain);
+        explicit FlowWriter(uuids::uuid const& flowId, std::filesystem::path const& domain);
```
KimonHoffmann: You can remove the `explicit` keywords, as both constructors have more than one non-optional arguments and thus don't qualify for implicit initialization anyway.

lib/internal/src/FlowReader.cpp
```diff
@@ -27,4 +28,10 @@ namespace mxl::lib
     {
         return _domain;
     }
+
+    bool FlowReader::checkPermissions() const
+    {
+        auto flowDefPath = makeFlowDescriptorFilePath(_domain, uuids::to_string(_flowId));
```
KimonHoffmann: Can be `const`.

lib/internal/src/PosixContinuousFlowReader.cpp
```diff
@@ -15,7 +15,12 @@ namespace mxl::lib
         , _flowData{std::move(data)}
         , _channelCount{_flowData->channelCount()}
         , _bufferLength{_flowData->channelBufferLength()}
-    {}
+    {
+        if (!checkPermissions())
+        {
+            throw std::runtime_error("Flow is not accessible due to insufficient permissions.");
```
KimonHoffmann: Please use curly braces for constructor invocations.

lib/internal/src/PosixContinuousFlowWriter.cpp
```diff
@@ -18,6 +19,11 @@ namespace mxl::lib
         , _earlySyncThreshold{}
         , _lastSyncSampleBatch{}
     {
+        if (!checkPermissions())
+        {
+            throw std::runtime_error("Flow is not accessible due to insufficient permissions.");
```
KimonHoffmann: Please use curly braces for constructor invocations.

lib/internal/tests/test_domainwatcher.cpp
```diff
@@ -102,7 +102,7 @@ struct MockFlowFiles
 struct MockWriter : mxl::lib::DiscreteFlowWriter
 {
     MockWriter(std::filesystem::path const& domain, uuids::uuid id, DomainWatcher::ptr const& watcher)
-        : mxl::lib::DiscreteFlowWriter(id)
+        : mxl::lib::DiscreteFlowWriter(id, domain)
```
KimonHoffmann: Please use curly braces for constructor invocations.

lib/internal/src/FlowReader.cpp
```diff
@@ -27,4 +28,10 @@ namespace mxl::lib
...
+        return std::filesystem::exists(flowDefPath) && std::filesystem::is_regular_file(flowDefPath) && (access(flowDefPath.c_str(), R_OK) == 0);
```
KimonHoffmann: `is_regular_file` subsumes `exists` so we can save on a syscall here. Also please add full qualification for `access`.

lib/internal/src/FlowWriter.cpp
```diff
@@ -20,4 +22,10 @@ namespace mxl::lib
     {
         return _flowId;
     }
+
+    bool FlowWriter::checkPermissions() const
+    {
+        // Verify that the domain exists, is a directory and that we can traverse and write into it.
+        return std::filesystem::exists(_domain) && std::filesystem::is_directory(_domain) && (access(_domain.c_str(), X_OK | W_OK) == 0);
```
KimonHoffmann: Similar comment as above:
`is_directory` subsumes `exists` so we can save on a syscall here. Also please add full qualification for `access`.

KimonHoffmann (changes requested):
Thank you for this!
Just a few small remarks...

lib/internal/include/mxl-internal/FlowWriter.hpp
```diff
@@ -70,10 +70,18 @@ namespace mxl::lib
...
 
     protected:
-        explicit FlowWriter(uuids::uuid&& flowId);
-        explicit FlowWriter(uuids::uuid const& flowId);
+        /**
+         * Verifies that we can traverse and write into the domain directory.
+         * \return true if the flow domain is writable, false otherwise.
+         */
+        [[nodiscard]]
+        bool checkPermissions() const;
```
jonasohland, line 78: Why do we need this in the first place? Should the opening of the flow files or their creation in the shared memory abstraction not already throw an error?

jonasohland (changes requested):

lib/internal/include/mxl-internal/FlowWriter.hpp
vt-tv: It is more explicit.  At the moment it is just failing without explanations.

vt-tv: And I had to help a few partners debug their issues related to permissions during NAB...

jonasohland: For 1.1 we I think we should think about better error reporting in general. If the error inside the shared memory abstraction would be caught and reported in the right place, there would be no need for such a preemptive check.

jonasohland (approved):

lib/internal/include/mxl-internal/FlowWriter.hpp
KimonHoffmann: I agree, these separate checks always smell like TOCTOU confusions waiting to happen. But I agree that until then it is an improvement to enable applications to do better error reporting in benign error scenarios.

lib/internal/src/PosixContinuousFlowReader.cpp
vt-tv: done

lib/internal/src/FlowReader.cpp
vt-tv: done

lib/internal/include/mxl-internal/FlowWriter.hpp
vt-tv: done

lib/internal/tests/test_domainwatcher.cpp
vt-tv: done

lib/internal/src/FlowReader.cpp
vt-tv: done

lib/internal/src/FlowWriter.cpp
vt-tv: done

lib/internal/src/PosixContinuousFlowWriter.cpp
vt-tv: Done

KimonHoffmann (approved):
Looks good, thank you!

PR #470, opened 2026-04-22
Fix race condition in DomainWatcher::stop()

### LINKED ISSUE #446: "DomainWatcher thread start/stop behavior" test intermittent failure
The "DomainWatcher thread start/stop behavior" regression test fails intermittently.

On my system, the failure rate is approximately 18–25 failures per 100 runs under normal conditions.

When the system is loaded using `stress-ng --cpu $(nproc))`, the failure rate drops to approximately 0–4 failures per 100 runs.

This behavior is observed at the v1.0.0 tag.

**System**
Ubunut 24.04
Static release buid
GCC 13
AMD 9700x cpu

**Extra**
Here's my test scirpt:
```
#!/usr/bin/env bash
set -euo pipefail

runs=100
test_dir="$HOME/mxldev/build/mxl/build/Linux-GCC-Release/static"
test_regex='^DomainWatcher thread start/stop behavior$'

failures=0

for ((i = 1; i <= runs; ++i)); do
    echo "Run $i/$runs"
    if ! ctest --test-dir "$test_dir" -V -R "$test_regex" --output-on-failure; then
        ((failures+=1))
    fi
done

echo
echo "Failures: $failures / $runs"

awk -v failures="$failures" -v runs="$runs" 'BEGIN {
    printf "Failure rate: %.2f%%\n", 100 * failures / runs
    if (failures > 0)
        printf "Observed average: 1 failure every %.2f runs\n", runs / failures
    else
        printf "Observed average: no failures observed\n"
}'
```

aerben:
stop() only set the _running flag without joining the thread, leaving a window where the event loop could process one more inotify event before checking the flag. Join the thread in stop() to guarantee the event loop has fully exited before the caller proceeds.

Fixes #446

### DIFF
--- a/lib/internal/include/mxl-internal/DomainWatcher.hpp
+++ b/lib/internal/include/mxl-internal/DomainWatcher.hpp
@@ -93,6 +93,10 @@ namespace mxl::lib
         void stop()
         {
             _running = false;
+            if (_watchThread.joinable())
+            {
+                _watchThread.join();
+            }
         }
 
         /** \brief Returns the number of writers registered for flow id 'id'
--- a/lib/internal/src/DomainWatcher.cpp
+++ b/lib/internal/src/DomainWatcher.cpp
@@ -111,11 +111,7 @@ namespace mxl::lib
 
     DomainWatcher::~DomainWatcher()
     {
-        _running = false;
-        if (_watchThread.joinable())
-        {
-            _watchThread.join();
-        }
+        stop();
 
 #ifdef __APPLE__
         ::close(_kq);


jonasohland:
Makes sense, I wonder why I did not fix that in the first place...
The only downside is that now a call to `stop()` now blocks 250ms in the worst case. It would be great to have an eventfd that wakes up the thread waiting on epoll when a stop is requested. That would also speed up every invocation of `mxl-info`, which always takes >250ms to run. That would be out of scope for this fix though.

The code in stop is now equivalent to the beginning of the destructor on line 114. Could you also replace the code there with a call to the modified `stop()`?

aerben:
Done @jonasohland 
Could you re-check?

jonasohland (approved):

vt-tv (approved):
good catch. Thanks!

PR #472, opened 2026-04-22
Use eventfd to immediately stop the domainwatcher thread

jonasohland:
Improves the shutdown behavior of all applications that create and destroy an mxl instance (on linux).
In the worst case this takes 250 milliseconds for `epoll_wait` to return and the DomainWatcher thread to exit.  
This makes for example the mxl-info tool feel kind of sluggish. It's not totally catastrophic, but I feel like it is bad manners to block the users thread for 250ms without a good reason.

Before:
```
time MXL_LOG_LEVEL=debug build/Linux-GCC-Debug/tools/mxl-info/mxl-info -d /dev/shm
[2026-04-22 16:34:03.140] [console] [debug] [Instance.cpp:425] History duration set to 200000000 ns
[2026-04-22 16:34:03.140] [console] [debug] [Instance.cpp:92] Instance created. MXL Domain: /dev/shm
[2026-04-22 16:34:03.390] [console] [debug] [Instance.cpp:99] Instance destroyed.
________________________________________________________
Executed in  258.46 millis    fish           external
   usr time    3.26 millis    0.00 millis    3.26 millis
   sys time    6.19 millis    1.81 millis    4.38 millis
```

After:
```
time MXL_LOG_LEVEL=debug build/Linux-GCC-Debug/tools/mxl-info/mxl-info -d /dev/shm
[2026-04-22 16:31:51.262] [console] [debug] [Instance.cpp:425] History duration set to 200000000 ns
[2026-04-22 16:31:51.262] [console] [debug] [Instance.cpp:92] Instance created. MXL Domain: /dev/shm
[2026-04-22 16:31:51.262] [console] [debug] [DomainWatcher.cpp:331] Domain watcher thread exit requested
[2026-04-22 16:31:51.262] [console] [debug] [Instance.cpp:99] Instance destroyed.
________________________________________________________
Executed in    7.85 millis    fish           external
   usr time    3.27 millis  390.00 micros    2.88 millis
   sys time    4.72 millis  875.00 micros    3.85 millis
```


vt-tv (approved):
LGTM

KimonHoffmann (commented):
Thank you for the submission! It would be great to have a mechanism on macOS as well to maintain feature parity. The macOS implementation could use a pipe for that purpose.

In order to limit the amount of differences spread over the entire class it might even make sense to use a pipe on both Linux and macOS!?

jonasohland:
> Thank you for the submission! It would be great to have a mechanism on macOS as well to maintain feature parity. The macOS implementation could use a pipe for that purpose.
> 
> In order to limit the amount of differences spread over the entire class it might even make sense to use a pipe on both Linux and macOS!?

A bit of research turned up `EVFILT_USER` as a replacement for `eventfd` at least for this usecase, so I would use that instead of a pipe. 

I’m also not a fan of all the #ifdefs in this file.
It currently looks like the GPU memory sharing code will also rely on sockets and, consequently, a polling and notification mechanism.
I would be happy to do a larger refactor at that point and encapsulate aspects such as epoll/kqueue, events, and sockets into proper RAII-safe abstractions.
However, for now, if it’s okay with you, I would prefer to just add another #ifdef...

KimonHoffmann:
> I’m also not a fan of all the #ifdefs in this file. It currently looks like the GPU memory sharing code will also rely on sockets and, consequently, a polling and notification mechanism. I would be happy to do a larger refactor at that point and encapsulate aspects such as epoll/kqueue, events, and sockets into proper RAII-safe abstractions. However, for now, if it’s okay with you, I would prefer to just add another #ifdef...

Sure, and I agree that we should tackle this in a more general fashion down the road.

pedro-alves-ferreira (approved):
LGTM

vt-tv:
@jonasohland  this broke the macOS build.

KimonHoffmann:
If I see this correctly this PR was not yet ready to be merged. We were still discussing feature parity between macOS and linux by means of `kqueue`s `EVFILT_USER`.

PR #473, opened 2026-04-22
Remove internal dependencies of mxl target from installed configs for shared library build

### LINKED ISSUE #468: Generated CMake config files are incorrect
When you install the MXL built as shared library, the generated files force the user to have MXL internal dependencies installed - that is:

```cmake
find_dependency(spdlog CONFIG)
find_dependency(fmt CONFIG)
find_dependency(stduuid CONFIG)
find_dependency(picojson)
```

Internal, statically linked and hidden MXL's dependencies shouldn't affect the user consuming the library but the way it is specified now - they do.

For some reason (might be related) we also export internal target - it's not used anywhere but it's there for no obvious reason:

```cmake
# Create imported target mxl::mxl-internal-headers
add_library(mxl::mxl-internal-headers INTERFACE IMPORTED)

set_target_properties(mxl::mxl-internal-headers PROPERTIES
  INTERFACE_COMPILE_FEATURES "cxx_std_20"
  INTERFACE_LINK_LIBRARIES "mxl::mxl-headers;stduuid;spdlog::spdlog;fmt::fmt;picojson::picojson"
)
```

There's also findpicojson.cmake file that we install, we should remove it as well.

**To Reproduce**
1. Build MXL using CMake
2. Install it
3. Inspect lib/cmake/mxl/*.cmake files

k0zmo:
Fixes #468


lib/internal/CMakeLists.txt
```diff
@@ -17,12 +17,32 @@ target_include_directories(mxl-internal-headers
...
+                $<BUILD_INTERFACE:spdlog::spdlog>
+                $<BUILD_INTERFACE:fmt::fmt>
+        )
+else()
+    target_link_libraries(mxl-internal-headers
+            INTERFACE
+                spdlog::spdlog
+                fmt::fmt
+        )
+endif()
```
KimonHoffmann, line 44: Maybe these two can be rolled into a single generator expression?

```
$<BUILD_SHARED_LIBS:$<BUILD_INTERFACE:spdlog::spdlog>>
```

lib/internal/CMakeLists.txt
```diff
@@ -106,12 +126,14 @@ install(TARGETS mxl-common EXPORT ${PROJECT_NAME}-targets
...
-        COMPONENT ${PROJECT_NAME}-dev
-    )
+# Static installs need this exported interface target because mxl-common's
+# imported link interface refers to it. Shared installs do not export it.
+if (NOT BUILD_SHARED_LIBS)
+    install(TARGETS mxl-internal-headers
+            EXPORT ${PROJECT_NAME}-targets
+            COMPONENT ${PROJECT_NAME}-dev
+        )
+endif()
```
KimonHoffmann, line 136: Back when I added these internal headers to the install targets this was a quick fix to resolve the missing dependency, because I was in a hurry and didn't have much time to really think about it. Even then I quickly contemplated, whether there is a better solution to avoid this dependency entirely when installed and now seems to be a good opportunity to pick this up again.

Do you have an idea if/how that might be achieved?

KimonHoffmann (approved):
Looks good to me, thanks a lot for cleaning this up!
No blockers from my side only a minor suggestion and the question, whether we might address this more completely from another angle.

vt-tv (approved):
Thanks for your contribution. LGTM

lib/internal/CMakeLists.txt
k0zmo: I could only make it work with following:
```cmake
target_link_libraries(mxl-internal-headers
        INTERFACE
            $<$<BOOL:${BUILD_SHARED_LIBS}>:$<BUILD_INTERFACE:spdlog::spdlog>>
            $<$<BOOL:${BUILD_SHARED_LIBS}>:$<BUILD_INTERFACE:fmt::fmt>>
            $<$<NOT:$<BOOL:${BUILD_SHARED_LIBS}>>:spdlog::spdlog>
            $<$<NOT:$<BOOL:${BUILD_SHARED_LIBS}>>:fmt::fmt>
    )
```
Which doesn't look better IMHO

lib/internal/CMakeLists.txt
k0zmo: I'll take a look at it

lib/internal/CMakeLists.txt
KimonHoffmann: Agreed 🙂

lib/internal/CMakeLists.txt
KimonHoffmann: Thank you!

PR #478, opened 2026-04-23
Replace mxl-common with mxl-internal-objects OBJECT library

### LINKED ISSUE #477: Leaking spdlog symbols from mxl-common
Following the ask from #473 - currently we build `mxl` and `mxl-common`, both as shared library (`mxl` depends on the latter), in default configuration. There are few problems with that approach.
- `RPATH` is not set on `libmxl.so` hence the user of it is not able to find the `libmxl-common.so` (`RPATH`, unless using `-Wl,--disable-old-dtags` is not transitive)
- `spdlog` symbols are exported from `libmxl-common.so` (because `libmxl.so` needs it) but that makes it problematic if you link it to something that uses `spdlog` as well in different version. For example, your application links to both `libmxl` and `librivermax` - both currently exporting symbols from `spdlog` and linked with different versions. This leads to a `SEGV` on first use of `spdlog`. 

I propose to convert `mxl-common` library into `mxl-internal-objects` `OBJECT` library, which would also fall in line similarly to `mxl-fabrics-objects`. This would result in a single library - `libmxl.so` (not including the `libmxl-fabrics.so`), with no internal symbols leaking.

k0zmo:
Fixes #477


lib/CMakeLists.txt
```diff
@@ -61,14 +61,14 @@ target_sources(mxl
             src/flow.cpp
             src/mxl.cpp
             src/time.cpp
+            $<TARGET_OBJECTS:mxl-internal-objects>
```
KimonHoffmann: Why not link the `mxl-internal-objects` via `target_link_libraries()` below?

KimonHoffmann (changes requested):
Thank you for this!
The transformation is fine by me, but IIRC one reason why we opted for a shared library was that the fabrics implementation also needs some of the internal functionality. Can you comment on this @jonasohland ?

KimonHoffmann:
> The transformation is fine by me, but IIRC one reason why we opted for a shared library was that the fabrics implementation also needs some of the internal functionality. Can you comment on this @jonasohland ?

I just realized that you guys are already discussing this in the linked issue, my bad. So please ignore my shoutout here. 😄

vt-tv:
@KimonHoffmann  what's next? is this ok?

lib/CMakeLists.txt
k0zmo: Fixed

k0zmo: I just tested it and with `BUILD_SHARED_LIBS=OFF` and using `mxl-internal-objects` via `target_link_libraries()` CMake complains:
```
CMake Error: install(EXPORT "mxl-targets" ...) includes target "mxl" which requires target "mxl-internal-objects" that is not in any export set.
```

KimonHoffmann: See here for two suggested workarounds to this problem:
https://gitlab.kitware.com/cmake/cmake/-/work_items/23168

To me this *feels* like a cmake bug, but I have not yet made an attempt to fully understand it :wink:.

KimonHoffmann: This thread sheds some further light on the issue:
https://gitlab.kitware.com/cmake/cmake/-/work_items/18935

Seems like just installing the objects library is the "cleanest" way to go, since this transforms the library into an `INTERFACE` library and effectively makes it a phony target that at install time, as long as we're not explicitly telling ti install its `OBJECTS`.

k0zmo: Thanks for the links. It should be fixed for both configurations now.

KimonHoffmann (approved):
Looks good, thank you!

vt-tv:
My mistake. I hit merge before approving. consider I approved this PR.

Issue #476, opened 2026-04-23
[Small Improvement] mxl-data-probe tool


k0zmo:
### Description

While testing ANC/data compatibility between vendors I wrote a tool to probe grains. If there's any demand for such thing I can open a PR.

The readme is following:

```md
## mxl-data-probe

A tool that opens an MXL ancillary Data flow (`media_type: video/smpte291`), reads one or more grains starting at the current head index, and prints the RFC-8331 ANC elements found in each grain.

```bash
./mxl-data-probe [OPTIONS]

OPTIONS:
  -h,     --help              Print this help message and exit
          --version           Display program version information and exit
  -d,     --domain TEXT:DIR REQUIRED
                              The MXL domain directory
  -f,     --flow TEXT REQUIRED
                              The flow id to read
  -c,     --count UINT [1]    Number of grains to read from the current head index
  -t,     --timeout-ms UINT [1000]
                              Timeout per grain read in milliseconds
```

Example:

```bash
./mxl-data-probe -d /dev/shm/mxl -f db3bd465-2772-484f-8fac-830b0471258b --count 3
```

Sample output:

<img width="2300" height="1015" alt="Image" src="https://github.com/user-attachments/assets/0805cfad-7104-489c-9502-fa0215947505" />

jonasohland:
Awesome! I think it would make a lot of sense to include this in the repo. It would be the missing mxl-gst-sink equivalent for ANC flows.

garethsb:
Resolved by #479, right?

> It would be the missing mxl-gst-sink equivalent for ANC flows.

There will shortly also be ANC support in `mxlsrc` and `mxlsink` too (#500).

PR #479, opened 2026-04-23
Add mxl-data-probe tool

k0zmo:
Implements #476

### DIFF
--- a/docs/Tools.md
+++ b/docs/Tools.md
@@ -107,6 +107,31 @@ Hint : Live monitoring of a flow details (updates every second)
 watch -n 1 -p ./mxl-info mxl:///dev/shm/mxl?id=5fbec3b1-1b0f-417d-9059-8b94a47197ed
 ```
 
+## mxl-data-probe
+
+A tool that opens an MXL ancillary Data flow (`media_type: video/smpte291`), reads one or more grains starting at the current head index, and prints the RFC-8331 ANC elements found in each grain.
+
+```bash
+./mxl-data-probe [OPTIONS]
+
+OPTIONS:
+  -h,     --help              Print this help message and exit
+          --version           Display program version information and exit
+  -d,     --domain TEXT:DIR REQUIRED
+                              The MXL domain directory
+  -f,     --flow TEXT REQUIRED
+                              The flow id to read
+  -c,     --count UINT [1]    Number of grains to read from the current head index
+  -t,     --timeout-ms UINT [1000]
+                              Timeout per grain read in milliseconds
+```
+
+Example:
+
+```bash
+./mxl-data-probe -d /dev/shm/mxl -f db3bd465-2772-484f-8fac-830b0471258b --count 3
+```
+
 ## mxl-gst-testsrc
 
 A binary that uses the gstreamer 'videotestsrc' and 'audiotestsrc' elements to produce video grains and/or audio samples which will be pushed to a MXL Flow. The flow is configured from a NMOS Flow json file. Here's an example of such file :
@@ -362,4 +387,4 @@ OPTIONS:
 
 ## Gstreamer filters
 
-MXL Source and Sink filters for GStreamer are available [here](../rust/gst-mxl-rs/README.md). 
+MXL Source and Sink filters for GStreamer are available [here](../rust/gst-mxl-rs/README.md).
--- a/docs/Tools.md
+++ b/docs/Tools.md
@@ -112,24 +112,35 @@ watch -n 1 -p ./mxl-info mxl:///dev/shm/mxl?id=5fbec3b1-1b0f-417d-9059-8b94a4719
 A tool that opens an MXL ancillary Data flow (`media_type: video/smpte291`), reads one or more grains starting at the current head index, and prints the RFC-8331 ANC elements found in each grain.
 
 ```bash
-./mxl-data-probe [OPTIONS]
+./mxl-data-probe [OPTIONS] [ADDRESS...]
+
+POSITIONALS:
+  ADDRESS TEXT ...            MXL URI
 
 OPTIONS:
   -h,     --help              Print this help message and exit
           --version           Display program version information and exit
-  -d,     --domain TEXT:DIR REQUIRED
-                              The MXL domain directory
-  -f,     --flow TEXT REQUIRED
-                              The flow id to read
+  -d,     --domain TEXT:DIR   The MXL domain directory
+  -f,     --flow TEXT         The flow id to read
   -c,     --count UINT [1]    Number of grains to read from the current head index
   -t,     --timeout-ms UINT [1000]
                               Timeout per grain read in milliseconds
+
+MXL URI format:
+mxl://[authority[:port]]/domain[?id=...]
+See: https://github.com/dmf-mxl/mxl/docs/Addressability.md
 ```
 
-Example:
+Example using command line options:
 
 ```bash
 ./mxl-data-probe -d /dev/shm/mxl -f db3bd465-2772-484f-8fac-830b0471258b --count 3
+```
+
+Example using an MXL URI:
+
+```bash
+./mxl-data-probe mxl:///dev/shm/mxl?id=db3bd465-2772-484f-8fac-830b0471258b --count 3
 ```
 
 ## mxl-gst-testsrc
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -1,22 +1,6 @@
 # SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project.
 # SPDX-License-Identifier: Apache-2.0
 
-include(FetchContent)
-
-FetchContent_Declare(
-    simple_uri_parser
-    GIT_REPOSITORY https://github.com/jholloc/simple-uri-parser.git
-    GIT_TAG        v1.0.0
-)
-
-FetchContent_MakeAvailable(simple_uri_parser)
-add_library(simple_uri_parser INTERFACE)
-
-target_include_directories(simple_uri_parser
-    INTERFACE
-        ${simple_uri_parser_SOURCE_DIR}
-)
-
 add_subdirectory(mxl-info)
 add_subdirectory(mxl-data-probe)
 add_subdirectory(mxl-gst)
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -1,6 +1,19 @@
 # SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project.
 # SPDX-License-Identifier: Apache-2.0
 
+include(GNUInstallDirs)
+
+cmake_path(RELATIVE_PATH CMAKE_INSTALL_FULL_LIBDIR
+           BASE_DIRECTORY "${CMAKE_INSTALL_FULL_BINDIR}"
+           OUTPUT_VARIABLE MXL_TOOLS_INSTALL_RPATH
+    )
+
+if(APPLE)
+    set(MXL_TOOLS_INSTALL_RPATH "@loader_path/${MXL_TOOLS_INSTALL_RPATH}")
+else()
+    set(MXL_TOOLS_INSTALL_RPATH "$ORIGIN/${MXL_TOOLS_INSTALL_RPATH}")
+endif()
+
 add_subdirectory(mxl-info)
 add_subdirectory(mxl-data-probe)
 add_subdirectory(mxl-gst)
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -1,6 +1,22 @@
 # SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project.
 # SPDX-License-Identifier: Apache-2.0
 
+include(FetchContent)
+
+FetchContent_Declare(
+    simple_uri_parser
+    GIT_REPOSITORY https://github.com/jholloc/simple-uri-parser.git
+    GIT_TAG        v1.0.0
+)
+
+FetchContent_MakeAvailable(simple_uri_parser)
+add_library(simple_uri_parser INTERFACE)
+
+target_include_directories(simple_uri_parser
+    INTERFACE
+        ${simple_uri_parser_SOURCE_DIR}
+)
+
 add_subdirectory(mxl-info)
 add_subdirectory(mxl-data-probe)
 add_subdirectory(mxl-gst)
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 add_subdirectory(mxl-info)
+add_subdirectory(mxl-data-probe)
 add_subdirectory(mxl-gst)
 
 if (MXL_ENABLE_FABRICS_OFI)
--- a/tools/mxl-data-probe/CMakeLists.txt
+++ b/tools/mxl-data-probe/CMakeLists.txt
@@ -49,7 +49,7 @@ target_link_libraries(mxl-data-probe
 
 set_target_properties(mxl-data-probe
         PROPERTIES
-            INSTALL_RPATH "$ORIGIN/../lib"
+            INSTALL_RPATH "${MXL_TOOLS_INSTALL_RPATH}"
     )
 
 install(TARGETS mxl-data-probe
--- a/tools/mxl-data-probe/CMakeLists.txt
+++ b/tools/mxl-data-probe/CMakeLists.txt
@@ -1,0 +1,53 @@
+# SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project.
+# SPDX-License-Identifier: Apache-2.0
+
+include(GNUInstallDirs)
+
+add_executable(mxl-data-probe)
+target_compile_features(mxl-data-probe
+        PRIVATE
+            cxx_std_20
+    )
+set_target_properties(mxl-data-probe
+        PROPERTIES
+            POSITION_INDEPENDENT_CODE    ON
+            VISIBILITY_INLINES_HIDDEN    ON
+            C_VISIBILITY_PRESET          hidden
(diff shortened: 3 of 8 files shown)


docs/Tools.md
```diff
@@ -107,6 +107,31 @@ Hint : Live monitoring of a flow details (updates every second)
...
+  -t,     --timeout-ms UINT [1000]
+                              Timeout per grain read in milliseconds
+```
+
+Example:
+
+```bash
+./mxl-data-probe -d /dev/shm/mxl -f db3bd465-2772-484f-8fac-830b0471258b --count 3
+```
+
```
vt-tv, line 139: could you also support the standard mxl uri?  mxl://domain?id=flowid
(if the domain is /dev/shm/mxl then that would be mxl:///dev/shm/mxl?id=...

tools/mxl-data-probe/main.cpp
```diff
@@ -0,0 +1,467 @@
...
+#include <mxl/flow.h>
+#include <mxl/flowinfo.h>
+#include <mxl/mxl.h>
+
+namespace
+{
+    constexpr std::size_t Rfc8331HeaderSizeBytes = 6;
+    constexpr std::uint16_t Lower8Bits = 0x00ffU;
+    constexpr std::uint16_t Lower10Bits = 0x03ffU;
+    constexpr std::uint16_t Lower11Bits = 0x07ffU;
```
KimonHoffmann: Please prefer right hand types for these definitions.
I.e.

```cpp
constexpr auto const XXX = std::bla{...};
```

tools/mxl-data-probe/main.cpp
```diff
@@ -0,0 +1,467 @@
...
+
+    char const* statusToString(::mxlStatus status) noexcept
```
KimonHoffmann: Even though it is a helper function in an anonymous namespace. this can be declared `constexpr` and should probably be declared `[[nodiscard]]`.

tools/mxl-data-probe/main.cpp
```diff
@@ -0,0 +1,467 @@
...
+
+        ~ScopedMxlInstance()
+        {
+            (void)::mxlDestroyInstance(_instance);
+        }
+
+        ScopedMxlInstance(ScopedMxlInstance const&) = delete;
+        ScopedMxlInstance& operator=(ScopedMxlInstance const&) = delete;
+
+        ::mxlInstance get() const noexcept
```
KimonHoffmann: `constexpr` + `[[nodsicard]]`

tools/mxl-data-probe/main.cpp
```diff
@@ -0,0 +1,467 @@
...
+        {
+            return _instance;
+        }
+
+    private:
+        ::mxlInstance _instance = nullptr;
```
KimonHoffmann: Please remove the initialized, since it's confusing two have two places to look for member initializations and in this case it is unused anyway.

tools/mxl-data-probe/main.cpp
```diff
@@ -0,0 +1,467 @@
...
+    };
+
+    class ScopedFlowReader
+    {
+    public:
+        ScopedFlowReader(::mxlInstance instance, std::string const& flowId)
+            : _instance{instance}
```
KimonHoffmann, line 93: To avoid the split initialization:

```cpp
ScopedFlowReader(::mxlInstance instance, std::string const& flowId)
    : _instance{instance}
    , _reader{nullptr}
{
    // ...
}
```

tools/mxl-data-probe/main.cpp
```diff
@@ -0,0 +1,467 @@
...
+            auto const status = ::mxlCreateFlowReader(instance, flowId.c_str(), "", &_reader);
+            if (status != MXL_STATUS_OK)
+            {
+                throw std::runtime_error{"Failed to create flow reader: " + std::string{statusToString(status)}};
+            }
+        }
+
+        ~ScopedFlowReader()
+        {
+            if (_reader != nullptr)
```
KimonHoffmann: `_reader == nullptr` can really never happen, can it?

tools/mxl-data-probe/main.cpp
```diff
@@ -0,0 +1,467 @@
...
+            {
+                (void)::mxlReleaseFlowReader(_instance, _reader);
+            }
+        }
+
+        ScopedFlowReader(ScopedFlowReader const&) = delete;
+        ScopedFlowReader& operator=(ScopedFlowReader const&) = delete;
+
+        ::mxlFlowReader get() const noexcept
```
KimonHoffmann, line 112: `constexpr` +`[[nodiscard]]`

tools/mxl-data-probe/main.cpp
```diff
@@ -0,0 +1,467 @@
...
+    class BigEndianWordReader
+    {
+    public:
+        explicit BigEndianWordReader(std::span<std::uint8_t const> bytes)
+            : _bytes{bytes}
+        {}
+
+        std::uint16_t read()
+        {
+            if ((_offset + sizeof(std::uint16_t)) > _bytes.size())
```
KimonHoffmann: Can in theory overflow, maybe: `(_bytes.size() - _offset) < sizeof(std::uint16_t)`.

tools/mxl-data-probe/main.cpp
```diff
@@ -0,0 +1,467 @@
...
+            {
+                throw std::runtime_error{"Out-of-bounds read while parsing RFC-8331 ANC payload."};
+            }
+
+            auto const value = static_cast<std::uint16_t>((static_cast<std::uint16_t>(_bytes[_offset]) << 8U) | _bytes[_offset + 1]);
+            _offset += sizeof(std::uint16_t);
+            return value;
+        }
+
+        std::size_t wordOffset() const noexcept
```
KimonHoffmann: `constexpr` + `[[nodiscard]]`

tools/mxl-data-probe/main.cpp
```diff
@@ -0,0 +1,467 @@
...
+    };
+
+    class BigEndianWordReader
+    {
+    public:
+        explicit BigEndianWordReader(std::span<std::uint8_t const> bytes)
+            : _bytes{bytes}
+        {}
+
+        std::uint16_t read()
```
KimonHoffmann: `constexpr`!?

tools/mxl-data-probe/main.cpp
```diff
@@ -0,0 +1,467 @@
...
+    private:
+        ::mxlInstance _instance = nullptr;
+        ::mxlFlowReader _reader = nullptr;
+    };
+
+    class BigEndianWordReader
+    {
+    public:
+        explicit BigEndianWordReader(std::span<std::uint8_t const> bytes)
+            : _bytes{bytes}
```
KimonHoffmann, line 126: Please avoid the split init in favor of `, offset{0}`.

tools/mxl-data-probe/main.cpp
```diff
@@ -0,0 +1,467 @@
...
+            _offset += sizeof(std::uint16_t);
+            return value;
+        }
+
+        std::size_t wordOffset() const noexcept
+        {
+            return _offset / sizeof(std::uint16_t);
+        }
+
+        std::size_t byteOffset() const noexcept
```
KimonHoffmann: `constexpr` + `[[nodiscard]]`

tools/mxl-data-probe/main.cpp
```diff
@@ -0,0 +1,467 @@
...
+    private:
+        std::span<std::uint8_t const> _bytes;
+        std::size_t _offset = 0;
+    };
+
+    class UdwUnpacker
+    {
+    public:
+        explicit UdwUnpacker(BigEndianWordReader& reader)
+            : _reader{&reader}
```
KimonHoffmann, line 164: Please avoid the split int in favor of:

```
, accumulatedBits{0}
, _bitCount{0}
```

tools/mxl-data-probe/main.cpp
```diff
@@ -0,0 +1,467 @@
...
+        {}
+
+        std::uint16_t read()
```
KimonHoffmann: Maybe `constexpr`!?

tools/mxl-data-probe/main.cpp
```diff
@@ -0,0 +1,467 @@
...
+        frame.ancCount = static_cast<std::uint8_t>(word1 >> 8U);
+        (void)reader.read();
+
+        auto const packetSize = static_cast<std::size_t>(frame.length) + Rfc8331HeaderSizeBytes;
+        if (packetSize > payload.size())
+        {
+            throw std::runtime_error{"RFC-8331 Length field exceeds the available grain payload."};
+        }
+
+        for (std::uint8_t i = 0; i < frame.ancCount; ++i)
```
KimonHoffmann: `auto i = std::uint8_t{0}`

tools/mxl-data-probe/main.cpp
```diff
@@ -0,0 +1,467 @@
...
+            auto const word3 = reader.read();
+
+            element.lineNumber = static_cast<std::uint16_t>((word0 >> 4U) & Lower11Bits);
+            element.did = static_cast<std::uint8_t>((word2 >> 6U) & Lower8Bits);
+            element.sdid = static_cast<std::uint8_t>(((word2 & 0x000fU) << 4U) | ((word3 >> 12U) & 0x000fU));
+            element.dataCount = static_cast<std::uint8_t>((word3 >> 2U) & Lower8Bits);
+            element.udw.reserve(element.dataCount);
+
+            auto unpacker = UdwUnpacker{reader};
+            for (std::uint8_t word = 0; word < element.dataCount; ++word)
```
KimonHoffmann: As above.

tools/mxl-data-probe/main.cpp
```diff
@@ -0,0 +1,467 @@
...
+        {.did = 0x41U, .sdid = 0x0CU, .description = "SMPTE ST 2108-1 HDR/WCG metadata"        },
+        {.did = 0x43U, .sdid = 0x02U, .description = "OP-47 Subtitling Distribution Packet"    },
+        {.did = 0x43U, .sdid = 0x03U, .description = "OP-47 VANC multipacket"                  },
+        {.did = 0x50U, .sdid = 0x01U, .description = "Wide Screen Signaling"                   },
+        {.did = 0x60U, .sdid = 0x60U, .description = "SMPTE ST 12-2 Ancillary Time Code"       },
+        {.did = 0x61U, .sdid = 0x01U, .description = "SMPTE ST 334 Caption Distribution Packet"},
+        {.did = 0x61U, .sdid = 0x02U, .description = "SMPTE ST 334 CEA-608 closed captions"    },
+    };
+
+    char const* describeAncDataType(std::uint8_t did, std::uint8_t sdid) noexcept
```
KimonHoffmann: `constexpr` + `[[nodsicard]]`

tools/mxl-data-probe/main.cpp
```diff
@@ -0,0 +1,467 @@
...
+        fmt::print("  RFC-8331 length: {} bytes\n", frame.length);
+        fmt::print("  ANC count: {}\n", frame.ancCount);
+
+        if (frame.elements.empty())
+        {
+            fmt::print("  No ANC elements\n");
+            return;
+        }
+
+        for (std::size_t i = 0; i < frame.elements.size(); ++i)
```
KimonHoffmann: As above.

KimonHoffmann (changes requested):
Thank you for this very nice and clean contribution!
Only a few small syntactical remarks from my end.

tools/mxl-data-probe/main.cpp
k0zmo: The whole class cannot be instantiated in the `constexpr` context (you need to call `mxlCreateFlowReader` which is not constexpr) so `get()` being `constexpr` doesn't buy us anything

vt-tv (changes requested):

docs/Tools.md
k0zmo: Done

docs/Tools.md
```diff
@@ -107,6 +107,31 @@ Hint : Live monitoring of a flow details (updates every second)
...
+A tool that opens an MXL ancillary Data flow (`media_type: video/smpte291`), reads one or more grains starting at the current head index, and prints the RFC-8331 ANC elements found in each grain.
+
+```bash
+./mxl-data-probe [OPTIONS]
+
+OPTIONS:
+  -h,     --help              Print this help message and exit
+          --version           Display program version information and exit
+  -d,     --domain TEXT:DIR REQUIRED
```
vt-tv: I guess this is strictly required anymore since we support the mxl URI

k0zmo: That's been corrected in the latest commit

vt-tv (approved):

tools/mxl-data-probe/main.cpp
KimonHoffmann: My stance on this is: If what the method does is a constant expression (and it is not too overboarding to implement it in a header just for the sake of it) I tend to declare it as `constexpr`, because the means to get to an instance may change at some point and that shouldn't be a concern for the method implementation.

tools/CMakeLists.txt
```diff
@@ -1,7 +1,24 @@
 # SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project.
 # SPDX-License-Identifier: Apache-2.0
 
+include(FetchContent)
+
+FetchContent_Declare(
+    simple_uri_parser
+    GIT_REPOSITORY https://github.com/jholloc/simple-uri-parser.git
+    GIT_TAG        v1.0.0
+)
```
KimonHoffmann: Please don't use `FetchContent`, as we settled on vcpkg for that purpose.

KimonHoffmann (changes requested):
Thank you for making all the changes, looks all good to me now!
Only thing left is one thing in the cmake build instructions I didn't catch the first time around.

tools/CMakeLists.txt
k0zmo: We already do in case of mxl-info. But I agree we shouldn't mix 2 package managers and fix both tools. However, simple-uri-parser is not available via vcpkg. I see uriparser or ada-url (there's also boost.url but I suppose that's out of question for someone). Do you have any suggestion which one to take?

KimonHoffmann: Oh, wow, that slipped by me. 😮
I have not used either of the two alternatives myself, but ada-url sounds like a solid alternative.

k0zmo: I ended up using ada-url. I checked uriparser but it's much more low level and tiresome to use correctly. The only downside of ada-url is the binary size - it adds 700KB (without stripping) and 150KB (stripped) for me in gcc-release configuration against the simple-uri-parser.

KimonHoffmann: Thank you for the change, 150KB is surprisingly hefty, but since it doesn't add to the core library I think we can live with it for now.

Would you mind filing a PR for a similar change to `mxl-info` after this?

tools/mxl-data-probe/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+            mxl
+            ada::ada
+            CLI11::CLI11
+            fmt::fmt
+            picojson::picojson
+    )
+
+set_target_properties(mxl-data-probe
+        PROPERTIES
+            INSTALL_RPATH "$ORIGIN/../lib"
```
KimonHoffmann: I know this has been discussed by Vincent and you, but I really don't think going about it this way is the right way to do it. Ideally `GNUInstallDirs` and proper install settings on all targets should take care of this without the need for manual intervention (which would break in any cases, where the lib-dir is nit `lib`, but maybe `lib64` or the likes.

KimonHoffmann (changes requested):

tools/CMakeLists.txt
KimonHoffmann: I see you already too care of `mxl-info`, so never mind my last request. Thanks for that!

tools/mxl-data-probe/CMakeLists.txt
k0zmo: Right now, it doesn't bring any value - as I wrote in #478, `RPATH` doesn't work transitively so you still need `LD_LIBRARY_PATH` or updating `ld.so.conf` file for linker. With a single .so file, this would enable running a tool (whether `mxl-info` or `mxl-data-probe`) without any extra steps. 

What exactly do you mean by 
> `GNUInstallDirs` and proper install settings on all targets should take care of this without the need for manual intervention

As far as I am aware, `GNUInstallDirs` only gives you where LIBDIR or BINDIR are located. It doesn't concern linker setup. For me, the proper install settings would be:

```cmake
set_target_properties(mxl-info
        PROPERTIES
            INSTALL_RPATH "${CMAKE_INSTALL_LIBDIR}"
    )
```

KimonHoffmann: Hmpf ... GitHub ate my comment... Here's a go at repeating what I just wrote:

Yes, exactly that, I meant use `GNUInstallDirs` to set appropriate install locations given a prefix, but was under the false assumption that cmake takes care of setting the `RPATH` accordingly, when linking against a library local to the current project.

Maybe a mixture of both is the best way to go:
```
cmake_path(RELATIVE_PATH CMAKE_INSTALL_FULL_LIBDIR BASE_DIRECTORY "${CMAKE_INSTALL_FULL_BINDIR}"  OUTPUT_VARIABLE MXL_INFO_INSTALL_RPATH)

if (NOT APPLE)
    set_target_properties(mxl-info
            PROPERTIES
                INSTALL_RPATH "$ORIGIN/${MXL_INFO_INSTALL_RPATH}"
        )
else()
    # Same with "@loader_path" maybe!?
endif ()
```

k0zmo: I reworked it and moved the (common) calculation part to tools' `CMakeLists.txt`.

KimonHoffmann (approved):
Looks good, thanks a lot!

PR #480, opened 2026-04-24
Fix A/V sync bug in mxl-gst-sink

pac-work:
- PTS values for video were calculated based on the timestamp of the "next frame in the future", PTS values for audio were calculated based on the actual timestamp of the bufer we just read. This caused A/V sync problems for essentially any signal which was not "synchronous and without latency".
- The actual buffer timestamp is now always used to calculate PTS and the PTS is offsetted automatically to be slightly in the future (by value common to both A and V pipelines), to get fluent playback.
- Moved the logic of presenting the data into the pipeline itself, instead of depending on some time calculations done from the place where we are actually reading (better separation of the MXL related vs. GStreamer related code).
- The application now works even with `read-delay` parameters in order of hundreds of ms and when A/V signals have varying offset (such as outputs from decoders).

### DIFF
--- a/tools/mxl-gst/sink.cpp
+++ b/tools/mxl-gst/sink.cpp
@@ -87,9 +87,30 @@ namespace
             MXL_INFO("Staring pipeline with base time: {} ns", _mxlBaseTime);
         }
 
-        void pushBuffer(GstBuffer* buffer, std::uint64_t now) noexcept
+        // The provided bufferTimestamp will be used to calculate the PTS of the buffer. For video, this should be the time when to show the frame to
+        // the user. For audio, this should be the time of the first sample in the buffer. These values should be based on the MXL origination time of
+        // the data used to construct the buffer.
+        void pushBuffer(GstBuffer* buffer, std::uint64_t bufferTimestampNs) const noexcept
         {
-            GST_BUFFER_PTS(buffer) = now - _mxlBaseTime;
+            // The PTS value is relative to the start of the pipeline (running time).
+            // By definition of how MXL works, all the data read from MXL have origination time in the past. When presenting the data to the user, we
+            // have to be presenting them slightly in the future. We adjust the PTS dynamically to make sure the playback stays smooth, but also that
+            // we present the data with the lowest possible latency.
+
+            constexpr auto const MIN_PTS_OFFSET_NS = std::uint64_t{1'000'000}; // We want all the data to be presented 1 ms in the future.
+            auto const runningTime = gst_element_get_current_running_time(_pipeline);
+            auto currentPtsOffset = std::uint64_t{_ptsOffset};
+            auto pts = bufferTimestampNs - _mxlBaseTime + currentPtsOffset;
+            if (_autoAdjustPtsOffset && pts < runningTime + MIN_PTS_OFFSET_NS)
+            {
+                auto const newPtsOffset = runningTime + MIN_PTS_OFFSET_NS + _mxlBaseTime - bufferTimestampNs;
+                if (_ptsOffset.compare_exchange_strong(currentPtsOffset, newPtsOffset))
+                {
+                    MXL_INFO("Pipeline(s) PTS offset adjusted to {} ns.", newPtsOffset);
+                    pts = bufferTimestampNs - _mxlBaseTime + newPtsOffset;
+                }
+            }
+            GST_BUFFER_PTS(buffer) = pts;
 
             int ret;
             ::g_signal_emit_by_name(_appSource, "push-buffer", buffer, &ret);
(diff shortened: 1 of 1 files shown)


tools/mxl-gst/sink.cpp
```diff
@@ -87,9 +87,30 @@ namespace
...
+        // the data used to construct the buffer.
+        void pushBuffer(GstBuffer* buffer, std::uint64_t bufferTimestampNs) const noexcept
         {
-            GST_BUFFER_PTS(buffer) = now - _mxlBaseTime;
+            // The PTS value is relative to the start of the pipeline (running time).
+            // By definition of how MXL works, all the data read from MXL have origination time in the past. When presenting the data to the user, we
+            // have to be presenting them slightly in the future. We adjust the PTS dynamically to make sure the playback stays smooth, but also that
+            // we present the data with the lowest possible latency.
+
+            constexpr std::uint64_t MIN_PTS_OFFSET_NS = 1'000'000; // We want all the data to be presented 1 ms in the future.
```
KimonHoffmann: `constexpr auto const ... = std::uint64_t{...};`

tools/mxl-gst/sink.cpp
```diff
@@ -87,9 +87,30 @@ namespace
...
+            auto const runningTime = gst_element_get_current_running_time(_pipeline);
+            std::uint64_t currentPtsOffset = _ptsOffset;
```
KimonHoffmann: `auto`

tools/mxl-gst/sink.cpp
```diff
@@ -87,9 +87,30 @@ namespace
...
+            auto pts = bufferTimestampNs - _mxlBaseTime + currentPtsOffset;
+            if (pts < runningTime + MIN_PTS_OFFSET_NS)
+            {
+                std::uint64_t const newPtsOffset = runningTime + MIN_PTS_OFFSET_NS + _mxlBaseTime - bufferTimestampNs;
```
KimonHoffmann: `auto const`

tools/mxl-gst/sink.cpp
```diff
@@ -178,6 +200,8 @@ namespace
         GstElement* _pipeline;
         GstElement* _appSource;
         std::uint64_t _mxlBaseTime;
+        // Has to be shared by all the pipelines, to make sure that we do not introduce audio / video offset.
+        inline static std::atomic_uint64_t _ptsOffset = 0;
```
KimonHoffmann: Please declare and initialize separately in case of a class-scope static variable. Also prefer the primary template `std::atomic<std::uint64_t>`, as this makes it explicit that the underlying type is exactly `std::uint64_t`.

KimonHoffmann (changes requested):
Thank you for this!
Could we make the initial PTS offset configurable via the command-line arguments, because there might be situations where users prefer a fixed phase relation even if that might imply a slightly higher latency.

pac-work:
> Thank you for this! Could we make the initial PTS offset configurable via the command-line arguments, because there might be situations where users prefer a fixed phase relation even if that might imply a slightly higher latency.

Sure @KimonHoffmann. Should the value be really just initial, if provided, and then be increased if needed (as it is done now), or should it stay the same and not be auto-adjusted, if provided? I do not mind any of the options, what would you prefer?

tools/mxl-gst/sink.cpp
pac-work: Done

tools/mxl-gst/sink.cpp
pac-work: Done

tools/mxl-gst/sink.cpp
pac-work: Done

tools/mxl-gst/sink.cpp
pac-work: Done

pac-work:
> Sure @KimonHoffmann. Should the value be really just initial, if provided, and then be increased if needed (as it is done now), or should it stay the same and not be auto-adjusted, if provided? I do not mind any of the options, what would you prefer?

For now, I did precisely what you said - the command line parameter just sets the initial value and that gets auto-increased, if it was too small. As stated earlier, no strong opinion on that, can be changed. For me the important thing is the auto-increase when nothing is specified.

vt-tv (approved):
LGTM

KimonHoffmann:
> > Thank you for this! Could we make the initial PTS offset configurable via the command-line arguments, because there might be situations where users prefer a fixed phase relation even if that might imply a slightly higher latency.
> 
> Sure @KimonHoffmann. Should the value be really just initial, if provided, and then be increased if needed (as it is done now), or should it stay the same and not be auto-adjusted, if provided? I do not mind any of the options, what would you prefer?

Good question! Just providing an initial value is the straightforward thing to do, but I think having a fixed value that does not change in either direction is the more valuable alternative.

pac-work:
> Good question! Just providing an initial value is the straightforward thing to do, but I think having a fixed value that does not change in either direction is the more valuable alternative.

Done. Use of the parameter will make the value fixed for the runtime of the program. Ready for reviewing, @KimonHoffmann.

KimonHoffmann (approved):
Looks good, thank you!

PR #486, opened 2026-04-30
Fix the status code returned, when clients pass an invalid reader to mxlFlowReaderGetSamples or mxlFlowReaderGetSamplesNonBlocking.

KimonHoffmann:
They both returned `MXL_ERR_INVALID_FLOW_WRITER`, when `MXL_ERR_INVALID_FLOW_READER` would have been correct.


jonasohland (approved):
🥲

vt-tv (approved):
good catch!

PR #487, opened 2026-04-30
Add functions to query the maximum number of samples that can be read or written to a continuous flow in a single transaction.

KimonHoffmann:
As discussed before these should help avoid confusion how many samples may be consumed or provided in one go.

### DIFF
--- a/lib/include/mxl/flow.h
+++ b/lib/include/mxl/flow.h
@@ -393,14 +393,29 @@ extern "C"
     mxlStatus mxlFlowWriterCancelGrain(mxlFlowWriter writer);
 
     /**
-     * Inform mxl that a user is done writing the grain that was previously opened.  This will in turn signal all readers waiting on the ringbuffer
-     * that a new grain is available. The mxlGrainInfo flags field in shared memory will be updated based on grain->flags This will increase the head
-     * and potentially the tail IF this grain is the new head.
+     * Inform mxl that a user is done writing the grain that was previously opened.
+     * This will in turn signal all readers waiting on the ringbuffer that a new grain is available. The mxlGrainInfo
+     * flags field in shared memory will be updated based on grain->flags This will increase the head and potentially
+     * the tail IF this grain is the new head.
      *
      * \return The result code. \see mxlStatus
      */
     MXL_EXPORT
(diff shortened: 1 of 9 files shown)


vt-tv (approved):
LGTM

lib/include/mxl/flow.h
```diff
@@ -393,15 +393,30 @@ extern "C"
...
+     * \param[in] reader A valid flow reader
+     * \param[out] maxReadLength A valid pointer referring to the location, in which to store
+     *      the maximum number of samples a read operation may retrieve from the flow \p reader operates on.
+     * \return The result code. \see mxlStatus
+     * \note Please note that a read of the specified length is only guaranteed to be possible, if
+     *      1. enough history is available, and
+     *      2. the read is performed at the current head index.
+     */
+    MXL_EXPORT
+    mxlStatus mxlFlowReaderGetMaxReadLength(mxlFlowReader reader, size_t* maxReadLength);
```
jonasohland: I think the function name should include a hint that it only applies to cont. flows.
Like `mxlContinuousFlowReaderGetMaxReadLength()`
or `mxlFlowReaderGetMaxReadLengthSamples()`
or `mxlFlowReaderGetMaxSamplesToRead()`

jonasohland (changes requested):

lib/include/mxl/flow.h
KimonHoffmann: Yes, that sounds reasonable. I would tend to pick the first alternative, does anyone else want to weigh in on this?

pedro-alves-ferreira: I like the first one too.

pedro-alves-ferreira (approved):
LGTM. Thanks

lib/include/mxl/flow.h
KimonHoffmann: After some more consideration I would rather pick the second alternative. The reason is that we currently do not expose the kind of reader/writer explicitly by name anywhere in the C API and I think it is sensible not to break with this practice just for these small ancillary functions.

jonasohland: I agree, and would prefer 2.

jonasohland (approved):

PR #498, opened 2026-05-01
Fix potential fabrics memory leak + fix blocking write API or writes where timeout < 100ms

jonasohland:
Fix a memory leak that would happen if someone would re-configure a fabrics initiator object by calling the setup function again.

Fix the blocking write function for initiators. This function would end up not doing internal state maintenance when the timeout for the blocking function call would be less than the internal minimum event queue poll interval.


garethsb (approved):
LGTM

vt-tv (approved):
LGTM

PR #499, opened 2026-05-01
Enable blocking API for EFA provider on libfabric >2.5

jonasohland:
Support for blocking completion queue access was added very recently to libfabric. This is what enables the very low CPU impact that we saw during the NAB interop on AWS.

This PR enables use of the blocking API in the EFA provider. It also adds a warning that is logged when using the EFA provider on a libfabric version that does not support the blocking API.


vt-tv (approved):
LGTM

mlefebvre1 (approved):

PR #501, opened 2026-05-04
Fix A/V sync bug in mxl-gst-sink (manual backport for 1.0)

pac-work:
- PTS values for video were calculated based on the timestamp of the "next frame in the future", PTS values for audio were calculated based on the actual timestamp of the bufer we just read. This caused A/V sync problems for essentially any signal which was not "synchronous and without latency".
- The actual buffer timestamp is now always used to calculate PTS and the PTS is offsetted to be slightly in the future (by value common to both A and V pipelines), to get fluent playback.
- Moved the logic of presenting the data into the pipeline itself, instead of depending on some time calculations done from the place where we are actually reading (better separation of the MXL related vs. GStreamer related code).
- This is dummed down version of existing commit from main (b3a683a2260a4b7121ab0a111baba186d8675ef7). It has been decided to not port that commit as a whole, because it introduced new command line parameter. This commit only fixes the immediate problem of A/V being out of sync, but things will still keep failing with large enough latencies / latency differences between A and V.

### DIFF
--- a/tools/mxl-gst/sink.cpp
+++ b/tools/mxl-gst/sink.cpp
@@ -87,9 +87,21 @@ namespace
             MXL_INFO("Staring pipeline with base time: {} ns", _mxlBaseTime);
         }
 
-        void pushBuffer(GstBuffer* buffer, std::uint64_t now) noexcept
+        // The provided bufferTimestamp will be used to calculate the PTS of the buffer. For video, this should be the time when to show the frame to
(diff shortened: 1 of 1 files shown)


vt-tv (approved):
LGTM

KimonHoffmann (approved):
Perfect, thanks for taking the time to separate the fix from the functional change in order to allow for a clean and compatible backport!

PR #504, opened 2026-05-06
Add continuous flow support to mxl-fabrics

### LINKED ISSUE #178: Fabrics: Continuous flow support
Implement support for continuous flows in the fabrics API. Blocked by: #168 and #12.

### LINKED ISSUE #181: Fabrics: PR Step 2: API implementation

mlefebvre1:
This pull request introduces support for the "continuous flows" to Fabrics.

This will close the following issues:  https://github.com/dmf-mxl/mxl/issues/178 and https://github.com/dmf-mxl/mxl/issues/181.

To visualize what was implemented https://github.com/dmf-mxl/mxl/blob/main/docs/fabrics/img/mxl-fabrics-cont-flow-rma.png

### DIFF
--- a/lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
+++ b/lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
@@ -1,0 +1,109 @@
+// SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "AudioBounceBuffer.hpp"
+#include <cstddef>
+#include <cstdint>
+
+namespace mxl::lib::fabrics::ofi
+{
+
+    // TODO: could be shared with PosixContinuousFlowReader
+    void AudioBounceBuffer::getMutableMultiBufferSlices(std::uint64_t index, std::size_t count, size_t bufferLength, std::size_t sampleWordSize,
+        std::size_t channelCount, std::uint8_t* baseBufferPtr, mxlMutableWrappedMultiBufferSlice& slice) noexcept
+    {
+        auto const startOffset = (index + bufferLength - count) % bufferLength;
+        auto const endOffset = (index % bufferLength);
+
+        auto const firstLength = (startOffset < endOffset) ? count : bufferLength - startOffset;
+        auto const secondLength = count - firstLength;
+
+        slice.base.fragments[0].pointer = baseBufferPtr + (sampleWordSize * startOffset);
+        slice.base.fragments[0].size = sampleWordSize * firstLength;
+
+        slice.base.fragments[1].pointer = baseBufferPtr;
+        slice.base.fragments[1].size = sampleWordSize * secondLength;
+
+        slice.stride = sampleWordSize * bufferLength;
+        slice.count = channelCount;
+    }
+
+    AudioBounceBufferEntry::AudioBounceBufferEntry(std::uint32_t size)
+        : _data(sizeof(AudioEntryHeader) + size)
+    {}
+
+    AudioEntryHeader const* AudioBounceBufferEntry::header() const noexcept
+    {
+        // The header is located at the beginning of the buffer, so we can reinterpret the start of the data as an AudioEntryHeader
+        return std::bit_cast<AudioEntryHeader const*>(_data.data());
+    }
+
+    std::uint8_t const* AudioBounceBufferEntry::samples() const noexcept
+    {
+        return _data.data() + sizeof(AudioEntryHeader); // Offset the pointer by the size of the header to get to the samples
+    }
+
+    std::uint8_t const* AudioBounceBufferEntry::data() const noexcept
+    {
+        return _data.data();
+    }
+
+    std::size_t AudioBounceBufferEntry::size() const noexcept
+    {
+        return _data.size() + sizeof(AudioEntryHeader);
+    }
+
+    AudioBounceBuffer::AudioBounceBuffer(size_t entryCount, std::uint32_t entrySize, DataLayout::AudioDataLayout layout)
+        : _entries(entryCount, AudioBounceBufferEntry(entrySize))
+        , _layout(layout)
+    {}
+
+    std::vector<Region> AudioBounceBuffer::getRegions() const noexcept
+    {
+        std::vector<Region> out;
+        out.reserve(_entries.size());
+        for (auto const& entry : _entries)
+        {
+            out.emplace_back(reinterpret_cast<std::uintptr_t>(entry.data()), entry.size(), nullptr, nullptr, Region::Location::host());
+        }
+        return out;
+    }
+
+    size_t AudioBounceBuffer::nbEntries() const noexcept
+    {
+        return _entries.size();
+    }
+
+    AudioEntryHeader const& AudioBounceBuffer::unpack(std::size_t entryIndex, Region& outRegion) const noexcept
+    {
+        auto const& entry = _entries.at(entryIndex);
+        auto header = entry.header();
+
+        // Using the given audio data layout, head index and number of samples recover the slices
+        mxlMutableWrappedMultiBufferSlice slices;
+        getMutableMultiBufferSlices(header->headIndex,
+            header->count,
+            _layout.bufferLength,
+            _layout.sampleSize,
+            _layout.channelCount,
+            reinterpret_cast<std::uint8_t*>(outRegion.base), // NOLINT
+            slices);
+
+        auto srcAddr = entry.samples();
+        for (auto& fragment : slices.base.fragments)
+        {
+            // check if the fragment present
+            if (fragment.size > 0)
+            {
+                for (size_t chan = 0; chan < slices.count; chan++)
+                {
+                    auto dstAddr = reinterpret_cast<std::uintptr_t>(fragment.pointer) + (slices.stride * chan);
+                    std::memcpy(reinterpret_cast<void*>(dstAddr), srcAddr, fragment.size); // NOLINT
+                    srcAddr += fragment.size;
+                }
+            }
+        }
+
+        return *header;
+    }
+}
--- a/lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
+++ b/lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
@@ -53,7 +53,7 @@ namespace mxl::lib::fabrics::ofi
         return _data.size() + sizeof(AudioEntryHeader);
     }
 
-    AudioBounceBuffer::AudioBounceBuffer(size_t entryCount, std::size_t entrySize, DataLayout::AudioDataLayout layout)
+    AudioBounceBuffer::AudioBounceBuffer(size_t entryCount, std::size_t entrySize, DataLayout::Continuous layout)
         : _entries(entryCount, AudioBounceBufferEntry(entrySize))
         , _layout(layout)
     {}
@@ -69,7 +69,7 @@ namespace mxl::lib::fabrics::ofi
         return out;
     }
 
-    std::size_t AudioBounceBuffer::nbEntries() const noexcept
+    std::size_t AudioBounceBuffer::entryCount() const noexcept
     {
         return _entries.size();
     }
--- a/lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
+++ b/lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
@@ -4,6 +4,8 @@
 #include "AudioBounceBuffer.hpp"
 #include <cstddef>
 #include <cstdint>
+#include <cstring>
+#include <bit>
 
 namespace mxl::lib::fabrics::ofi
 {
--- a/lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
+++ b/lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
@@ -56,8 +56,8 @@ namespace mxl::lib::fabrics::ofi
         return _data.size();
     }
 
-    AudioBounceBuffer::AudioBounceBuffer(size_t entryCount, std::size_t entrySize, DataLayout::Continuous layout)
-        : _entries{entryCount, AudioBounceBufferEntry{static_cast<uint32_t>(entrySize)}}
+    AudioBounceBuffer::AudioBounceBuffer(std::size_t entryCount, std::size_t entrySize, DataLayout::Continuous layout)
+        : _entries{entryCount, AudioBounceBufferEntry{static_cast<std::uint32_t>(entrySize)}}
         , _layout{layout}
     {}
 
--- a/lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
+++ b/lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
@@ -6,6 +6,7 @@
 #include <cstdint>
 #include <cstring>
 #include <bit>
+#include "Exception.hpp"
 
 namespace mxl::lib::fabrics::ofi
 {
@@ -56,13 +57,13 @@ namespace mxl::lib::fabrics::ofi
     }
 
     AudioBounceBuffer::AudioBounceBuffer(size_t entryCount, std::size_t entrySize, DataLayout::Continuous layout)
-        : _entries(entryCount, AudioBounceBufferEntry(entrySize))
-        , _layout(layout)
+        : _entries{entryCount, AudioBounceBufferEntry{static_cast<uint32_t>(entrySize)}}
+        , _layout{layout}
     {}
 
     std::vector<Region> AudioBounceBuffer::getRegions() const noexcept
     {
-        std::vector<Region> out;
+        auto out = std::vector<Region>{};
         out.reserve(_entries.size());
         for (auto const& entry : _entries)
         {
@@ -81,10 +82,18 @@ namespace mxl::lib::fabrics::ofi
         return _entries.empty() ? 0 : _entries.front().size();
     }
 
-    AudioEntryHeader const& AudioBounceBuffer::unpack(std::size_t entryIndex, Region& outRegion) const
+    AudioEntryHeader const& AudioBounceBuffer::unpack(std::size_t entryIndex, Region const& outRegion) const
     {
         auto const& entry = _entries.at(entryIndex);
         auto const* header = entry.header();
+
+        auto const maxCountPerEntry = (entrySize() - sizeof(AudioEntryHeader)) / (_layout.channelCount * _layout.sampleSize);
+
+        if (header->count > maxCountPerEntry)
+        {
+            throw Exception::invalidArgument(
+                "Invalid 'count' {} received in the header. That number of samples per channel would bust the bounce buffer entry.", header->count);
+        }
 
         // Using the given audio data layout, head index and number of samples recover the slices
         mxlMutableWrappedMultiBufferSlice slices;
@@ -102,10 +111,10 @@ namespace mxl::lib::fabrics::ofi
             // check if the fragment present
             if (fragment.size > 0)
             {
-                for (std::size_t chan = 0; chan < slices.count; chan++)
+                for (auto chan = std::size_t{0}; chan < slices.count; chan++)
                 {
-                    auto dstAddr = reinterpret_cast<std::uintptr_t>(fragment.pointer) + (slices.stride * chan);
-                    std::memcpy(reinterpret_cast<void*>(dstAddr), srcAddr, fragment.size); // NOLINT
+                    auto const dstAddr = reinterpret_cast<std::uint8_t*>(fragment.pointer) + (slices.stride * chan);
+                    std::memcpy(dstAddr, srcAddr, fragment.size);
                     srcAddr += fragment.size;
                 }
             }
--- a/lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
+++ b/lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
@@ -9,7 +9,7 @@ namespace mxl::lib::fabrics::ofi
 {
 
     // TODO: could be shared with PosixContinuousFlowReader
-    void AudioBounceBuffer::getMutableMultiBufferSlices(std::uint64_t index, std::size_t count, size_t bufferLength, std::size_t sampleWordSize,
+    void AudioBounceBuffer::getMutableMultiBufferSlices(std::uint64_t index, std::size_t count, std::size_t bufferLength, std::size_t sampleWordSize,
         std::size_t channelCount, std::uint8_t* baseBufferPtr, mxlMutableWrappedMultiBufferSlice& slice) noexcept
     {
         auto const startOffset = (index + bufferLength - count) % bufferLength;
@@ -53,7 +53,7 @@ namespace mxl::lib::fabrics::ofi
         return _data.size() + sizeof(AudioEntryHeader);
     }
 
-    AudioBounceBuffer::AudioBounceBuffer(size_t entryCount, std::uint32_t entrySize, DataLayout::AudioDataLayout layout)
+    AudioBounceBuffer::AudioBounceBuffer(size_t entryCount, std::size_t entrySize, DataLayout::AudioDataLayout layout)
         : _entries(entryCount, AudioBounceBufferEntry(entrySize))
         , _layout(layout)
     {}
@@ -69,15 +69,20 @@ namespace mxl::lib::fabrics::ofi
         return out;
     }
 
-    size_t AudioBounceBuffer::nbEntries() const noexcept
+    std::size_t AudioBounceBuffer::nbEntries() const noexcept
     {
         return _entries.size();
+    }
+
+    std::size_t AudioBounceBuffer::entrySize() const noexcept
+    {
+        return _entries.empty() ? 0 : _entries.front().size();
     }
 
     AudioEntryHeader const& AudioBounceBuffer::unpack(std::size_t entryIndex, Region& outRegion) const noexcept
     {
         auto const& entry = _entries.at(entryIndex);
-        auto header = entry.header();
+        auto const* header = entry.header();
 
         // Using the given audio data layout, head index and number of samples recover the slices
         mxlMutableWrappedMultiBufferSlice slices;
@@ -95,7 +100,7 @@ namespace mxl::lib::fabrics::ofi
             // check if the fragment present
             if (fragment.size > 0)
             {
-                for (size_t chan = 0; chan < slices.count; chan++)
+                for (std::size_t chan = 0; chan < slices.count; chan++)
                 {
                     auto dstAddr = reinterpret_cast<std::uintptr_t>(fragment.pointer) + (slices.stride * chan);
                     std::memcpy(reinterpret_cast<void*>(dstAddr), srcAddr, fragment.size); // NOLINT
--- a/lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
+++ b/lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
@@ -79,7 +79,7 @@ namespace mxl::lib::fabrics::ofi
         return _entries.empty() ? 0 : _entries.front().size();
     }
 
-    AudioEntryHeader const& AudioBounceBuffer::unpack(std::size_t entryIndex, Region& outRegion) const noexcept
+    AudioEntryHeader const& AudioBounceBuffer::unpack(std::size_t entryIndex, Region& outRegion) const
     {
         auto const& entry = _entries.at(entryIndex);
         auto const* header = entry.header();
--- a/lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
+++ b/lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
@@ -50,7 +50,7 @@ namespace mxl::lib::fabrics::ofi
 
     std::size_t AudioBounceBufferEntry::size() const noexcept
     {
-        return _data.size() + sizeof(AudioEntryHeader);
+        return _data.size();
     }
 
     AudioBounceBuffer::AudioBounceBuffer(size_t entryCount, std::size_t entrySize, DataLayout::Continuous layout)
--- a/lib/fabrics/ofi/src/internal/AudioBounceBuffer.hpp
+++ b/lib/fabrics/ofi/src/internal/AudioBounceBuffer.hpp
@@ -63,7 +63,7 @@ namespace mxl::lib::fabrics::ofi
     class AudioBounceBuffer
     {
     public:
-        AudioBounceBuffer(size_t entryCount, std::size_t entrySize, DataLayout::Continuous layout);
+        AudioBounceBuffer(std::size_t entryCount, std::size_t entrySize, DataLayout::Continuous layout);
 
         /** \brief Return the regions corresponding to the bounce buffer entries. Each region corresponds to one bounce buffer entry. The regions are
          * returned in the same order as the bounce buffer entries. */
@@ -85,7 +85,7 @@ namespace mxl::lib::fabrics::ofi
          */
         AudioEntryHeader const& unpack(std::size_t entryIndex, Region const& outRegion) const;
 
-        static void getMutableMultiBufferSlices(std::uint64_t index, std::size_t count, size_t bufferLength, std::size_t sampleWordSize,
+        static void getMutableMultiBufferSlices(std::uint64_t index, std::size_t count, std::size_t bufferLength, std::size_t sampleWordSize,
             std::size_t channelCount, std::uint8_t* baseBufferPtr, mxlMutableWrappedMultiBufferSlice& slice) noexcept;
 
     private:
--- a/lib/fabrics/ofi/src/internal/AudioBounceBuffer.hpp
+++ b/lib/fabrics/ofi/src/internal/AudioBounceBuffer.hpp
@@ -18,33 +18,36 @@ namespace mxl::lib::fabrics::ofi
     };
 
     /**\brief An audio bounce buffer entry is backed by a collection of samples
-     * The layout is:  AudioEntryHeader, followed by samples of ch0, followed by samples of ch1, etc.
+     * The layout is:  AudioEntryHeader, followed by samples.
      */
     class AudioBounceBufferEntry
 
     {
     public:
-        /** Construct an AudioBounceBufferEntry with the given size in bytes. The size should be large enough to hold the samples for all channels.
+        /** \brief Construct an AudioBounceBufferEntry with the given size in bytes. The size should be large enough to hold the samples for all
+         * channels of a given batch size.
          *\note The header size is not included in the size parameter, it will be added on top of it. The total size of the bounce buffer entry will
-         * be header size + size parameter.
+         * be header size + samples size.
          */
         AudioBounceBufferEntry(std::uint32_t size);
 
-        /** Return the header of the bounce buffer entry. The header is located at the beginning of the entry data. */
+        /** \brief Return a view on the header of the bounce buffer entry.
+         */
         [[nodiscard]]
         AudioEntryHeader const* header() const noexcept;
 
-        /** Return a pointer to the samples data of the bounce buffer entry. The samples data is located immediately after the header.
+        /** \brief Return a view to the samples data of the bounce buffer entry.
          */
         [[nodiscard]]
         std::uint8_t const* samples() const noexcept;
 
-        /** Return a pointer to the beginning of the bounce buffer entry data, including the header.
+        /** \brief Return a view to the beginning of the bounce buffer entry data.
          */
         [[nodiscard]]
         std::uint8_t const* data() const noexcept;
 
-        /** The total size of the bounce buffer entry in bytes, including header and samples. */
+        /** \brief The total size of the bounce buffer entry in bytes.
+         */
         [[nodiscard]]
         std::size_t size() const noexcept;
 
@@ -61,19 +64,25 @@ namespace mxl::lib::fabrics::ofi
     class AudioBounceBuffer
     {
     public:
-        AudioBounceBuffer(size_t entryCount, std::uint32_t entrySize, DataLayout::AudioDataLayout layout);
+        AudioBounceBuffer(size_t entryCount, std::size_t entrySize, DataLayout::AudioDataLayout layout);
 
-        /** Return the regions corresponding to the bounce buffer entries. Each region corresponds to one bounce buffer entry. The regions are
+        /** \brief Return the regions corresponding to the bounce buffer entries. Each region corresponds to one bounce buffer entry. The regions are
          * returned in the same order as the bounce buffer entries. */
         [[nodiscard]]
         std::vector<Region> getRegions() const noexcept;
 
-        /** Return the number of entries in the bounce buffer. */
+        /** \brief Return the number of entries in the bounce buffer. */
         [[nodiscard]]
-        size_t nbEntries() const noexcept;
+        std::size_t nbEntries() const noexcept;
 
-        /** Copy the samples from the bounce buffer entry to the provided output region, and return the header of the entry. Internally the header
-         * will be used to correctly map the bounce buffer entry slices to the output region slices.
+        /** \briof Return the size of a bounce buffer entry in bytes.
+         */
+        [[nodiscard]]
+        std::size_t entrySize() const noexcept;
+
+        /** \brief Copy the samples from the bounce buffer entry to the provided output region, and return the header of the entry.
+         *
+         * Internally the header will be used to correctly map the bounce buffer entry slices to the output region slices.
          */
         AudioEntryHeader const& unpack(std::size_t entryIndex, Region& outRegion) const noexcept;
 
--- a/lib/fabrics/ofi/src/internal/AudioBounceBuffer.hpp
+++ b/lib/fabrics/ofi/src/internal/AudioBounceBuffer.hpp
@@ -1,0 +1,87 @@
+// SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project.
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <cstdint>
+#include "Region.hpp"
+
+namespace mxl::lib::fabrics::ofi
+{
+
+    /** \brief Header for an audio bounce buffer entry.
+     */
+    struct AudioEntryHeader
+    {
+        std::uint64_t headIndex;
+        std::size_t count;
+    };
+
+    /**\brief An audio bounce buffer entry is backed by a collection of samples
+     * The layout is:  AudioEntryHeader, followed by samples of ch0, followed by samples of ch1, etc.
+     */
+    class AudioBounceBufferEntry
+
+    {
+    public:
+        /** Construct an AudioBounceBufferEntry with the given size in bytes. The size should be large enough to hold the samples for all channels.
+         *\note The header size is not included in the size parameter, it will be added on top of it. The total size of the bounce buffer entry will
+         * be header size + size parameter.
+         */
+        AudioBounceBufferEntry(std::uint32_t size);
+
+        /** Return the header of the bounce buffer entry. The header is located at the beginning of the entry data. */
+        [[nodiscard]]
+        AudioEntryHeader const* header() const noexcept;
+
+        /** Return a pointer to the samples data of the bounce buffer entry. The samples data is located immediately after the header.
+         */
+        [[nodiscard]]
+        std::uint8_t const* samples() const noexcept;
+
+        /** Return a pointer to the beginning of the bounce buffer entry data, including the header.
+         */
+        [[nodiscard]]
+        std::uint8_t const* data() const noexcept;
+
+        /** The total size of the bounce buffer entry in bytes, including header and samples. */
+        [[nodiscard]]
+        std::size_t size() const noexcept;
+
+    private:
+        std::vector<std::uint8_t> _data;
+    };
+
+    /** \brief A bounce buffer for audio samples.
+     *
+     * For audio samples, a bounce buffer is used to receive the samples from the network before copying them to the final destination region. This is
+     * necessary because the samples to transfer on the Initiator are scattered at multiple locations (one per channel), but only a single memory
+     * address can be provided for the destination.
+     */
+    class AudioBounceBuffer
+    {
+    public:
+        AudioBounceBuffer(size_t entryCount, std::uint32_t entrySize, DataLayout::AudioDataLayout layout);
+
+        /** Return the regions corresponding to the bounce buffer entries. Each region corresponds to one bounce buffer entry. The regions are
+         * returned in the same order as the bounce buffer entries. */
+        [[nodiscard]]
+        std::vector<Region> getRegions() const noexcept;
+
+        /** Return the number of entries in the bounce buffer. */
+        [[nodiscard]]
+        size_t nbEntries() const noexcept;
+
+        /** Copy the samples from the bounce buffer entry to the provided output region, and return the header of the entry. Internally the header
+         * will be used to correctly map the bounce buffer entry slices to the output region slices.
+         */
+        AudioEntryHeader const& unpack(std::size_t entryIndex, Region& outRegion) const noexcept;
+
+        static void getMutableMultiBufferSlices(std::uint64_t index, std::size_t count, size_t bufferLength, std::size_t sampleWordSize,
+            std::size_t channelCount, std::uint8_t* baseBufferPtr, mxlMutableWrappedMultiBufferSlice& slice) noexcept;
+
+    private:
+        std::vector<AudioBounceBufferEntry> _entries;
+        DataLayout::AudioDataLayout _layout;
+    };
+}
--- a/lib/fabrics/ofi/src/internal/AudioBounceBuffer.hpp
+++ b/lib/fabrics/ofi/src/internal/AudioBounceBuffer.hpp
@@ -83,7 +83,7 @@ namespace mxl::lib::fabrics::ofi
          *
          * Internally the header will be used to correctly map the bounce buffer entry slices to the output region slices.
          */
-        AudioEntryHeader const& unpack(std::size_t entryIndex, Region& outRegion) const noexcept;
+        AudioEntryHeader const& unpack(std::size_t entryIndex, Region& outRegion) const;
 
         static void getMutableMultiBufferSlices(std::uint64_t index, std::size_t count, size_t bufferLength, std::size_t sampleWordSize,
             std::size_t channelCount, std::uint8_t* baseBufferPtr, mxlMutableWrappedMultiBufferSlice& slice) noexcept;
--- a/lib/fabrics/ofi/src/internal/AudioBounceBuffer.hpp
+++ b/lib/fabrics/ofi/src/internal/AudioBounceBuffer.hpp
@@ -21,7 +21,6 @@ namespace mxl::lib::fabrics::ofi
      * The layout is:  AudioEntryHeader, followed by samples.
      */
     class AudioBounceBufferEntry
-
     {
     public:
         /** \brief Construct an AudioBounceBufferEntry with the given size in bytes. The size should be large enough to hold the samples for all
@@ -64,7 +63,7 @@ namespace mxl::lib::fabrics::ofi
     class AudioBounceBuffer
     {
     public:
-        AudioBounceBuffer(size_t entryCount, std::size_t entrySize, DataLayout::AudioDataLayout layout);
+        AudioBounceBuffer(size_t entryCount, std::size_t entrySize, DataLayout::Continuous layout);
 
         /** \brief Return the regions corresponding to the bounce buffer entries. Each region corresponds to one bounce buffer entry. The regions are
          * returned in the same order as the bounce buffer entries. */
@@ -73,7 +72,7 @@ namespace mxl::lib::fabrics::ofi
 
         /** \brief Return the number of entries in the bounce buffer. */
         [[nodiscard]]
-        std::size_t nbEntries() const noexcept;
+        std::size_t entryCount() const noexcept;
 
         /** \briof Return the size of a bounce buffer entry in bytes.
          */
@@ -91,6 +90,6 @@ namespace mxl::lib::fabrics::ofi
 
     private:
         std::vector<AudioBounceBufferEntry> _entries;
-        DataLayout::AudioDataLayout _layout;
+        DataLayout::Continuous _layout;
     };
 }
--- a/lib/fabrics/ofi/src/internal/AudioBounceBuffer.hpp
+++ b/lib/fabrics/ofi/src/internal/AudioBounceBuffer.hpp
@@ -28,7 +28,7 @@ namespace mxl::lib::fabrics::ofi
          *\note The header size is not included in the size parameter, it will be added on top of it. The total size of the bounce buffer entry will
          * be header size + samples size.
          */
-        AudioBounceBufferEntry(std::uint32_t size);
+        explicit AudioBounceBufferEntry(std::uint32_t size);
 
         /** \brief Return a view on the header of the bounce buffer entry.
          */
@@ -83,7 +83,7 @@ namespace mxl::lib::fabrics::ofi
          *
          * Internally the header will be used to correctly map the bounce buffer entry slices to the output region slices.
          */
-        AudioEntryHeader const& unpack(std::size_t entryIndex, Region& outRegion) const;
+        AudioEntryHeader const& unpack(std::size_t entryIndex, Region const& outRegion) const;
 
         static void getMutableMultiBufferSlices(std::uint64_t index, std::size_t count, size_t bufferLength, std::size_t sampleWordSize,
             std::size_t channelCount, std::uint8_t* baseBufferPtr, mxlMutableWrappedMultiBufferSlice& slice) noexcept;
--- a/lib/fabrics/ofi/src/internal/AudioBounceBuffer.hpp
+++ b/lib/fabrics/ofi/src/internal/AudioBounceBuffer.hpp
@@ -74,7 +74,7 @@ namespace mxl::lib::fabrics::ofi
         [[nodiscard]]
         std::size_t entryCount() const noexcept;
 
-        /** \briof Return the size of a bounce buffer entry in bytes.
+        /** \brief Return the size of a bounce buffer entry in bytes.
          */
         [[nodiscard]]
         std::size_t entrySize() const noexcept;
--- a/lib/fabrics/ofi/src/internal/DataLayout.hpp
+++ b/lib/fabrics/ofi/src/internal/DataLayout.hpp
@@ -67,13 +67,13 @@ namespace mxl::lib::fabrics::ofi
          * \throws std::bad_variant_access if the DataLayout is not of discrete type.
          */
         [[nodiscard]]
-        Discrete const& asDiscrete() const noexcept;
+        Discrete const& asDiscrete() const;
 
         /** \brief Get the DataLayout as Continuous.
          * \throws std::bad_variant_access if the DataLayout is not of continuous type.
          */
         [[nodiscard]]
-        Continuous const& asContinuous() const noexcept;
+        Continuous const& asContinuous() const;
 
     private:
         using InnerLayout = std::variant<Discrete, Continuous>;
--- a/lib/fabrics/ofi/src/internal/DataLayout.hpp
+++ b/lib/fabrics/ofi/src/internal/DataLayout.hpp
@@ -18,16 +18,16 @@ namespace mxl::lib::fabrics::ofi
     class DataLayout
     {
     public:
-        /** \brief Video layout variant of DataLayout.
+        /** \brief Discrete layout variant of DataLayout.
          */
-        struct VideoDataLayout
+        struct Discrete
         {
             std::array<std::uint32_t, MXL_MAX_PLANES_PER_GRAIN> sliceSizes; /**< Number of slices per plane. \see MXL_MAX_PLANES_PER_GRAIN */
         };
 
-        /** \brief Audio layout variant of DataLayout.
+        /** \brief Continuous layout variant of DataLayout.
          */
-        struct AudioDataLayout
+        struct Continuous
         {
             std::size_t sampleSize;   /**< Size of each audio sample in bytes. */
             std::size_t channelCount; /**< Number of audio channels. */
@@ -40,7 +40,7 @@ namespace mxl::lib::fabrics::ofi
          * \return A DataLayout representing the specified video layout.
          */
         [[nodiscard]]
-        static DataLayout fromVideo(std::array<std::uint32_t, MXL_MAX_PLANES_PER_GRAIN> const& sliceSizes) noexcept; // NOLINT
+        static DataLayout fromDiscrete(std::array<std::uint32_t, MXL_MAX_PLANES_PER_GRAIN> const& sliceSizes) noexcept; // NOLINT
 
         /** \brief Create a DataLayout representing audio data.
          * \param sampleSize The size of each audio sample in bytes.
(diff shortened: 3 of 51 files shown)


lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    AudioBounceBufferEntry::AudioBounceBufferEntry(std::uint32_t size)
+        : _data(sizeof(AudioEntryHeader) + size)
+    {}
+
+    AudioEntryHeader const* AudioBounceBufferEntry::header() const noexcept
+    {
+        // The header is located at the beginning of the buffer, so we can reinterpret the start of the data as an AudioEntryHeader
+        return std::bit_cast<AudioEntryHeader const*>(_data.data());
```
k0zmo, line 41: Should probably add `<bit>` and `<cstring>` for `std::memcpy`

lib/fabrics/ofi/src/internal/LocalRegion.cpp
```diff
@@ -51,4 +52,36 @@ namespace mxl::lib::fabrics::ofi
...
+
+    LocalRegionGroupSpan::LocalRegionGroupSpan(std::span<LocalRegion const> region, std::span<::iovec const> iovec, std::span<void* const> desc)
+        : _inner(region)
+        , _iovec(iovec)
+        , _descs(desc)
+    {}
+
+    std::size_t LocalRegionGroupSpan::byteSize() const noexcept
+    {
+        return std::accumulate(_inner.begin(), _inner.end(), 0, [](size_t sum, auto const region) { return sum + region.len; });
```
k0zmo: Pedantic by it should be `std::accumulate(_inner.begin(), _inner.end(), std::size_t{0}, ...)` to match the accumulator type in lambda

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -308,6 +331,109 @@ class AppInitator
...
+
+            auto previousHeadIndex = headIndex;
+
+            // If we get here, we have transfered the samples, we can work on the next samples.
+            headIndex += batchSize;
+            MXL_DEBUG("Transferred samples headIndex={} count={} nextHeadIndex={}", previousHeadIndex, batchSize, headIndex);
+            mxlSleepForNs(mxlGetNsUntilIndex(headIndex, &configInfo.common.grainRate));
+        }
+
+        // We're done let's remote the target
```
k0zmo: typo: remove

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -308,6 +331,109 @@ class AppInitator
...
+            {
+                status = mxlFabricsInitiatorMakeProgressBlocking(_initiator, 10);
+                if (status == MXL_ERR_INTERRUPTED)
+                {
+                    return MXL_STATUS_OK;
+                }
+
+                if (status != MXL_ERR_NOT_READY && status != MXL_STATUS_OK)
+                {
+                    return status;
```
k0zmo, line 400: Wouldn't this skip the cleanup after the while loop?

lib/tests/fabrics/test_basics.cpp
```diff
@@ -564,10 +565,716 @@ TEST_CASE_PERSISTENT_FIXTURE(mxl::tests::mxlDomainFixture, "Fabrics: Transfer Gr
...
+                mxlFabricsInitiatorMakeProgressBlocking(initiator, std::chrono::milliseconds(20).count());
+                auto status = mxlFabricsTargetReadSamples(target, 20, &dummyIndex, &dummyCount);
+                if (status == MXL_ERR_INTERRUPTED)
+                {
+                    FAIL("Peer disconnected before the transfer completed");
+                }
+                if (status == MXL_STATUS_OK)
+                {
+                    // transfer complete
+                    return;
```
k0zmo: return here will skip the cleanup

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
```diff
@@ -72,12 +86,182 @@ namespace mxl::lib::fabrics::ofi
...
+        {
+            throw Exception::invalidState("Memory already registered.");
+        }
+
+        // Register the audio region provided by the user. When the actual protocol instance is created, the protocol will register additional
+        // regions for the bounce buffer entry headers, but we can only do that once we have the actual protocol instance since the bounce buffer
+        // entry count is a parameter of the protocol instance.
+        domain->registerRegion(_region, FI_WRITE);
+
+        _localRegion = domain->localRegions().front();
```
k0zmo, line 246: In case of multi-targets, wouldn't this be no longer correct? I think we should do similar thing here as with `RMAGrainEgressProtocoLTemplate` and get the size of localRegions before calling `registerRegion`

mlefebvre1: This should and is called exclusively during initiator setup to register the audio region containing the samples that will be transferred to all targets. This is unlike `RMASampleEgressProtocol::registerMemory()` which is called on `addTarget()` and registers a list of `AudioEntryHeader` entries which are different for each target.  

So, what happens during the transfer setup is the following: the scatter-gather list is built of one of those `AudioEntryHeader` buffer and the samples. The `AudioEntryHeader` is different per targets, but the samples are the same for each targets.

tools/mxl-fabrics-demo/demo.cpp
mlefebvre1: Good catch, I've moved the clean-up to the dtor. Also found and fixed a bug in the removeTarget function now that this is called during clean-up.

lib/tests/fabrics/test_basics.cpp
mlefebvre1: I also took the opportunity to clean-up the tests by creating abstractions for waiting for connection, waiting for transfer and waiting for completions.

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
k0zmo: That's reasonable - but the code relies on a "non-local invariant" that I fear might be easy to break later if initiator setup
  starts registering anything else before the egress template. Maybe we could add a small comment:
```
// This template registration runs during initiator setup before any per-target
// header registrations, so the audio sample region is always the first local region.
```
or something from your comment above

mlefebvre1: The invariant is already partially covered by this check: 
```
if (_localRegion)
{
    throw Exception::invalidState("Memory already registered.");
}
```

But, I could add an additional check that `domain->localRegions.size() == 0`

k0zmo: The comment and the first check is enough.

lib/fabrics/ofi/src/internal/AudioBounceBuffer.hpp
```diff
@@ -0,0 +1,95 @@
...
+        /** \brief Return the regions corresponding to the bounce buffer entries. Each region corresponds to one bounce buffer entry. The regions are
+         * returned in the same order as the bounce buffer entries. */
+        [[nodiscard]]
+        std::vector<Region> getRegions() const noexcept;
+
+        /** \brief Return the number of entries in the bounce buffer. */
+        [[nodiscard]]
+        std::size_t entryCount() const noexcept;
+
+        /** \briof Return the size of a bounce buffer entry in bytes.
```
lapointejp:
```suggestion
        /** \brief Return the size of a bounce buffer entry in bytes.
```

lib/fabrics/ofi/src/internal/DataLayout.hpp
```diff
@@ -18,35 +18,65 @@ namespace mxl::lib::fabrics::ofi
...
+        bool isContinuous() const noexcept;
 
-        /** \brief Check if the DataLayout is of video type.
-         * \return true if the DataLayout is of video type, false otherwise.
+        /** \brief Get the DataLayout as Discrete.
+         * \throws std::bad_variant_access if the DataLayout is not of discrete type.
          */
         [[nodiscard]]
-        bool isVideo() const noexcept;
+        Discrete const& asDiscrete() const noexcept;
```
lapointejp: From the comment, this shouldn't be noexcept.

lib/fabrics/ofi/src/internal/DataLayout.hpp
```diff
@@ -18,35 +18,65 @@ namespace mxl::lib::fabrics::ofi
...
 
-        /** \brief Get the DataLayout as a VideoDataLayout.
-         * \throws std::bad_variant_access if the DataLayout is not of video type.
+        /** \brief Get the DataLayout as Continuous.
+         * \throws std::bad_variant_access if the DataLayout is not of continuous type.
          */
         [[nodiscard]]
-        VideoDataLayout const& asVideo() const noexcept;
+        Continuous const& asContinuous() const noexcept;
```
lapointejp: Same as the above on the noexcept.

lib/fabrics/ofi/src/internal/Domain.hpp
```diff
@@ -50,6 +50,12 @@ namespace mxl::lib::fabrics::ofi
          */
         void registerRegions(std::vector<Region> const& regions, std::uint64_t access);
 
+        /** \brief Register a single memory region to this domain.
+         *
+         * The domain will own its own version of the registered memory region.
```
lapointejp, line 55: This comment is made about the Region object, not the underlying buffer, right ?

lib/fabrics/ofi/src/internal/Endpoint.cpp
```diff
@@ -346,14 +344,51 @@ namespace mxl::lib::fabrics::ofi
...
+        auto nbWrites = (localGroup.size() + iovLimit - 1) / iovLimit; // ceil(group.size() / iovLimit)
+
+        for (std::size_t i = 0; i < nbWrites; i++)
+        {
+            auto begin = i * iovLimit;
+            auto end = begin + std::min(iovLimit, localGroup.size() - begin);
+
+            auto localGroupSpan = localGroup.span(begin, end);
+
+            // vadlidate that we don't bust the remote region
```
lapointejp:
```suggestion
            // Validate that we don't bust the remote region.
```

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.hpp
```diff
@@ -14,11 +15,22 @@ namespace mxl::lib::fabrics::ofi
...
+        void registerMemory(std::shared_ptr<Domain> domain) override;
+
         /** \copydoc EgressProtocol::transferGrain()
          */
-        virtual void transferGrain(Endpoint& ep, std::uint64_t localIndex, std::uint64_t remoteIndex, std::uint32_t payloadOffset,
+        void transferGrain(Endpoint& ep, std::uint64_t localIndex, std::uint64_t remoteIndex, std::uint32_t payloadOffset,
             SliceRange const& sliceRange, ::fi_addr_t destAddr = FI_ADDR_UNSPEC) override;
 
+        /** \copydoc EgressProtocol::transferSamples()
+         * \note This is unsupported in this protocol since it is designed for video data. Calling this function will result in an
```
lapointejp:
```suggestion
         * \note This is unsupported in this protocol since it is designed for discrete data. Calling this function will result in an
```

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.hpp
```diff
@@ -28,38 +40,123 @@ namespace mxl::lib::fabrics::ofi
...
-        DataLayout _layout;
+        DataLayout::Discrete _layout;
         std::vector<LocalRegion> _localRegions;
         std::size_t _pending = 0;
     };
 
-    /** \brief Egress protocol for RMA writer endpoint.
-     *
-     * Handles transferring data to remote targets using remote write operations without bounce buffering.
+    /** \brief Template for creating an Egress protocol for RMA writer endpoint to Handle transferring grains to remote targets using remote write
```
lapointejp:
```suggestion
    /** \brief Template for creating an Egress protocol for RMA writer endpoint to handle transferring grains to remote targets using remote write
```

lib/fabrics/ofi/src/internal/ProtocolIngressRMA.hpp
```diff
@@ -37,9 +55,67 @@ namespace mxl::lib::fabrics::ofi
...
     };
 
+    /** \brief Ingress protocol for RMA writer endpoint for audio samples.
+     */
+    class RMASampleIngressProtocol final : public IngressProtocol
+    {
+    public:
+        /** Construct an RMASampleIngressProtocol with the given region and data layout.
+         * \param region The memory region containing audio. The audio samples will be first received in one of the bounce buffer entry and will
+   then be copied to this region.
```
lapointejp: Should be indented with the prefixing * ?

lib/fabrics/ofi/src/internal/RDMTarget.cpp
```diff
@@ -84,24 +85,70 @@ namespace mxl::lib::fabrics::ofi
 
     std::optional<Target::GrainReadResult> RDMTarget::readGrain()
     {
-        return readNextGrain<QueueReadMode::NonBlocking>({});
+        if (!_protocol->canReadGrains())
+        {
+            throw Exception::invalidState("This target does not support reading grains.");
```
lapointejp: There is an inconsistency between this throw and those from the next few functions.
I'm relatively happy with saying target given you're fundamentally calling the "RDMTarget" for this function.

However I would keep the state message consistent between this one and the next couple similar ones.

lib/fabrics/ofi/src/internal/Target.hpp
```diff
@@ -36,10 +45,23 @@ namespace mxl::lib::fabrics::ofi
 
         /** \brief Determine if new data can be consumed.
          *
-         * A blocking version of read. see read().
+         * A blocking version of read. see readGrain().
```
lapointejp:
```suggestion
         * A blocking version of readGrain. see readGrain().
```

lib/fabrics/ofi/src/internal/Target.hpp
```diff
@@ -36,10 +45,23 @@ namespace mxl::lib::fabrics::ofi
...
+        /** \brief Determine if new data can be consumed.
+         *
+         * A non-blocking operation that also drives the connection forward. Continuous invocation of this function is necessary for connection
+         * establishment and ongoing progress.
+         */
+        virtual std::optional<SampleReadResult> readSamples() = 0;
+
+        /** \brief Determine if new data can be consumed.
+         *
+         * A blocking version of read. see readSamples().
```
lapointejp:
```suggestion
         * A blocking version of readSamples. see readSamples().
```

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -102,6 +123,7 @@ class AppInitator
     }
 
     mxlStatus setup(std::string const& targetInfoStr)
+
```
lapointejp: Superfluous line return ?

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -197,6 +220,26 @@ class AppInitator
...
+        {
+            MXL_ERROR("Unsupported data format {}", configInfo.common.format);
+            return MXL_ERR_INVALID_ARG;
+        }
+
+        return status;
+    }
+
+    mxlStatus runDiscrete(mxlFlowConfigInfo const& configInfo)
+    { // Extract the FlowInfo structure.
```
lapointejp: Comment on next line ?

lapointejp (changes requested):
Thanks for contributing this. Just a couple nitpicks for you :)

lib/fabrics/ofi/src/internal/Domain.hpp
mlefebvre1: Exactly

lib/fabrics/ofi/src/internal/RDMTarget.cpp
mlefebvre1: Changed it to be consistent with all other instances.

lapointejp (approved):

vt-tv (approved):
Approving on behalf @k0zmo

vt-tv:
Even though we have 2 approvals I will wait for @KimonHoffmann's review before merging.

lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
```diff
@@ -0,0 +1,116 @@
...
+#include <bit>
+
+namespace mxl::lib::fabrics::ofi
+{
+
+    // TODO: could be shared with PosixContinuousFlowReader
+    void AudioBounceBuffer::getMutableMultiBufferSlices(std::uint64_t index, std::size_t count, std::size_t bufferLength, std::size_t sampleWordSize,
+        std::size_t channelCount, std::uint8_t* baseBufferPtr, mxlMutableWrappedMultiBufferSlice& slice) noexcept
+    {
+        auto const startOffset = (index + bufferLength - count) % bufferLength;
```
KimonHoffmann, line 18: Do we need to assert that `count` is less than `bufferLength` or some even smaller value derived from it here?

lib/fabrics/ofi/src/internal/AudioBounceBuffer.hpp
```diff
@@ -0,0 +1,95 @@
...
+     */
+    class AudioBounceBufferEntry
+    {
+    public:
+        /** \brief Construct an AudioBounceBufferEntry with the given size in bytes. The size should be large enough to hold the samples for all
+         * channels of a given batch size.
+         *\note The header size is not included in the size parameter, it will be added on top of it. The total size of the bounce buffer entry will
+         * be header size + samples size.
+         */
+        AudioBounceBufferEntry(std::uint32_t size);
```
KimonHoffmann: Should probably be `explicit`.

lib/fabrics/ofi/src/internal/FabricInfo.hpp
```diff
@@ -119,6 +119,10 @@ namespace mxl::lib::fabrics::ofi
          */
         FabricInfo owned() noexcept;
 
+        /** \brief Return the IOV limit for this fabric info, which is the maximum number of iovecs that can be posted in a single send operation.
+         */
+        std::size_t txIovLimit() const noexcept;
```
KimonHoffmann, line 136: Should probably be `[[nodiscard]]` along with most of the other methods, above!?

lib/fabrics/ofi/src/internal/ImmData.hpp
```diff
@@ -52,4 +52,45 @@ namespace mxl::lib::fabrics::ofi
...
+        {
+            std::uint32_t bounceBufferEntry;
+        };
+
+    public:
+        /** \brief Create immediate data from packed data.
+         *
+         * \param data The packed immediate data.
+         */
+        ImmDataSample(std::uint32_t data) noexcept;
```
KimonHoffmann: Should probably be `explicit`.

lib/fabrics/ofi/src/internal/ImmData.hpp
```diff
@@ -52,4 +52,45 @@ namespace mxl::lib::fabrics::ofi
...
+
+        /** \brief Create immediate data from bounce buffer entry.
+         *
+         * \param index The bounce buffer entry.
+         */
+        ImmDataSample(std::uint64_t entry) noexcept;
```
KimonHoffmann: This should probably also be `explicit`.

In addition to this I think it is quite dangerous to have two semantically different constructors that only differ in the bits width of their integral argument.

Maybe `std::in_place_type_t<>` with a tag type could help here?

lib/fabrics/ofi/src/internal/ImmData.cpp
```diff
@@ -28,4 +28,25 @@ namespace mxl::lib::fabrics::ofi
     {
         return _inner;
     }
+
+    ImmDataSample::ImmDataSample(std::uint32_t data) noexcept
+    {
+        _inner = data;
```
KimonHoffmann: Please use the constructor initializer list instead.

lib/fabrics/ofi/src/internal/ImmData.cpp
```diff
@@ -28,4 +28,25 @@ namespace mxl::lib::fabrics::ofi
...
+    }
+
+    ImmDataSample::ImmDataSample(std::uint64_t entry) noexcept
+    {
+        auto bounceBufferEntry = static_cast<std::uint32_t>(entry);
+        _inner = std::bit_cast<std::uint32_t>(Unpacked{.bounceBufferEntry = bounceBufferEntry});
+    }
```
KimonHoffmann: I'm not clear what this constructor achieves ... to me it looks to be an elaborate way of saying:

```cpp
_inner = static_cast<std::uint32_t>(entry);
```

Which could then also be handled in the constructor initializer list.

lib/fabrics/ofi/src/internal/ImmData.cpp
```diff
@@ -28,4 +28,25 @@ namespace mxl::lib::fabrics::ofi
...
+
+    ImmDataSample::Unpacked ImmDataSample::unpack() const noexcept
+    {
+        return std::bit_cast<Unpacked>(_inner);
+    }
+
+    std::uint32_t ImmDataSample::data() const noexcept
+    {
+        return _inner;
+    }
```
KimonHoffmann: Not clear about this one either in conjunction with it's `unpack()` counterpart either.

Regardless of all of that the whole type looks, as if it could be `constexpr` and implemented in the header due to its simplicity.

lib/fabrics/ofi/src/internal/Initiator.hpp
```diff
@@ -42,6 +42,15 @@ namespace mxl::lib::fabrics::ofi
...
         virtual void transferGrain(std::uint64_t grainIndex, std::uint16_t startSlice, std::uint16_t endSlice) = 0;
 
+        /** \brief Transfer samples to all targets.
+         *
+         * This is a non-blocking operation. The transfer is complete only after makeProgress() or makeProgressBlocking() returns false.
+         *
+         * \param headIndex The head index to transfer.
+         * \param count The number of samples per channel to transfer.
+         */
+        virtual void transferSamples(uint64_t headIndex, std::size_t count) = 0;
```
KimonHoffmann: Please use the fully qualified `std::uint64_t`.

KimonHoffmann:
TY!
Review is in progress, but not yet finished I'll see what I can do tonight.

lib/fabrics/ofi/src/internal/Protocol.hpp
```diff
@@ -32,6 +32,8 @@ namespace mxl::lib::fabrics::ofi
         [[nodiscard]]
         virtual std::vector<RemoteRegion> registerMemory(std::shared_ptr<Domain> domain) = 0;
 
+        virtual std::optional<TargetInfoBounceBufferInfo> bounceBufferInfo() = 0;
```
KimonHoffmann: Judging from the method name it seems like this could be `const` qualified?

Also the method should most certainly be annotated `[[nodiscard]]`.

lib/fabrics/ofi/src/internal/Protocol.hpp
```diff
@@ -56,17 +68,32 @@ namespace mxl::lib::fabrics::ofi
...
          */
         virtual void transferGrain(Endpoint& ep, std::uint64_t localIndex, std::uint64_t remoteIndex, std::uint32_t payloadOffset,
             SliceRange const& sliceRange, ::fi_addr_t destAddr = FI_ADDR_UNSPEC) = 0;
 
+        /** \brief Transfer samples to a remote target.
+         * \param headIndex The head index of the samples to transfer.
+         * \param count The number of samples per channel to transfer.
+         * \param destAddr The destination address. This is ignored for connection-oriented endpoints.
+         */
+        virtual void transferSamples(Endpoint& ep, std::uint64_t headIndex, std::size_t count, ::fi_addr_t destAddr = FI_ADDR_UNSPEC) = 0;
```
KimonHoffmann: Why is the `endpoint` in many of the methods of this interface passed as a mutable reference, even though it seems as if it only supplies constant context information?

lib/fabrics/ofi/src/internal/RDMInitiator.hpp
```diff
@@ -127,6 +131,10 @@ namespace mxl::lib::fabrics::ofi
         void transferGrainToTarget(Endpoint::Id targetId, std::uint64_t localIndex, std::uint64_t remoteIndex, std::uint64_t payloadOffset,
             std::uint16_t startSlice, std::uint16_t endSlice) override;
 
+        /** \copydoc Initiator::transferSamples()
+         */
+        void transferSamples(std::uint64_t headIndex, std::size_t count) override;
```
KimonHoffmann: `virtual` + `override` to make identification easier by looking at the left hand side. Also applies to the other overridden methods in this class.

lib/fabrics/ofi/src/internal/ProtocolIngressRMA.hpp
```diff
@@ -22,13 +24,29 @@ namespace mxl::lib::fabrics::ofi
...
+        std::optional<TargetInfoBounceBufferInfo> bounceBufferInfo() override;
+
         /** \copydoc IngressProtocol::start()
          */
         void start(Endpoint& endpoint) override;
 
         /** \copydoc IngressProtocol::processCompletion()
          */
-        std::optional<Target::GrainReadResult> readGrain(Endpoint& endpoint, Completion const& completion) override;
+        std::optional<Target::ReadResult> read(Endpoint& endpoint, Completion const& completion) override;
```
KimonHoffmann: `virtual` + `override` to make identification of overridden methods easier by looking at the aligned left hand side of the method declaration.

Would also apply to the other methods in this class and probably others, as well.

lib/fabrics/ofi/src/internal/RDMTarget.hpp
```diff
@@ -30,10 +30,18 @@ namespace mxl::lib::fabrics::ofi
          */
         std::optional<Target::GrainReadResult> readGrain() final;
 
+        /** \copydoc Target::readSamples()
+         */
+        std::optional<Target::SampleReadResult> readSamples() final;
```
KimonHoffmann: IMHO `final` methods overrides don't really add something here. Shouldn't the class as a whole be marked `final` instead?!

lib/fabrics/ofi/src/internal/Region.hpp
```diff
@@ -202,9 +202,10 @@ namespace mxl::lib::fabrics::ofi
     class MxlRegions
     {
     public:
-        MxlRegions(std::vector<Region> regions, DataLayout dataLayout)
+        MxlRegions(std::vector<Region> regions, DataLayout dataLayout, std::uint32_t maxSyncBatchSize = 0)
             : _regions(std::move(regions))
             , _layout(dataLayout)
+            , _maxSyncBatchSize(maxSyncBatchSize)
```
KimonHoffmann: Curly brace initialization for all members.

lib/fabrics/ofi/src/internal/Region.hpp
```diff
@@ -223,13 +224,17 @@ namespace mxl::lib::fabrics::ofi
...
+        std::uint32_t maxSyncBatchSize() const noexcept;
+
     private:
         friend MxlRegions mxlFabricsRegionsFromFlow(FlowData& flow);
         friend MxlRegions mxlFabricsRegionsFromMutableFlow(FlowData& flow);
 
     private:
         std::vector<Region> _regions;
         DataLayout _layout;
+        std::uint32_t _maxSyncBatchSize{0};
```
KimonHoffmann: Please remove the initializer here, as the field is covered in the one and only constructor and having initialization spread across class declaration and constructor implementation is adding unnecessary mental load.

lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+        slice.base.fragments[1].pointer = baseBufferPtr;
+        slice.base.fragments[1].size = sampleWordSize * secondLength;
+
+        slice.stride = sampleWordSize * bufferLength;
+        slice.count = channelCount;
+    }
+
+    AudioBounceBufferEntry::AudioBounceBufferEntry(std::uint32_t size)
+        : _data(sizeof(AudioEntryHeader) + size)
```
KimonHoffmann, line 35: Please use curly brace initialization.

lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    std::size_t AudioBounceBufferEntry::size() const noexcept
+    {
+        return _data.size();
+    }
+
+    AudioBounceBuffer::AudioBounceBuffer(size_t entryCount, std::size_t entrySize, DataLayout::Continuous layout)
+        : _entries(entryCount, AudioBounceBufferEntry(entrySize))
+        , _layout(layout)
```
KimonHoffmann: Please use curly brace initialization for these, including the nested call to `AudioBounceBufferEntry`s constructor.

lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
```diff
@@ -0,0 +1,116 @@
...
+    {}
+
+    std::vector<Region> AudioBounceBuffer::getRegions() const noexcept
+    {
+        std::vector<Region> out;
```
KimonHoffmann: Could use `auto` + right hand type.

lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
```diff
@@ -0,0 +1,116 @@
...
+            reinterpret_cast<std::uint8_t*>(outRegion.base), // NOLINT
+            slices);
+
+        auto srcAddr = entry.samples();
+        for (auto& fragment : slices.base.fragments)
+        {
+            // check if the fragment present
+            if (fragment.size > 0)
+            {
+                for (std::size_t chan = 0; chan < slices.count; chan++)
```
KimonHoffmann: Could use `auto` + right hand type.

lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
```diff
@@ -0,0 +1,116 @@
...
+                {
+                    auto dstAddr = reinterpret_cast<std::uintptr_t>(fragment.pointer) + (slices.stride * chan);
```
KimonHoffmann: `auto const`?!

KimonHoffmann: Also, since this value is subsequently cast to a void pointer type, it would be more straightforward to do the arithmetic with a `std::uint8_t*`, because then `dstAddr` can be cast implicitly to `void, when passed to `std::memcpy()` below.

KimonHoffmann: ... which would probably allow getting rid of the `// NOLINT` annotation.

lib/fabrics/ofi/src/internal/Endpoint.cpp
```diff
@@ -346,14 +344,51 @@ namespace mxl::lib::fabrics::ofi
...
     {
-        auto rmaIov = remote.toRmaIov();
-        return writeImpl(token, localGroup.asIovec(), localGroup.size(), const_cast<void**>(localGroup.desc()), &rmaIov, destAddr, immData);
+        auto remoteRmaIov = remote.toRmaIov();
+        auto remoteOffset = std::size_t{0};
+
+        auto iovLimit = info().txIovLimit();
+        auto nbWrites = (localGroup.size() + iovLimit - 1) / iovLimit; // ceil(group.size() / iovLimit)
+
+        for (std::size_t i = 0; i < nbWrites; i++)
```
KimonHoffmann: Can use `auto` + right hand type.

lib/fabrics/ofi/src/internal/Endpoint.cpp
```diff
@@ -346,14 +344,51 @@ namespace mxl::lib::fabrics::ofi
...
     std::size_t Endpoint::write(Completion::Token token, LocalRegionGroup const& localGroup, RemoteRegion const& remote, ::fi_addr_t destAddr,
         std::optional<std::uint32_t> immData)
     {
-        auto rmaIov = remote.toRmaIov();
-        return writeImpl(token, localGroup.asIovec(), localGroup.size(), const_cast<void**>(localGroup.desc()), &rmaIov, destAddr, immData);
+        auto remoteRmaIov = remote.toRmaIov();
+        auto remoteOffset = std::size_t{0};
+
+        auto iovLimit = info().txIovLimit();
+        auto nbWrites = (localGroup.size() + iovLimit - 1) / iovLimit; // ceil(group.size() / iovLimit)
```
KimonHoffmann: Can both be `const` if I saw that correctly.

lib/fabrics/ofi/src/internal/Endpoint.cpp
```diff
@@ -346,14 +344,51 @@ namespace mxl::lib::fabrics::ofi
...
+
+        for (std::size_t i = 0; i < nbWrites; i++)
+        {
+            auto begin = i * iovLimit;
+            auto end = begin + std::min(iovLimit, localGroup.size() - begin);
+
+            auto localGroupSpan = localGroup.span(begin, end);
```
KimonHoffmann: Can all three be `const` if I saw that correctly.

lib/fabrics/ofi/src/internal/Endpoint.cpp
```diff
@@ -346,14 +344,51 @@ namespace mxl::lib::fabrics::ofi
...
+            if ((remoteOffset + localGroupSpan.byteSize()) > remote.len)
+            {
+                throw Exception::invalidArgument("Local region group is too large for the remote region.");
+            }
+
+            remoteRmaIov.addr = remote.addr + remoteOffset;
+            remoteRmaIov.len = localGroupSpan.byteSize();
+
+            auto isLastWrite = (i == nbWrites - 1);
+            auto immDataForWrite = isLastWrite ? immData : std::nullopt;
```
KimonHoffmann: Can both be `const`?!

lib/fabrics/ofi/src/internal/FabricInfo.cpp
```diff
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "FabricInfo.hpp"
+#include <cstddef>
```
KimonHoffmann: This should be in the `FabricInfo.hpp` header because this is where `txIovLimit()` is declared.

lib/fabrics/ofi/src/internal/LocalRegion.cpp
```diff
@@ -3,7 +3,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "LocalRegion.hpp"
+#include <cstddef>
```
KimonHoffmann: As above.

lib/fabrics/ofi/src/internal/LocalRegion.cpp
```diff
@@ -51,4 +53,36 @@ namespace mxl::lib::fabrics::ofi
...
     }
+
+    LocalRegionGroupSpan LocalRegionGroup::span(std::size_t begin, std::size_t end) const
+    {
+        if (end < begin)
+        {
+            throw Exception::invalidArgument("end {} is smaller than begin {}", end, begin);
+        }
+
+        auto spanLength = end - begin;
```
KimonHoffmann: Can be `const`.

lib/fabrics/ofi/src/internal/LocalRegion.cpp
```diff
@@ -51,4 +53,36 @@ namespace mxl::lib::fabrics::ofi
...
+        if (spanLength > _inner.size())
+        {
+            throw Exception::invalidArgument(
+                "requested span size {} will be bigger than the actual size of the full vector {}", spanLength, _inner.size());
+        }
+
+        return LocalRegionGroupSpan{
+            std::span(_inner.begin() + begin, spanLength),
+            std::span(_iovs.begin() + begin, spanLength),
+            std::span(_descs.begin() + begin, spanLength),
```
KimonHoffmann: These three should use curly brace initialization, as well!?

lib/fabrics/ofi/src/internal/LocalRegion.hpp
```diff
@@ -119,4 +131,45 @@ namespace mxl::lib::fabrics::ofi
...
+            return _inner.size();
+        }
+
+        /** \brief Returns the sum of each region size in the span.
+         */
+        [[nodiscard]]
+        std::size_t byteSize() const noexcept;
+
+    private:
+        friend class LocalRegionGroup;
```
KimonHoffmann: Friendship declarations have no access modifier, so it should be at the front of the class to make that clear.

```cpp
class What
{
    friend class That;
    public:
        char const* getHelloWorldString() const noexcept;
};
```

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
```diff
@@ -3,27 +3,36 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "ProtocolEgressRMA.hpp"
+#include <cstddef>
```
KimonHoffmann: As before this should be in the header, where the method that needs `std::size_t` is declared.

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
```diff
@@ -3,27 +3,36 @@
...
 {
-    RMAGrainEgressProtocol::RMAGrainEgressProtocol(Completion::Token token, TargetInfo info, DataLayout layout, std::vector<LocalRegion> localRegions)
+
+    RMAGrainEgressProtocol::RMAGrainEgressProtocol(Completion::Token token, TargetInfo info, DataLayout::Discrete layout,
+        std::vector<LocalRegion> localRegions)
         : _token(token)
         , _remoteInfo(std::move(info))
-        , _layout(std::move(layout))
+        , _layout(layout)
         , _localRegions(std::move(localRegions))
```
KimonHoffmann: These should use curly brace initialization.

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
```diff
@@ -3,27 +3,36 @@
...
-        auto layout = _layout.asVideo();
-
-        auto localSize = sliceRange.transferSize(payloadOffset, layout.sliceSizes[0]);
-        auto localOffset = sliceRange.transferOffset(payloadOffset, layout.sliceSizes[0]);
-        auto remoteSize = sliceRange.transferSize(payloadOffset, layout.sliceSizes[0]);
-        auto remoteOffset = sliceRange.transferOffset(payloadOffset, layout.sliceSizes[0]);
+        auto localSize = sliceRange.transferSize(payloadOffset, _layout.sliceSizes[0]);
+        auto localOffset = sliceRange.transferOffset(payloadOffset, _layout.sliceSizes[0]);
+        auto remoteSize = sliceRange.transferSize(payloadOffset, _layout.sliceSizes[0]);
+        auto remoteOffset = sliceRange.transferOffset(payloadOffset, _layout.sliceSizes[0]);
```
KimonHoffmann: These can probably all be `const`!? ... as well as the two below!?

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
```diff
@@ -47,8 +61,8 @@ namespace mxl::lib::fabrics::ofi
         return std::exchange(_pending, 0);
     }
 
-    RMAGrainEgressProtocolTemplate::RMAGrainEgressProtocolTemplate(DataLayout layout, std::vector<Region> regions)
-        : _dataLayout(layout)
+    RMAGrainEgressProtocolTemplate::RMAGrainEgressProtocolTemplate(DataLayout::Discrete layout, std::vector<Region> regions)
+        : _layout(layout)
         , _regions(std::move(regions))
```
KimonHoffmann: Curly brace initializations.

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
```diff
@@ -72,12 +86,187 @@ namespace mxl::lib::fabrics::ofi
 
         struct MakeUniqueEnabler : RMAGrainEgressProtocol
         {
-            MakeUniqueEnabler(Completion::Token token, TargetInfo info, DataLayout layout, std::vector<LocalRegion> localRegion)
+            MakeUniqueEnabler(Completion::Token token, TargetInfo info, DataLayout::Discrete layout, std::vector<LocalRegion> localRegion)
                 : RMAGrainEgressProtocol(token, std::move(info), layout, std::move(localRegion))
```
KimonHoffmann: Should also use curly brace initialization.

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
```diff
@@ -72,12 +86,187 @@ namespace mxl::lib::fabrics::ofi
...
+    }
+
+    RMASampleEgressProtocol::RMASampleEgressProtocol(Completion::Token token, TargetInfo info, DataLayout::Continuous layout, LocalRegion localRegion,
+        std::size_t bounceBufferEntryCount)
+        : _token(token)
+        , _remoteInfo(std::move(info))
+        , _layout(layout)
+        , _localRegion(localRegion)
+        , _entryHeaders(bounceBufferEntryCount)
+        , _bounceBufferEntryCount(bounceBufferEntryCount)
```
KimonHoffmann: Should also use curly brace initialization.

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
```diff
@@ -72,12 +86,187 @@ namespace mxl::lib::fabrics::ofi
...
+                Region::Location::host()); // Host, because the bounce buffer will always be stored on host memory.
+        }
+
+        // The domain is shared with all endpoints, if there's more than 1 target, we can't assume that the first audio header entry for this target
+        // will be at index 1 (right after the audio samples buffer). Read the number of regions already registered to the domain, then
+        // register the entry header regions, and finally get the local regions again to retrieve the actual regions corresponding to the entry
+        // headers we just registered. Use the offset to correctly get the entry header regions in the list of local regions.
+        auto offset = domain->localRegions().size();
+        domain->registerRegions(entryHeaderRegions, FI_WRITE);
+        auto localRegions = domain->localRegions();
```
KimonHoffmann: Can both be `const`?!

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
```diff
@@ -72,12 +86,187 @@ namespace mxl::lib::fabrics::ofi
...
+
+        // 1- Create the scatter-gather list for the transfer and prepend the audio header.
+        auto sgl = makeScatterGatherList(_layout, headIndex, count, _localRegion);
+        // set the header and prepend it to the scatter-gather list.
+        _entryHeaders[_bounceBufferEntryIndex].headIndex = headIndex;
+        _entryHeaders[_bounceBufferEntryIndex].count = count;
+        sgl.insert(sgl.begin(), _entryHeaderRegions[_bounceBufferEntryIndex]);
+
+        // 2- Get the remote region
+        auto remoteRegion = _remoteInfo.remoteRegions[_bounceBufferEntryIndex % _remoteInfo.remoteRegions.size()];
```
KimonHoffmann: Can all be `const` (`entrySizeRequired`, `sgl` & `remoteRegion`)?

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
```diff
@@ -72,12 +86,187 @@ namespace mxl::lib::fabrics::ofi
...
 
+    std::size_t RMASampleEgressProtocol::reset()
+    {
+        return std::exchange(_pending, 0);
+    }
+
+    std::vector<LocalRegion> RMASampleEgressProtocol::makeScatterGatherList(DataLayout::Continuous const& layout, std::uint64_t headIndex,
+        std::size_t count, LocalRegion const& region)
+    {
+        mxlMutableWrappedMultiBufferSlice slice = {};
```
KimonHoffmann: Could use `auto`+ right hand type.

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
```diff
@@ -72,12 +86,187 @@ namespace mxl::lib::fabrics::ofi
...
+        AudioBounceBuffer::getMutableMultiBufferSlices(headIndex,
+            count,
+            layout.bufferLength,
+            layout.sampleSize,
+            layout.channelCount,
+            reinterpret_cast<std::uint8_t*>(region.addr), // NOLINT
+            slice);
+
+        // Double the scatter-gather list length if the second fragment is present.
+        auto sgListLen = slice.base.fragments[1].size > 0 ? 2 * slice.count : slice.count;
```
KimonHoffmann: Braces around the predicate make this more readable IMHO. Also can probably be `const`.

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
```diff
@@ -72,12 +86,187 @@ namespace mxl::lib::fabrics::ofi
...
+        // Create the scatter-gather list using the slices. We create at least one scatter-gather entry per channel. We potentially create an
+        // additional one per channel if 2 fragments are present (wrap-around). When a fragment is not present its size will be 0.
+        auto sgList = std::vector<LocalRegion>{};
+        sgList.reserve(sgListLen);
+        for (auto& fragment : slice.base.fragments)
+        {
+            // check if the fragment present
+            if (fragment.size > 0)
+            {
+                for (size_t chan = 0; chan < slice.count; chan++)
```
KimonHoffmann: Please use fully qualified `std::size_t`, also could use `auto` + right hand type.

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
```diff
@@ -72,12 +86,187 @@ namespace mxl::lib::fabrics::ofi
...
+                {
+                    auto srcAddr = reinterpret_cast<std::uintptr_t>(fragment.pointer) + (slice.stride * chan);
```
KimonHoffmann: Could be `const`.

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
```diff
@@ -72,12 +86,187 @@ namespace mxl::lib::fabrics::ofi
...
+                }
+            }
+        }
+
+        return sgList;
+    }
+
+    RMASampleEgressProtocolTemplate::RMASampleEgressProtocolTemplate(DataLayout::Continuous layout, Region region)
+        : _layout(layout)
+        , _region(region)
```
KimonHoffmann: Curly initialization.

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
```diff
@@ -72,12 +86,187 @@ namespace mxl::lib::fabrics::ofi
...
+        if (!remoteInfo.bounceBufferInfo)
+        {
+            throw Exception::invalidArgument("Remote target does not have bounce buffer info required for sample egress protocol.");
+        }
+
+        struct MakeUniqueEnabler : RMASampleEgressProtocol
+        {
+            MakeUniqueEnabler(Completion::Token token, TargetInfo info, DataLayout::Continuous layout, LocalRegion localRegion,
+                std::uint32_t bounceBufferEntryCount)
+                : RMASampleEgressProtocol(token, std::move(info), layout, localRegion, bounceBufferEntryCount)
```
KimonHoffmann: Curly initialization.

lib/fabrics/ofi/src/internal/ProtocolIngressRMA.cpp
```diff
@@ -3,12 +3,17 @@
...
 #include "Exception.hpp"
 #include "ImmData.hpp"
 #include "Region.hpp"
 
 namespace mxl::lib::fabrics::ofi
 {
+    //
+    // RMAGrainIngressProtocol implementations below
     RMAGrainIngressProtocol::RMAGrainIngressProtocol(std::vector<Region> regions)
         : _regions(std::move(regions))
```
KimonHoffmann: Curly initialization.

lib/fabrics/ofi/src/internal/ProtocolIngressRMA.cpp
```diff
@@ -76,4 +98,106 @@ namespace mxl::lib::fabrics::ofi
         return _immDataBuffer->toLocalRegion();
     }
 
+    //
+    // RMASampleIngressProtocol implementations below
+    RMASampleIngressProtocol::RMASampleIngressProtocol(Region region, DataLayout::Continuous const& layout, std::uint32_t maxSyncBatchSize)
+        : _bounceBuffer(makeAudioBounceBuffer(layout, maxSyncBatchSize))
+        , _region(region)
```
KimonHoffmann: Curly initializaiton.

lib/fabrics/ofi/src/internal/ProtocolIngressRMA.cpp
```diff
@@ -76,4 +98,106 @@ namespace mxl::lib::fabrics::ofi
...
+        domain->registerRegions(_bounceBuffer.getRegions(), FI_REMOTE_WRITE);
+        _isMemoryRegistered = true;
+
+        return domain->remoteRegions();
+    }
+
+    std::optional<TargetInfoBounceBufferInfo> RMASampleIngressProtocol::bounceBufferInfo()
+    {
+        auto entrySize = _bounceBuffer.entrySize(); // All entries have the same size, we can just take the size of the first one
+        auto entryCount = _bounceBuffer.entryCount();
```
KimonHoffmann: Can both be `const`.

lib/fabrics/ofi/src/internal/ProtocolIngressRMA.cpp
```diff
@@ -76,4 +98,106 @@ namespace mxl::lib::fabrics::ofi
...
+
+        auto immData = completionData->data();
+        if (!immData)
+        {
+            throw Exception::invalidState("Received a completion without immediate data.");
+        }
+
+        auto entryIndex = ImmDataSample{static_cast<std::uint32_t>(*immData)}.unpack().bounceBufferEntry;
+
+        auto header = _bounceBuffer.unpack(entryIndex, _region);
```
KimonHoffmann: As it looks to me all local variables can be `const`.

lib/fabrics/ofi/src/internal/ProtocolIngressRMA.cpp
```diff
@@ -76,4 +98,106 @@ namespace mxl::lib::fabrics::ofi
...
+        return _immDataBuffer->toLocalRegion();
+    }
+
+    AudioBounceBuffer RMASampleIngressProtocol::makeAudioBounceBuffer(DataLayout::Continuous const& layout, std::uint32_t maxSyncBatchSize)
+    {
+        auto oneSampleSize = layout.sampleSize * layout.channelCount;
+        auto entrySize = oneSampleSize * maxSyncBatchSize;
+        auto historySize = layout.bufferLength * oneSampleSize;
+
+        auto entryCount = (historySize + entrySize - 1) / entrySize; // ceil(historySize / entrySize)
```
KimonHoffmann: Can all be `const`. Maybe also qualify the literal `1` as `1U`!?

lib/fabrics/ofi/src/internal/RCInitiator.cpp
```diff
@@ -213,6 +219,14 @@ namespace mxl::lib::fabrics::ofi
         }
     }
 
+    void RCInitiatorEndpoint::transferSamples(uint64_t headIndex, size_t count)
```
KimonHoffmann: Please use the fully qualified variants of `uint64_t` and `size_t`.

lib/fabrics/ofi/src/internal/RCInitiator.cpp
```diff
@@ -213,6 +219,14 @@ namespace mxl::lib::fabrics::ofi
...
+    {
+        if (auto state = std::get_if<Connected>(&_state); state != nullptr)
```
KimonHoffmann: `const`?!

lib/fabrics/ofi/src/internal/RDMInitiator.cpp
```diff
@@ -72,12 +72,20 @@ namespace mxl::lib::fabrics::ofi
             std::move(_state));
     }
 
-    void RDMInitiatorTarget::transfer(Endpoint& ep, std::uint64_t localIndex, std::uint64_t remoteIndex, std::uint64_t remoteOffset,
+    void RDMInitiatorTarget::transferGrain(Endpoint& ep, std::uint64_t localIndex, std::uint64_t remoteIndex, std::uint64_t remotePayloadOffset,
         SliceRange const& sliceRange)
     {
         if (auto state = std::get_if<Activated>(&_state); state != nullptr)
```
KimonHoffmann: `const`?!

lib/fabrics/ofi/src/internal/RDMInitiator.cpp
```diff
@@ -72,12 +72,20 @@ namespace mxl::lib::fabrics::ofi
...
         {
-            _proto->transferGrain(ep, localIndex, remoteIndex, remoteOffset, sliceRange, state->fiAddr);
+            _proto->transferGrain(ep, localIndex, remoteIndex, remotePayloadOffset, sliceRange, state->fiAddr);
+        }
+    }
+
+    void RDMInitiatorTarget::transferSamples(Endpoint& ep, std::uint64_t headIndex, std::size_t count)
+    {
+        if (auto state = std::get_if<Activated>(&_state); state != nullptr)
```
KimonHoffmann: `const`?!

lib/fabrics/ofi/src/internal/RDMInitiator.cpp
```diff
@@ -173,13 +181,14 @@ namespace mxl::lib::fabrics::ofi
 
     void RDMInitiator::addTarget(TargetInfo const& targetInfo)
     {
-        if (_remoteEndpoints.find(targetInfo.id) != _remoteEndpoints.end())
+        if (_remoteEndpoints.contains(targetInfo.id))
         {
             throw Exception::exists("A target with endpoint id {} has already been added to this initiator.", targetInfo.id);
         }
 
         auto token = Completion::randomToken();
```
KimonHoffmann: Can probably be `const`!?

lib/fabrics/ofi/src/internal/RDMInitiator.cpp
```diff
@@ -202,15 +211,23 @@ namespace mxl::lib::fabrics::ofi
...
 
     void RDMInitiator::transferGrainToTarget(Endpoint::Id targetId, std::uint64_t localIndex, std::uint64_t remoteIndex, std::uint64_t payloadOffset,
         std::uint16_t startSlice, std::uint16_t endSlice)
     {
         // A completion will be posted to the completion queue per transfer, after which the counter will be decremented again
-        findRemoteByEndpoint(targetId).transfer(_endpoint, localIndex, remoteIndex, payloadOffset, SliceRange::make(startSlice, endSlice));
+        findRemoteByEndpoint(targetId).transferGrain(_endpoint, localIndex, remoteIndex, payloadOffset, SliceRange::make(startSlice, endSlice));
+    }
+
+    void RDMInitiator::transferSamples(uint64_t headIndex, size_t count)
```
KimonHoffmann: Please use the fully qualified versions of the types.

lib/fabrics/ofi/src/internal/Region.cpp
```diff
@@ -159,7 +165,7 @@ namespace mxl::lib::fabrics::ofi
         {
             auto& discreteFlow = static_cast<DiscreteFlowData const&>(flow);
             std::vector<Region> regions;
```
KimonHoffmann: Can use `auto` + right hand type.

lib/fabrics/ofi/src/internal/Region.cpp
```diff
@@ -159,7 +165,7 @@ namespace mxl::lib::fabrics::ofi
...
-
+            regions.reserve(discreteFlow.grainCount());
             for (std::size_t i = 0; i < discreteFlow.grainCount(); ++i)
```
KimonHoffmann: Can use `auto` + right hand type.

lib/fabrics/ofi/src/internal/TargetInfo.cpp
```diff
@@ -39,6 +40,15 @@ namespace mxl::lib::fabrics::ofi
             regions.emplace_back(region);
         }
         root["regions"] = picojson::value(regions);
+
+        if (bounceBufferInfo)
+        {
+            auto bounceBufferInfoObj = picojson::object{};
```
KimonHoffmann, line 46: `const` ?!

lib/fabrics/ofi/src/internal/TargetInfo.cpp
```diff
@@ -76,9 +86,18 @@ namespace mxl::lib::fabrics::ofi
             regions.emplace_back(addr, len, rkey);
         }
 
+        auto bounceBufferInfo = std::optional<TargetInfoBounceBufferInfo>{std::nullopt};
+        if (root.contains("bounceBufferInfo"))
+        {
+            auto bounceBufferInfoObj = root.at("bounceBufferInfo").get<picojson::object>();
+            auto entryCount = std::stoull(bounceBufferInfoObj.at("entryCount").get<std::string>());
+            auto entrySize = std::stoull(bounceBufferInfoObj.at("entrySize").get<std::string>());
```
KimonHoffmann: All three can probably be `const`.

KimonHoffmann (changes requested):
Thanks for this gargantuan effort!
Looks good, except for `ImmDataSample`, which I found confusing and the purpose of which is not really clear to me.
The other comments just syntactic remarks.

lib/fabrics/ofi/src/internal/Protocol.hpp
mlefebvre1: That's a good observation, I will make const references.

lib/fabrics/ofi/src/internal/RDMTarget.hpp
mlefebvre1: Agreed, I will change this for the class

lib/fabrics/ofi/src/internal/RDMInitiator.hpp
mlefebvre1: I'm not sure I like that just for cosmetic reasons, we encode the wrong intent.

lib/fabrics/ofi/src/internal/RDMTarget.hpp
mlefebvre1: Actually we can't, because we need to sub-class `RDMTarget` to enable the creation of a unique pointer if we want to keep the constructor private. So we can't mark it as final... I guess the work around was to mark its virtual methods as final.

```cpp
// In RDMTarget::setup()
struct MakeUniqueEnabler : RDMTarget
{
    MakeUniqueEnabler(Endpoint ep, std::unique_ptr<IngressProtocol> proto)
        : RDMTarget(std::move(ep), std::move(proto))
    {}
};

return {std::make_unique<MakeUniqueEnabler>(std::move(endpoint), std::move(protocol)), std::move(targetInfo)};
```

lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
mlefebvre1: This is a vector, it would not call the same constructor.

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
mlefebvre1: Yes, except `sgl` which `insert` is called on it

lib/fabrics/ofi/src/internal/TargetInfo.cpp
mlefebvre1: We're modifying the object the two lines below.
```c++
bounceBufferInfoObj["entryCount"] = picojson::value(std::to_string(bounceBufferInfo->entryCount));
bounceBufferInfoObj["entrySize"] = picojson::value(std::to_string(bounceBufferInfo->entrySize));
```

lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
mlefebvre1: Yes it could make sense to add it.

On another note, if you look carefully, this function's code is basically the same code as PosixContinuousFlowWritter.cpp:80-96 or PosixContinuousFlowReader.cpp:131-147. So I was wondering should we extract this logic into it's own function that could be shared by these two instances and this instance?

lib/fabrics/ofi/src/internal/ImmData.hpp
mlefebvre1: Actually since for audio only the bounce buffer index is carried through the immediate data, this whole type is useless and I will just remove it.

lib/fabrics/ofi/src/internal/ImmData.hpp
mlefebvre1: I removed the type

lib/fabrics/ofi/src/internal/ImmData.cpp
mlefebvre1: Type removed

lib/fabrics/ofi/src/internal/ImmData.cpp
mlefebvre1: Type removed

lib/fabrics/ofi/src/internal/ImmData.cpp
mlefebvre1: Type removed

lib/fabrics/ofi/src/internal/ProtocolIngressRMA.hpp
mlefebvre1: See my comment above

lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
mlefebvre1: > Do we need to assert that `count` is less than `bufferLength` or some even smaller value derived from it here?

I've added the assertion at the call site (AudioBounceBuffer::unpack), because this is where we receive the count information and we should perform the validation against the bounce buffer entry size.

```cpp
auto const maxCountPerEntry = (entrySize() - sizeof(AudioEntryHeader)) / (_layout.channelCount * _layout.sampleSize);

if (header->count > maxCountPerEntry)
{
    throw Exception::invalidArgument(
        "Invalid 'count' {} received in the header. That number of samples per channel would bust the bounce buffer entry.", header->count);
}
```

mlefebvre1:
> Thanks for this gargantuan effort! Looks good, except for `ImmDataSample`, which I found confusing and the purpose of which is not really clear to me. The other comments just syntactic remarks.

Thank you for this thorough review! I've addressed most of your comments and resolved them.  The others I've not resolved, I've added a question/reply.

Regarding the `ImmDataSample` type, I've decided to simply remove it as it serves no purposes.

lib/fabrics/ofi/src/internal/RDMInitiator.hpp
KimonHoffmann: This is how we do it in other files for the MXL library, and I'm not sure what you mean by wrong intent?
The method **is** virtual and it **does** override, so I don't see any misinformation here.

lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
KimonHoffmann: > On another note, if you look carefully, this function's code is basically the same code as PosixContinuousFlowWritter.cpp:80-96 or PosixContinuousFlowReader.cpp:131-147. So I was wondering should we extract this logic into its own function that could be shared by these two instances and this instance?

Yes, sounds good, feel free to factor it out!
I'm not sure the copies in `PosixContinuousFlow*.cpp` need a similar assertion, since (if I remember correctly) they already check for those conditions in the caller. But since you have now also added the check at the call site the three variants are now the same in that respect.

lib/fabrics/ofi/src/internal/AudioBounceBuffer.hpp
```diff
@@ -0,0 +1,95 @@
...
+    /** \brief A bounce buffer for audio samples.
+     *
+     * For audio samples, a bounce buffer is used to receive the samples from the network before copying them to the final destination region. This is
+     * necessary because the samples to transfer on the Initiator are scattered at multiple locations (one per channel), but only a single memory
+     * address can be provided for the destination.
+     */
+    class AudioBounceBuffer
+    {
+    public:
+        AudioBounceBuffer(size_t entryCount, std::size_t entrySize, DataLayout::Continuous layout);
```
KimonHoffmann: One full qualification I seem to have missed before.

lib/fabrics/ofi/src/internal/AudioBounceBuffer.hpp
```diff
@@ -0,0 +1,95 @@
...
+        [[nodiscard]]
+        std::size_t entrySize() const noexcept;
+
+        /** \brief Copy the samples from the bounce buffer entry to the provided output region, and return the header of the entry.
+         *
+         * Internally the header will be used to correctly map the bounce buffer entry slices to the output region slices.
+         */
+        AudioEntryHeader const& unpack(std::size_t entryIndex, Region const& outRegion) const;
+
+        static void getMutableMultiBufferSlices(std::uint64_t index, std::size_t count, size_t bufferLength, std::size_t sampleWordSize,
```
KimonHoffmann: And another one.

KimonHoffmann (changes requested):
Thanks you for applying all suggested changes so quickly!
I caught two more types missing their qualification.

lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
mlefebvre1: Thinking about this, I will omit this change for this PR, because it is actually more complicated than it looks. Ideally, I would add this method to the `ContinuousFlowData` and would require that the bounce buffer have access to the `FlowData` of the underlying flow. Unfortunately, the current target/initiator setup workflow prevents me from achieving this without much effort.

lib/fabrics/ofi/src/internal/RDMInitiator.hpp
mlefebvre1: Sorry I mistaken the comment here with the one about marking the function as final. Normally, I tend to omit virtual as it is just redundant with override, but if that's the standard used in MXL, I can change it.

On that specific class we should actually mark the functions final instead of override, because they are truly final.

mlefebvre1:
> Thanks you for applying all suggested changes so quickly! I caught two more types missing their qualification.

I've addressed these (and some more that slipped in the previous PR). I've also added the virtual qualifier everywhere it was needed and changed some `override` to `final` qualifiers to indicate that these methods are final when we can't mark the class final even though it is final.

KimonHoffmann (approved):
Looks good, thanks a lot!

PR #505, opened 2026-05-06
mxlsrc: attach MXL flow from negotiate, not start

### LINKED ISSUE #503: [gst-mxl-rs] mxlsrc transition to PLAYING blocks the caller until the flow exists
**Describe the bug**

`mxlsrc.set_state(gst::State::Playing)` blocks until the flow is created, see `init_mxl_reader`.

One option is that `mxlsrc` should model the behaviour of a `filesrc` - error out immediately if file doesn't exist. But it seems more appropriate to transition to `PLAYING` but just not return any buffers until it can open the flow, more like a network receiver source element?

garethsb:
Resolves #503.

**I am not a Rust programmer, so I relied heavily on Cursor help for idiomatic Rust! 😁**

`BaseSrc::start` runs on the state-change thread; blocking on `FlowNotFound` there prevented reaching PLAYING before the flow was created.

- `start()`: only `unlock_stop`; defer reader setup to `negotiate()`.
- `negotiate()`: after validating domain/flow ids, call `init()` when no reader state yet, then resolve caps from the MXL flow definition.
- `init_mxl_reader()`: poll loop checks `clock_wait.flushing` so `unlock()` can interrupt during teardown.

### DIFF
--- a/rust/gst-mxl-rs/src/mxlsrc/imp.rs
+++ b/rust/gst-mxl-rs/src/mxlsrc/imp.rs
@@ -38,9 +38,9 @@ pub(crate) static CAT: LazyLock<gst::DebugCategory> = LazyLock::new(|| {
     gst::DebugCategory::new("mxlsrc", gst::DebugColorFlags::empty(), Some("MXL Source"))
 });
 
-struct ClockWait {
+pub(crate) struct ClockWait {
     clock_id: Option<gst::SingleShotClockId>,
-    flushing: bool,
+    pub(crate) flushing: bool,
 }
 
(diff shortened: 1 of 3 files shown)


pedro-alves-ferreira (approved):
LGTM. Thanks for fixing it.

vt-tv:
@garethsb  should it be backported in 1.0?

pedro-alves-ferreira:
> @garethsb should it be backported in 1.0?

No. We didn't have the Rust-based GStreamer plugins in 1.0.

vt-tv (approved):
LGTM

PR #520, opened 2026-05-21
Fix intermittent failures of st2038_round_trip_via_mxl test

### LINKED ISSUE #519: Intermittent failures of the `st2038_round_trip_via_mxl`
**Describe the bug**
Spurious intermittent failures of the `st2038_round_trip_via_mxl` test in the `gst-mxl-rs` test suite.

**To Reproduce**
Run `cargo test st2038_round_trip_via_mxl` in the `rust` subdirectory of the project

**Expected behavior**
Test should pass

On my system, I get ~12 failures in 50 runs.

The failure occurs because an `.expect()` panics:
```
thread 'st2038_round_trip_via_mxl' (12525) panicked at gst-mxl-rs/tests/data_round_trip.rs:196:14:
appsink produced a sample within 2s
```

jonasohland:
# Details
The `st2038_round_trip_via_mxl` test in the `gst-mxl-rs` test suite is failing both locally and in CI.

Dramatically increasing the number of buffers pushed to the appsrc in this tests fixes the failures for me. It seems like the estimate of 2 buffers that need to be pushed on top of the expected 7 buffers to account for scheduler ordering is too optimistic.

# Backport requirements
gst-mxl-rs is not in v1.0, so no backport necessary

fixes #519

### DIFF
--- a/rust/gst-mxl-rs/tests/data_round_trip.rs
+++ b/rust/gst-mxl-rs/tests/data_round_trip.rs
@@ -181,7 +181,7 @@ fn st2038_round_trip_via_mxl() {
     // Push a couple more buffers than we're going to pull, because the
     // consumer's first MXL read index can be a grain or two after the
     // producer's first MXL write index depending on scheduler ordering.
-    const PUSH_COUNT: usize = PULL_COUNT + 2;
+    const PUSH_COUNT: usize = PULL_COUNT + 64;
     for i in 0..PUSH_COUNT {
         let pts = gst::ClockTime::from_nseconds(i as u64 * GRAIN_PERIOD_NS);
         let bytes = ST2038_TEST_PACKETS[i % ST2038_TEST_PACKETS.len()];
--- a/rust/gst-mxl-rs/tests/data_round_trip.rs
+++ b/rust/gst-mxl-rs/tests/data_round_trip.rs
@@ -117,6 +117,7 @@ impl Drop for TestDomainGuard {
 /// consumer receives exact-byte matches that themselves alternate in the
 /// same order once steady state is reached.
 #[test]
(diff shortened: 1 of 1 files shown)


jonasohland:
@garethsb Can you take a look at this fix?

garethsb:
Hm, I think this test has been run and not failed in CI and doesn't fail in my local environment so I'd like to understand what's going on, especially if it needs to be dramatically increased as you're finding in your env.

jonasohland:
There were failed runs for example in the continuous flow fabrics PR here: https://github.com/dmf-mxl/mxl/actions/runs/25918879744/job/76182392143

Logs have expired unfortunately...

jonasohland:
I did not test any other values than 64, could be that just increasing it by 1-2 also fixes 99% of runs.

jonasohland:
After a few attempts I managed to have it fail on main here: https://github.com/dmf-mxl/mxl/actions/runs/26101036912/job/77152756238
2026-05-21-123550_hyprshot

vt-tv (approved):
LGTM

pedro-alves-ferreira:
@garethsb

For some reason I was notified by email but I can't find the comment on GitHub.

That sounds good. @SimaoPFonseca Do you have any other comments or do you want to try and implement it as Gareth suggests?

garethsb:
I've been doing a bit more digging (with 🤖 help). Can we catch up next week to discuss?

pedro-alves-ferreira:
> I've been doing a bit more digging (with 🤖 help). Can we catch up next week to discuss?

Of course.

garethsb (approved):
Changed the workaround to just skip the test case in CI until we can resolve the synchronization issue or race condition properly.

pedro-alves-ferreira (approved):
LGTM

PR #524, opened 2026-05-26
Refactor MxlSink to use element clock directly in render functions

garethsb:
# Problem

The `mxlsink` element assumed it was directly contained in a Pipeline, whereas it might be nested in a Bin.

# Details

- Removed the pipeline reference from the State struct.
- Updated render functions for audio, video, and data to accept an element reference and use its clock instead of the pipeline's clock.
- Simplified state management by eliminating unnecessary strong references that could lead to reference cycles.

# Backport requirements

None, `gst-mxl-rs` is only in main, not v1.0 branch.


pedro-alves-ferreira (approved):
LGTM

jonasohland (approved):
LGMT

PR #525, opened 2026-05-26
gst-mxl-rs: emit unquoted interlace-mode in mxlsrc caps

garethsb:
# Details

`mxlsrc::set_json_caps` was building the video caps' `interlace-mode` field via `serde_json::to_string(&video.interlace_mode)`. Because `InterlaceMode` is a serde enum with `#[serde(rename = "progressive")]` etc., this serialised a single variant to the *JSON encoding* `"progressive"` including the surrounding quote characters, so the caps ended up as `interlace-mode=(string)"\"progressive\""`.

Add `InterlaceMode::as_str()` returning the bare wire name that matches the `#[serde(rename = ...)]` attributes and use it from `set_json_caps`, so the caps contain the unquoted string, e.g. `progressive` that was intended.

This won't be the last PR on this, because those NMOS `interlaced_tff` and `interlaced_bff` values are not appropriate for GStreamer, which wants just `interleaved` for a single buffer containing interlaced fields or `alternate` for a buffer per field.

However, because the policy may end up being to remove all interlaced support from the MXL SDK, I'm not getting into this now.

# Backport requirements

None

### DIFF
--- a/rust/gst-mxl-rs/src/mxlsrc/mxl_helper.rs
+++ b/rust/gst-mxl-rs/src/mxlsrc/mxl_helper.rs
@@ -57,12 +57,7 @@ pub(crate) fn set_json_caps(src: &MxlSrc, json: FlowDefDetails) -> Result<(), gs
                     "framerate",
                     gst::Fraction::new(video.grain_rate.numerator, video.grain_rate.denominator),
                 )
-                .field(
-                    "interlace-mode",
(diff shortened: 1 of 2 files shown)


pedro-alves-ferreira (approved):
LGTM

vt-tv (approved):
LGTM

PR #526, opened 2026-05-28
Add fabrics docs from the wiki to the repo.

### LINKED ISSUE #181: Fabrics: PR Step 2: API implementation

jonasohland:
# Details
Add the fabrics docs from the project wiki to the repo.

# Backport requirements
None

### DIFF
--- a/docs/Fabrics.md
+++ b/docs/Fabrics.md
@@ -40,7 +40,8 @@ As stated in the proposition, the preferred method to use to transfer data is Re
 Remote Write supports scatter-gather functionality on the source side, allowing the initiator to specify multiple non-contiguous buffers as the source. However, the destination is limited to a single contiguous address, which prevents unpacking the results back into multiple non-contiguous buffers and creates an asymmetry in the transfer operation. To address this problem, we have to use a bouncing buffer on the target side. On completion, the target is responsible for unpacking data back to non-contiguous buffers.
 
 ## 3.2 Send/Recv
-There are some cases where the hardware does not support Remote Write. One example of this is some AWS instances that have EFA enabled hardware only have support for Send/Recv transfers. One problem with Send/Recv is that there is no guarantee that a transfer will always end up in the buffer the target expects. One example of this is when a target posts a receive for grain index X, but the initiator never transfers X and instead transfers X+1. Grain X+1 ends up in grain X’s buffer region, which is incorrect. To avoid this situation, we must use a bouncing buffer. Naturally, we would like to avoid that extra copy to save on CPU resources and reduce latency.
+There are some cases where the hardware does not support Remote Write. One example of this is some AWS instances that have EFA enabled hardware only have support for Send/Recv transfers.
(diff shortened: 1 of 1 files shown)


jonasohland:
closes #181

docs/Fabrics.md
```diff
@@ -0,0 +1,335 @@
...
+- Support for continuous and discrete flows memory layouts
+- Support user media buffers (that follow mxl memory layouts)
+- Support both Host and Device memory
+- Support different providers
+    - TCP
+    - Verbs (RDMA)
+    - EFA 
+- Node / Service (ip:port) to establish communication (connected or connection-less)
+- Support One-Way transfers (Remote Write)
+- Support Two-Way transfers (Send/Recv) as fallback (if One-Way transfers are not supported)
```
k0zmo, line 17: Does this document describe current state or the "endgame"? I don't remember seeing send/recv implementation

docs/Fabrics.md
```diff
@@ -0,0 +1,335 @@
...
+
+## 3.2 Send/Recv
+There are some cases where the hardware does not support Remote Write. One example of this is some AWS instances that have EFA enabled hardware only have support for Send/Recv transfers. One problem with Send/Recv is that there is no guarantee that a transfer will always end up in the buffer the target expects. One example of this is when a target posts a receive for grain index X, but the initiator never transfers X and instead transfers X+1. Grain X+1 ends up in grain X’s buffer region, which is incorrect. To avoid this situation, we must use a bouncing buffer. Naturally, we would like to avoid that extra copy to save on CPU resources and reduce latency.
```
k0zmo: The way it is stated now, it's not entirely true. In general sense, having a single outstanding fi_recv() you don't need extra bounce buffer. Maybe something along the line:

With untagged Send/Recv, the receiver cannot direct an incoming message to a specific buffer as posted receives are consumed in FIFO order. Since the target doesn't know in advance which grain the initiator will send next, it cannot pre-post the grain's final buffer and rely on the data landing there. To land data directly in the correct per-grain buffer one would need tag matching (FI_TAGGED, tag = grain index), otherwise a single bounce buffer is used and the grain index is recovered from the immediate data, which results in an extra copy.

Also, for older EFA, I recall they claim RDMA write support but it's emulated using send/recv and bounce buffer (because of out-of-order semantics). Is there a reason for us to implement the fallback then? Unless there are more devices (verbs?) without RDMA write support

jonasohland: Also an answer to your other comment:
This is/was more like a design document. I think we have agreed not to touch send/recv for now. I will add a paragraph about send/recv not currently being implemented, and reword that this paragraph based on your suggestion.

k0zmo (approved):

vt-tv (approved):
approving on behalf of @k0zmo

sviengkhou (approved):
LGTM

PR #528, opened 2026-05-28
Garbage-collect orphan flow directories in Instance constructor

### LINKED ISSUE #527: mxlGarbageCollectFlows isn't automatically called
`mxlGarbageCollectFlows` docs say:

https://github.com/dmf-mxl/mxl/blob/666a3e1fd78be138903ba81d999f26b016383445/lib/include/mxl/mxl.h#L87-L91

but I don't see any calls apart from in `mxl-info`?

garethsb:
Resolves #527 

# Details

Invoke `Instance::garbageCollect()` once at the end of the constructor body to honour the documented contract. The call is best-effort: any exception is caught and logged inside `garbageCollect` itself, so a failure here cannot abort instance creation.

Add a Catch2 test that fabricates a leftover `<flowId>.mxl-flow/data` file with no flock held -- the on-disk fingerprint of a crashed writer -- then constructs a fresh instance and asserts the directory is gone.

# Backport requirements

Maybe?

### DIFF
--- a/lib/internal/src/Instance.cpp
+++ b/lib/internal/src/Instance.cpp
@@ -90,6 +90,19 @@ namespace mxl::lib
         std::call_once(loggingFlag, [&]() { initializeLogging(); });
         parseOptions(options);
         MXL_DEBUG("Instance created. MXL Domain: {}", mxlDomain.string());
+
+        // Honour the contract documented for `mxlGarbageCollectFlows` in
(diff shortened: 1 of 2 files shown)


vt-tv:
The documentation was stale.  It was agreed a while ago that we don't want to automatically invoke garbage collection.

PR #529, opened 2026-05-28
feat(rust): expose garbage_collect_flows

garethsb:
# Details

Wrap the C `mxlGarbageCollectFlows` entry point in a safe Rust method on `MxlInstance`. Long-running applications can use this to reclaim shared memory left behind by writer processes that exited or crashed before their destructors ran (SIGKILL, segfault, host reboot), without having to recreate the instance.

Add a smoke test in `basic_tests.rs` that runs the call on an empty domain to verify the FFI binding is wired correctly; the behavioural contract is covered by the C++ test suite.

Use this feature in the gst-mxl-rs `mxlsrc` and `mxlsink` elements which makes force killing a writer pipeline and relaunching it work correctly.

# Backport requirements

None

### DIFF
--- a/rust/gst-mxl-rs/src/mxlsink/imp.rs
+++ b/rust/gst-mxl-rs/src/mxlsink/imp.rs
@@ -450,5 +450,12 @@ fn init_mxl_instance(
             )
         })?;
 
+    // Best-effort: reclaim any flow directories left behind by a writer that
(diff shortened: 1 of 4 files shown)


pedro-alves-ferreira (approved):
LGTM

vt-tv (approved):
LGTM.  (not a rust expert however)

PR #534, opened 2026-05-28
Print domain_def.json fields if available

### LINKED ISSUE #532: [Small Improvement] mxl-info: display domain_id if present
### Description

A domain folder can specify a domain id (found in a domain_def.json file at the root of the domain).  If preset, print it out when invoking mxl-info.

See : https://specs.amwa.tv/bcp-007-03/branches/v1.0-dev/docs/NMOS-With-MXL.html

### Additional Context

_No response_

vt-tv:
# Details

Simple addition to mxl-info : print domain_def.json details if found.

# Backport requirements

no.

### DIFF
--- a/tools/mxl-info/main.cpp
+++ b/tools/mxl-info/main.cpp
@@ -25,6 +25,72 @@ namespace
 {
     namespace detail
     {
+        /// Helper class to parse the domain definition JSON file and extract relevant information for display.
+        class DomainDefParser
+        {
+        public:
+            /// Construct the parser and parse the domain definition JSON file at the given path.
+            /// \param domainDefPath The path to the domain definition JSON file.
+            /// \throws std::runtime_error if the file cannot be read or parsed.
+            explicit DomainDefParser(std::filesystem::path const& domainDefPath)
+            {
+                // Read the whole file into a string
+                auto file = std::ifstream{domainDefPath};
+                if (!file)
+                {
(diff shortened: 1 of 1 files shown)


tools/mxl-info/main.cpp
```diff
@@ -564,6 +630,24 @@ int main(int argc, char** argv)
         std::cerr << "ERROR: Domain must be specified either via --domain or in the URI." << std::endl;
         return EXIT_FAILURE;
     }
+    else
+    {
+        try
+        {
+            auto const domainDefPath = std::filesystem::path{domain} / "domain_def.json";
+            if (std::filesystem::exists(domainDefPath))
```
jonasohland, line 638: Probably not an issue in practice, but still a race condition.
It would better to just read the file and catch ENOENT.

vt-tv: It is also caught later as an exception in the method.

jonasohland: But it will log an error in that case.
I see that `std::ifstream` does not even throw an appropriate exception. So a proper fix would be to throw a std::system_error with the ENOENT error code, and check for it in the top level exception handler.
Again, probably not an issue in practice, so I am just going to approve as-is.

jonasohland (approved):

garethsb (approved):

PR #535, opened 2026-05-29
Include libfabric error code in fabrics Event::Error::toString

dclasen-qd:
# Details

`Event::Error::toString()` only renders the provider-specific component via
`fi_eq_strerror()`. When the failure is at the libfabric layer (an `FI_E*`
error with `prov_errno == 0`), `fi_eq_strerror()` returns `"Success"`, so a
real connect failure shows up as a bare `Success` line on every retry and
reads as noise.

Render both `fi_strerror(_err)` and the provider's `fi_eq_strerror()` output,
e.g. `fi_err=111 (Connection refused) prov_err=0 (Success)` instead of just
`Success`.

Downstream tracking: qvest-digital/go-mxl#51.

# Backport requirements

None

### DIFF
--- a/lib/fabrics/ofi/src/internal/Event.cpp
+++ b/lib/fabrics/ofi/src/internal/Event.cpp
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "Event.hpp"
+#include <fmt/format.h>
 #include <rdma/fabric.h>
 #include "EventQueue.hpp"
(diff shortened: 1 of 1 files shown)


jonasohland (approved):
Thanks!

garethsb (approved):

jonasohland (changes requested):
Code needs to be clang-formatted.

jonasohland (approved):

PR #539, opened 2026-05-31
fix(fabrics/ofi): guard target setup logging against null node/service

aerben:
`RCTarget::setup` and `RDMTarget::setup` format `mxlFabricsEndpointAddress.node/service` via fmt before calling `fi_getinfo`, but those `char const*` fields can be null and fmt throws on null, which makes `MXL_INFO` drop the record and (in RCTarget) makes the no-fabric `Exception::make` re-throw as a format error instead of `MXL_ERR_NO_FABRIC`. Substitute with "<unspecified>" for formatting only

### DIFF
--- a/lib/fabrics/ofi/src/internal/RCTarget.cpp
+++ b/lib/fabrics/ofi/src/internal/RCTarget.cpp
@@ -20,7 +20,18 @@ namespace mxl::lib::fabrics::ofi
 
     std::pair<std::unique_ptr<RCTarget>, std::unique_ptr<TargetInfo>> RCTarget::setup(mxlFabricsTargetConfig const& config)
     {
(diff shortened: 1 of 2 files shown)


jonasohland:
Hi! Thank you for this fix. You will need to run clang-format on the changed files.
I am working on a PR that will replace all of this code, but if it fixes a problem you have right now, we can still merge it.

aerben:
@jonasohland yeah, I stumbled upon this while testing, so a merge is much appreciated. Fix the clang issue

jonasohland (approved):
Thanks!

vt-tv (approved):
LGTM.  Thanks for this contribution.

PR #540, opened 2026-05-31
Add optional target info file output/input to fabrics demo app

jonasohland:
# Details
Extend the `--target-info` command line option of the fabrics demo app. Existing apps should not break through this change.

Before, on the target side this option was ignored. Now it can specify a file to which the target info should be written. It is still always logged as base64. A leading '@' is stripped from the path if put there accidentally.

On the initiator, it will still parse the command line option as a base64-encoded target info blob. But when prefixed with an '@', it will instead read a raw target info string from a file at the path following the '@'.

Example:
```bash
# Target
mxl-fabrics-demo -d /dev/shm (...) --target-info [@]target.json  # Written to target.json
mxl-fabrics-demo -d /dev/shm (...)                               # Still visible in the log

# Initiator
mxl-fabrics-demo -d /dev/shm (...) --target-info @target.json                  # Read from target.json
mxl-fabrics-demo -d /dev/shm (...) --target-info "$(base64 -w0 < target.json)" # Command line arg interpreted as base64 target info
```

# Backport requirements
Nope.

### DIFF
--- a/tools/mxl-fabrics-demo/demo.cpp
+++ b/tools/mxl-fabrics-demo/demo.cpp
@@ -796,8 +796,8 @@ int main(int argc, char** argv)
         std::string flowDescriptor{std::istreambuf_iterator<char>(file), std::istreambuf_iterator<char>()};
         mxl::lib::FlowParser descriptorParser{flowDescriptor};
 
-        std::string resolvedTargetInfo;
-        if (!targetInfo.empty() && targetInfo[0] == '@')
+        auto resolvedTargetInfo = std::string{};
+        if (!targetInfo.empty() && targetInfo.starts_with('@'))
         {
             // Read raw target info from file
             auto filePath = std::string{targetInfo.begin() + 1, targetInfo.end()};
--- a/tools/mxl-fabrics-demo/demo.cpp
+++ b/tools/mxl-fabrics-demo/demo.cpp
@@ -552,6 +552,7 @@ public:
             MXL_ERROR("Failed to convert target info to string with status '{}'", static_cast<int>(status));
             return status;
         }
+        targetInfoStr.pop_back();
 
         MXL_INFO("Target info:  {}", base64::to_base64(targetInfoStr));
 
--- a/tools/mxl-fabrics-demo/demo.cpp
+++ b/tools/mxl-fabrics-demo/demo.cpp
@@ -557,7 +557,7 @@ public:
 
         if (!targetInfoFile.empty())
         {
-            std::ofstream of{targetInfoFile};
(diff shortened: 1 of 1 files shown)


tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -784,6 +795,24 @@ int main(int argc, char** argv)
         std::string flowDescriptor{std::istreambuf_iterator<char>(file), std::istreambuf_iterator<char>()};
         mxl::lib::FlowParser descriptorParser{flowDescriptor};
 
+        std::string resolvedTargetInfo;
+        if (!targetInfo.empty() && targetInfo[0] == '@')
+        {
+            // Read raw target info from file
+            std::string filePath{targetInfo.begin() + 1, targetInfo.end()};
```
vt-tv: auto const filePath = std::string{...} syntax?

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -784,6 +795,24 @@ int main(int argc, char** argv)
...
+            std::ifstream of{filePath};
```
vt-tv: same

vt-tv (changes requested):

vt-tv (approved):
LGTM

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -784,6 +796,24 @@ int main(int argc, char** argv)
         std::string flowDescriptor{std::istreambuf_iterator<char>(file), std::istreambuf_iterator<char>()};
         mxl::lib::FlowParser descriptorParser{flowDescriptor};
 
+        std::string resolvedTargetInfo;
+        if (!targetInfo.empty() && targetInfo[0] == '@')
```
mlefebvre1: Could you use `starts_with('@')` instead?

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -849,7 +879,8 @@ int main(int argc, char** argv)
             return status;
         }
 
-        if (status = app.printInfo(); status != MXL_STATUS_OK)
+        auto targetInfoPath = targetInfo.empty() || targetInfo[0] != '@' ? targetInfo : std::string{targetInfo.begin() + 1, targetInfo.end()};
```
mlefebvre1, line 882: Sorry, forgot to comment previously, but we could use starts_with here also.

mlefebvre1 (approved):
Looks good, just a last minor thing.

PR #541, opened 2026-06-01
Add docs action and support script

peterbrightwell:
# Details

Add docs action to create GitHub pages using Zensical and versioned with mike.

Includes a script to:
- make a copy of the main README in the docs tree on gh-pages
- rewrite links to targe the repo where they would break because they are going outside the `docs/` tree

Addresses #364

Some Copilot code

# Backport requirements

Should backport to previous branches that we want to include in the version


peterbrightwell:
Tested on my fork before merging, and then here using manual workflow dispatch (had to make the gh-pages branch manually first).  

Please try out any further changes on `test-doc-build`  branch (nb can't use `user/abc/xyz` format branch name because it's treated as a version name and `/` not allowed.

.github/scripts/prepare-docs.sh
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project.
```
vt-tv, line 2: 2026 everywhere

PR #544, opened 2026-06-04
Add some timings clarifications based on NAB learnings.

### LINKED ISSUE #509: [TSC] Document the learnings from the NAB Interops
Just checking if we need any further actions to address the learnings from last weeks interop.

vt-tv:
# Details

Add some clarifications related to timing based on NAB learnings.

# Backport requirements

yes

### DIFF
--- a/docs/Timing.md
+++ b/docs/Timing.md
@@ -5,6 +5,8 @@
 
 As described in the [architecture](./Architecture.md) page, the grains of an MXL flow are organized in a ring buffer. Each index of the ring buffer correspond to a timestamp relative to the PTP epoch as defined by SMPTE 2059-1. MXL does NOT require a PTP/SMPTE 2059 time source : it only _leverages_ the epoch and clock definitions (TAI time) as defined in SMPTE 2059-1.
 
+A comprehensive definition of the DMF and MXL timing model is being developed by the [JT-DMF End-to-End Synchronisation Working Group][def].
+
 ## Requirements
 
 For single host MXL environments, any stable time sources such as NTP or SMPTE ST 2059 is adequate.
@@ -16,6 +18,8 @@ Any NTP locked source that can be traced back to a Stratum 0 time source (atomic
 ## Ring buffers
 
 Grains are organized as ring buffers. The grain count is a function of the flow grain_rate and desired history length. For example if a flow grain_rate is 50/1, each grain will be worth 20'000'000 ns (1/50 \* 1'000'000) of data. A ring buffer with 100ms of history will have 5 grains (5 \* 20ms/grain = 100ms). Time starts at the SMPTE 2059-1 epoch and is expressed a nanoseconds since that epoch.
+
+Ring buffer indices encode timing information — **the ring buffer cannot be treated as a simple queue of grains.** A writer media function must carefully compute the grain index at which it writes based on the frequency domain of the incoming grains. If the writer cannot guarantee that the external source shares a common frequency domain with the MXL host, it must implement a re-synchronization strategy (grain drop, grain repeat, dynamic audio resampling, etc.) to maintain alignment with the host's timing.
 
 ### Example indexing
 
@@ -82,6 +86,8 @@ MXL Flow replication between hosts will preserve the indexing of the grains. In 
 
 - A 2110-20 receiver media function can unroll the RTP timestamp of the first RTP packet of the frame and use it to compute the grain index write to. This index may be slightly in the past.
 - A media function that generates grains locally (CG, Test Pattern Generator, etc) can simply use the current TAI time as a timestamp and convert it to a grain index.
+- A well behaved writer media function should try to 'pace itself' according to the MXL host clock.  It should write to the ring buffer at a mostly constant rate and avoid writing large number of grains at once.
+- A well behaved writer media function will not create flows whose latency drift over time. An implementer can monitor a flow latency using the 'mxl-info' tool part of the SDK.  The flow latency should be low and mostly constant (minimal variation) over a long period of time (hours).
 
 ### Example Flow Readers behaviors
 
@@ -94,8 +100,9 @@ $$
 ReadIndex = min(F1_{head} ... FN_{head})
 $$
 
-
 ## Time functions
 
 All time helpers operate on nanoseconds and rational time bases. Continuous flows typically pass the sample rate (`grainRate`) directly to these
 helpers when scheduling wake-ups or calculating how far to advance `headIndex`. Refer to `mxl/time.h` for the canonical API surface.
+
+[def]: https://www.amwa.tv/jtdmf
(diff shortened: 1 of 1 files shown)


docs/Timing.md
```diff
@@ -17,6 +19,8 @@ Any NTP locked source that can be traced back to a Stratum 0 time source (atomic
 
 Grains are organized as ring buffers. The grain count is a function of the flow grain_rate and desired history length. For example if a flow grain_rate is 50/1, each grain will be worth 20'000'000 ns (1/50 \* 1'000'000) of data. A ring buffer with 100ms of history will have 5 grains (5 \* 20ms/grain = 100ms). Time starts at the SMPTE 2059-1 epoch and is expressed a nanoseconds since that epoch.
 
+Ring buffer indices encode timing information — **the ring buffer cannot be treated as a simple queue of grains.** A writer media function must carefully compute the grain index at which it writes based on the frequency domain of the incoming grains. If the writer cannot guarantee that the external source shares a common frequency domain with the MXL host, it must implement a re-synchronization strategy (grain drop, grain repeat, dynamic audio resampling, etc.) to maintain alignment with the host's timing.
```
lapointejp:
```suggestion
Ring buffer indices encode timing information — **the ring buffer cannot be treated as a simple queue of grains.** A writer media function must carefully compute the grain index at which it writes based on the frequency domain of the incoming grains. If the writer cannot guarantee that the external source shares a common clock frequency with the MXL host, it must implement a re-synchronization strategy (grain drop, grain repeat, dynamic audio resampling, etc.) to maintain alignment with the host's timing.
```
Nitpick on wording.

docs/Timing.md
```diff
@@ -82,6 +86,8 @@ MXL Flow replication between hosts will preserve the indexing of the grains. In
 
 - A 2110-20 receiver media function can unroll the RTP timestamp of the first RTP packet of the frame and use it to compute the grain index write to. This index may be slightly in the past.
 - A media function that generates grains locally (CG, Test Pattern Generator, etc) can simply use the current TAI time as a timestamp and convert it to a grain index.
+- A well behaved writer media function should try to 'pace itself' according to the MXL host clock.  It should write to the ring buffer at a mostly constant rate and avoid writing large number of grains at once.
```
lapointejp: I would change the wording to account for 1 sample = 1 grain for audio.
Suggest something that might look like: "It should write to the ring buffer at a mostly constant rate and avoid writing chunks of media representing large amounts of time (i.e. multiple frames for video)."

lapointejp (changes requested):

docs/Timing.md
vt-tv: The terminology is chosen to be similar to what is used in the DMF Sync workgroup (common frequency domains)

docs/Timing.md
vt-tv: Changed.

docs/Timing.md
lapointejp: Doesn't the group call it a "time domain" ?
Would be good to reference terminology if we can when there is a document published.

lapointejp (approved):

docs/Timing.md
```diff
@@ -82,6 +86,8 @@ MXL Flow replication between hosts will preserve the indexing of the grains. In
 
 - A 2110-20 receiver media function can unroll the RTP timestamp of the first RTP packet of the frame and use it to compute the grain index write to. This index may be slightly in the past.
 - A media function that generates grains locally (CG, Test Pattern Generator, etc) can simply use the current TAI time as a timestamp and convert it to a grain index.
+- A well behaved writer media function should try to 'pace itself' according to the MXL host clock.  It should write to the ring buffer at a mostly constant rate and avoid writing chunks of media representing large amounts of time (i.e. multiple frames for video, multiple seconds of audio, etc)."
```
pac-work: Hello, @vt-tv, homework accomplished :)

Do we want to be more precise with some kind of recommendation? For example adding to this point:

```
Also, every grain written to the ring buffer should stay available there for at least half of it’s duration (for example for 50 fps, that would be at least `(1s/50)/2 = 10 ms`), before writer can overwrite it with a new one.
```

Or something in those lines? That would at least avoid arguments about whether it was the writer’s or a reader’s fault...

But the formulation can even stay just as is, if you prefer. 

It seemed like you also had an opinion, @pedro-alves-ferreira?

pedro-alves-ferreira: Yes, my point was more or less the same: to clearly state how much ahead in time can a producer write, and how much behind in time can a consumer read, without stepping on each other's feet.

lapointejp: How about:
`The average write frequency should match the flow's grain rate and the time between two write calls should remain lower than 2x the duration of the chunk being written and shall never exceed half the ring buffer depth` ?

pedro-alves-ferreira: > How about: `The average write frequency should match the flow's grain rate and the time between two write calls should remain lower than 2x the duration of the chunk being written and shall never exceed half the ring buffer depth` ?

I like the principle but I have a few concerns:
- we should clearly say that the reference is the local MXL clock, e.g. "(...) average write frequency (as measured using the local MXL clock) (...)"
- when you say "never exceed half the ring buffer depth", do you mean the time between two consecutive writes, or the accumulated drift from the reference clock?
- If we allow a writer to be lagging behind up to half of the ring buffer duration, the implication is that the minimum latency is precisely half of the ring buffer duration, because readers will have to be reading "the other half" of the ring buffer. Given that the duration is defined by the domain, I wonder what side effects that will cause.

pac-work: > the time between two write calls should remain lower than 2x the duration of the chunk being written

This may force certain elements along the pipeline to introduce artificial latency, just to be sure they adhere to this. Imagine you have a decoder (or anything else which is dependent on upstream data). You will get "bursty" batches of frames to write into MXL, but you will not be exactly sure when will the next "batch" came. Even if your domain is large enough and you could easily just drop everything there, this rule would force you to not do so and pace yourself to make sure you can keep writing till your next batch comes from upstream and you have more data to write.

I was trying to rather look at this from the perspective of what is actually problematic on bursty behavior and in my eyes, the problem is that the reader may not have enough time to read and process what was written. That is why I tried to go in the direction of "_every written frame has to stay available for at least X_". Or is there any other problem we are trying to mitigate by this rule in question?

vt-tv: that's indeed the direction : your decoder mediafunction will need to implement some pacing.

docs/Timing.md
vt-tv: As discussed in the TSC call, the 'frequency domain' terminology is not used anymore in the end-to-end timing JT-DMF group.  Will apply @lapointejp suggestion.

docs/Timing.md
felixpou: TSC

- if we can relate the 'pacing' to the %size of the ring buffer, this would provide guidance on the expected.

- @vt-tv will come back with changed text

felixpou (approved):
good enough for now. in the future we may have to specify more precisely the tolerance of asynchronicity vs real time, but let see first if this langage provides sufficient guidance.

PR #546, opened 2026-06-04
Update lychee-action to v2.0.2

vt-tv:
# Details

Fix a dependabot reported securtity issue.

# Backport requirements

No.

PR #547, opened 2026-06-04
Use lychee directly for docs workflow

### LINKED ISSUE #364: [Small Improvement]  Create a GitHub Pages for clean documentation

peterbrightwell:
# Details

#546 caused some lychee argument passing problems with lychee-action v2, which made the workflow fail, so instead installing lychee directly.

Using snap install (not ideal I know) because cargo install takes too long. If there's a running with cargo binstall installed we can use that.

# Backport requirements

No

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -79,11 +79,12 @@ jobs:
         run: ./.github/scripts/prepare-docs.sh
 
       - run: zensical build --clean
-
+      
+      - name: Install lychee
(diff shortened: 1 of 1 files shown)


vt-tv (approved):
LGTM

jonasohland (approved):

PR #553, opened 2026-06-05
Accept lychee timeout, bump setup-python

peterbrightwell:
# Details

Documentation build has sometimes failed when a target times out. This PR means such timeouts will be ignored so the workflow doesn't fail (which is probably better? We already ignore various HTTP errors such as not authorised.)

It also bumps setup-python which was giving a node version warning.

Tested on my fork.

# Backport requirements

No

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -7,7 +7,7 @@ on:
   push:
     branches:
       - main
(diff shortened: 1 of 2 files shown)


vt-tv (approved):
LGTM

jonasohland (approved):

PR #558, opened 2026-06-12
Ensure CMAKE install directory matches link search directory

hori985:
# Details

On Linux distributions that default to `lib64` for 64-bit libraries (Fedora, RHEL, openSUSE, Arch), CMake installs `libmxl.so` under `out/lib64` while the `mxl-sys` build script unconditionally emits `cargo:rustc-link-search=.../out/lib`, causing the linker to fail with unable to find library `-lmxl`.
This fix explicitly sets `CMAKE_INSTALL_LIBDIR=lib` in the CMake invocation, ensuring the install path and the link search path always agree regardless of system defaults.

### Changed:
- `rust/mxl-sys/build.rs`: added `.define("CMAKE_INSTALL_LIBDIR", "lib")` to the `cmake::Config` builder

# Backport requirements

This is a build system bug affecting any `lib64`-defaulting distro. Recommend backporting to all supported release branches.

### DIFF
--- a/rust/mxl-sys/build.rs
+++ b/rust/mxl-sys/build.rs
@@ -50,6 +50,7 @@ fn get_bindgen_specs() -> BindgenSpecs {
             .define("BUILD_DOCS", "OFF")
             .define("BUILD_TESTS", "OFF")
             .define("BUILD_TOOLS", "OFF")
+            .define("CMAKE_INSTALL_LIBDIR", "lib")
             .build();
 
         println!("cargo:rustc-link-search={}", dst.join("lib").display());


jonasohland:
Interesting, I am on Arch, and I never saw this problem.

jonasohland (approved):
LGTM

hori985:
@jonasohland 
> Interesting, I am on Arch, and I never saw this problem.

It happens to me on Fedora, idk if it's just my machine being crazy, but I was guessing that it's not such a bad idea to hardcode install dir since the linking is also hardcoded. :shrug:

vt-tv:
@pedro-alves-ferreira What do you think about this change?  any risk of breaking the crates or application using them (gstreamer plugins, etc)

KimonHoffmann (changes requested):
Isn't this the wrong way around to go about this?

The Rust bindings should search for the library where its is located according to the standards implied by the distribution, rather than forcing the library to be located at the location the bindings expect it at.

pedro-alves-ferreira:
> @pedro-alves-ferreira What do you think about this change? any risk of breaking the crates or application using them (gstreamer plugins, etc)

I think this is fine.

pedro-alves-ferreira:
> Isn't this the wrong way around to go about this?
> 
> The Rust bindings should search for the library where its is located according to the standards implied by the distribution, rather than forcing the library to be located at the location the bindings expect it at.

@KimonHoffmann This is only used when building the bindings and running the Rust applications in "development mode", that is, when the MXL SDK is not installed. In runtime, it will load libmxl.so from the system-defined ld path. We are not hardcoding the RPATH.

pedro-alves-ferreira (approved):
LGTM

KimonHoffmann (approved):
Thanks for the clarification @pedro-alves-ferreira.
Looks good to me then.

PR #564, opened 2026-06-16
fabrics: Capabilities and Interfaces implementation

### LINKED ISSUE #576: [Feature] Fabrics: Interfaces and Capabillites
### Description

Make internal interface handling more robust, expose interface information to API consumers, fix IPv6

### Linked Requirement
[dmf-mxl/mxl-requirements#38: [REQ] Inter-Host memory sharing](https://github.com/dmf-mxl/mxl-requirements/issues/38)

jonasohland:
Hi all, 

Sorry this PR is a bit longer (and later) than anticipated, but it adds a lot of value and fixes a lot of issues and things that were not properly defined before.
For example: Using ipv6 was basically not possible before, but now it is totally transparent and works the same as ipv4.
Users of EFA will also get useful error messages and can even discover beforehand if their instance/efa is supported or not.

I have tested this on Verbs (NVIDIA ConnectX and Intel E810), EFA (C7 and C5n), TCP and SHM also work as expected and report correctly through the new `mxlFabricsGetInterfaces`.

This also gives us a better starting position to support device memory with fabrics when we get there.

Quick summary:

- Adds a mxlFabricsGetInterfaces / mxlFabricsFreeInterfaceList API pair that enumerates fabric interfaces available on the host. The call accepts an optional filter (provider + address) and returns a linked list where each node
  describes one interface: provider, endpoint address, capability flags, max message size, and JSON device attributes.

- Capability flags (REMOTE_WRITE, SEND_RECEIVE, BLOCKING_OPERATIONS) now carry explicit dual semantics: informational when returned from GetInterfaces, requirements when passed to the setup functions. The header docs were
  updated to make this clear.

- Internally, the query and setup paths are now separated. ProviderConfig translates MXL capability flags into libfabric parameters; FabricInterfaceProbe either enumerates all matching interfaces (query) or selects a single one
  and enforces capability requirements (setup).

- The demo tool was updated to use GetInterfaces for automatic interface selection (EFA > VERBS > TCP > SHM) so --provider is no longer required on the command line.

There is also a new mxl-fabrics-info CLI tool that prints discovered interfaces in a compact table or verbose form, with optional provider and address filters.

Example output of mxl-fabrics-demo on my local machine:
```bash
$ mxl-fabrics-info
interface shm-0 node jonas-tower-arch dev <none>
interface tcp-0 node 100.105.158.138 dev tailscale0
interface tcp-1 node 10.135.0.150 dev wlan0
interface tcp-2 node fe80::ae19:8eff:fee6:a500 dev wlan0
interface tcp-3 node 127.0.0.1 dev lo
interface tcp-4 node ::1 dev lo
```
Verbose output and filtering by address:
```bash
$ mxl-fabrics-info -v -n ::1
build/Linux-GCC-Debug/tools/mxl-fabrics-info/mxl-fabrics-info -n ::1 -v                                                                    [17:51:29]
interface tcp-0
  provider:      tcp
  node:          ::1
  max message:   18446744073709551615 bytes
  capabilities:  BLOCKING_OPERATIONS | REMOTE_WRITE | SEND_RECEIVE
  info:
    device_name: lo
    ep_addr_format: FI_SOCKADDR_IN6
    ep_protocol: FI_PROTO_SOCK_TCP
    ep_type: FI_EP_MSG
    fi_domain_cq_cnt: 256
    fi_domain_ep_cnt: 8192
    fi_domain_mr_cnt: 0
    fi_domain_name: lo
Found [  1] interfaces for [  tcp] provider.
```
Example verbs (Intel E810):
```bash
$ mxl-fabrics-info -p verbs -n 10.80.132.2
interface verbs-0
  provider:      verbs
  node:          10.80.132.2
  max message:   2147483647 bytes
  capabilities:  BLOCKING_OPERATIONS | REMOTE_WRITE | SEND_RECEIVE
  info:
    device_firmware: 1.72
    device_id: 0x1592
    device_name: rocep0s16
    device_vendor_id: 0x8086
    device_version: 2
    ep_addr_format: FI_SOCKADDR_IN
    ep_protocol: FI_PROTO_RDMA_CM_IB_RC
    ep_type: FI_EP_MSG
    fi_domain_cq_cnt: 8189
    fi_domain_ep_cnt: 4092
    fi_domain_mr_cnt: 4194302
    fi_domain_name: rocep0s16
    link_speed: 100 gbit/s
    link_state: up
    link_type: Ethernet
    pci_bus_id: 0
    pci_device_id: 16
    pci_domain_id: 0
    pci_function_id: 0
```

Example output on EFA that does not support remote write:
```bash
$ mxl-fabrics-info -v -p efa
interface efa-0
  provider:      efa
  node:          fe80::478:afff:fe48:6177
  max message:   8928 bytes
  capabilities:  BLOCKING_OPERATIONS | SEND_RECEIVE
  info:
    device_driver: efa
    device_firmware: 0.0.0.0
    device_id: 0xefa0
    device_name: rdmap0s31
    device_vendor_id: 0x1d0f
    device_version: 6
    ep_addr_format: FI_ADDR_EFA
    ep_protocol: FI_PROTO_EFA
    ep_type: FI_EP_RDM
    fi_domain_cq_cnt: 512
    fi_domain_ep_cnt: 256
    fi_domain_mr_cnt: 262144
    fi_domain_name: rdmap0s31-rdm
    link_address: EFA-fe80::478:afff:fe48:6177
    link_speed: 50 gbit/s
    link_state: up
    link_type: Ethernet
    pci_bus_id: 0
    pci_device_id: 31
    pci_domain_id: 0
    pci_function_id: 0
```

There are also some small misc fixes included in this request:
- libfabric version bumped in devcontainer install script.
- FI_RMA cap added unconditionally on setup path.
- Various bug fixes: segfault on exit (uninitialized memory read in demo), incorrect picojson include in mxl-fabrics-info, boolean sub-expression bracketing, and other small stuff.
- Tests added: test_ProviderConfig.cpp, test_interfaces.cpp, expanded test_Address.cpp.


vt-tv:
@jonasohland  is this PR linked to a ticket on the TSC board? It looks like all caps related tickets are already closed/merged.

lib/fabrics/ofi/src/internal/FabricAddress.cpp
```diff
@@ -0,0 +1,385 @@
...
+        }
+
+        std::optional<std::string> hostString(std::string const& str)
+        {
+            return str.substr(0, str.find(':'));
+        }
+
+        std::optional<std::string> serviceString(::sockaddr_in const& sin)
+        {
+            return sin.sin_port == 0 ? std::nullopt : std::make_optional(std::to_string(sin.sin_port));
```
k0zmo, line 156: Shouldn't this be `std::to_string(ntohs(sin.sin_port))`?

lib/fabrics/ofi/src/internal/FabricAddress.cpp
```diff
@@ -0,0 +1,385 @@
...
+        }
+
+        std::optional<std::string> serviceString(::sockaddr_in6 const& sin)
+        {
+            return sin.sin6_port == 0 ? std::nullopt : std::make_optional(std::to_string(sin.sin6_port));
```
k0zmo, line 162: Same here

lib/fabrics/ofi/src/internal/FabricAddress.cpp
```diff
@@ -0,0 +1,385 @@
...
+        FabricAddress::Native parseInternetAddress(std::string const& ipString, std::uint16_t port = 0)
+        {
+            if (std::ranges::find(ipString, ':') != ipString.end())
+            {
+                auto addr = ::in6_addr{};
+                if (!::inet_pton(AF_INET6, ipString.c_str(), &addr))
+                {
+                    throw Exception::invalidArgument("invalid node address");
+                }
+                return FabricAddress::Native{
```
k0zmo, line 34: I think it should be:
```
.sin_port = htons(port)
.sin6_port = htons(port)
```
when constructing `sockaddr_in`

lib/fabrics/ofi/src/internal/RDMTarget.cpp
```diff
@@ -10,57 +10,44 @@
...
-        // throws fmt::format_error("string pointer is null") on a nullptr, which MXL_INFO
-        // silently swallows. Substitute a printable sentinel for the log only — the original
-        // pointers still go to fi_getinfo so the wildcard semantics are preserved.
-        char const* const nodeStr = config.interface.address.node ? config.interface.address.node : "<unspecified>";
-        char const* const serviceStr = config.interface.address.service ? config.interface.address.service : "<unspecified>";
-
-        MXL_INFO("setting up target [endpoint = {}:{}, provider = {}]", nodeStr, serviceStr, config.interface.provider);
-
-        auto provider = providerFromAPI(config.interface.provider);
+        auto const provider = providerFromString(info->fabric_attr->prov_name);
```
k0zmo, line 27: We don't call `requireCapability` here? All the others call it:
`RCInitiator::setup()` requires `FI_WRITE`
`RCTarget::setup()` requires `FI_REMOTE_WRITE`
`RDMInitiator::setup()` requires `FI_WRITE`

lib/fabrics/ofi/src/internal/ProviderConfig.cpp
```diff
@@ -0,0 +1,173 @@
...
+            auto result = std::uint64_t{0};
+            if (capabilities->interfaceCaps & MXL_FABRICS_IFACE_CAP_REMOTE_WRITE)
+            {
+                result |= isTarget ? FI_REMOTE_WRITE : FI_WRITE;
+            }
+            if (capabilities->interfaceCaps & MXL_FABRICS_IFACE_CAP_SEND_RECEIVE)
+            {
+                result |= FI_SEND | FI_RECV;
+            }
+            return result | FI_RMA;
```
k0zmo: I think `FI_RMA` is redundant here (`FI_REMOTE_WRITE` and `FI_WRITE` should be enough) but if applied, it should only be for `FI_REMOTE_WRITE` or `FI_WRITE`, not unconditionally.

jonasohland: Some providers provide duplicate capability sets, one with FI_RMA, one without. I don't no why, and it does not really make sense to me...
So this is just here to filter out the duplicate set that does not have FI_RMA. I am moving it to the correct branch above, because unconditionally adding it here is wrong.

lib/fabrics/ofi/src/internal/FabricAddress.cpp
```diff
@@ -0,0 +1,385 @@
...
+        }
+
+        std::optional<std::string> serviceString(EfaAddress)
+        {
+            return std::nullopt;
+        }
+
+        std::optional<std::string> serviceString(std::string const& str)
+        {
+            auto const pos = str.find(':') + 1;
```
k0zmo: if `.find(':')` returns `npos` and we increment it by 1, we'll end up with `pos` equal to 0. The increment should be done after the `if` check

garethsb: I'd also want to name the parameter something more descriptive than`str`

dclasen-qd:
Nice work @jonasohland ! Really looking forward to integrate that in [qvest-digital/mxl-k8s](https://github.com/qvest-digital/mxl-k8s) and start wrapping my head around some early support for node-capability-aware scheduling in k8s. :thinking: 
Just to sanity check that I got it right, this will drop the `ANY` provider and the fallthrough to `TCP`, right? Provider auto-selection, as is wasn't really implemented in the first place, will stay a concern for downstream applications to implement? Or do you plan to ship it in libmxl-fabrics in the future with a best-effort selection/cascade?

vt-tv:
@jonasohland  @k0zmo  can you resolve the conversations that can be resolved? I am waiting for the conversations to be resolved before approving.

k0zmo:
I don't have option to do so, missing permissions most likely

jonasohland:
The last one that is open I still need to address. All others should be closed.

lib/fabrics/include/mxl/fabrics.h
```diff
@@ -65,28 +64,43 @@ extern "C"
...
     } mxlFabricsEndpointAddress;
 
+    /** Capability flags describing what an interface supports.
+     *
+     * When returned by mxlFabricsGetInterfaces(), these flags are informational: they describe the capabilities
+     * of each discovered interface. The caller reads them to decide which interface to use.
+     *
+     * When passed to mxlFabricsTargetSetup() or mxlFabricsInitiatorSetup() (via mxlFabricsInterfaceConfig), these
+     * flags express requirements. At least one transfer capability (REMOTE_WRITE or SEND_RECEIVE) must be set.
+     * If no transfer capability is set, REMOTE_WRITE is assumed.
```
KimonHoffmann: This seems slightly self contradictory "... at least one transfer capability [...] *must* be set ..." vs. "If no transfer capability is set [...] is assumed."

I'd resolve this by either not assuming anything and enforcing the stated requirement instead, or by defining a `MXL_FABRICS_IFACE_[CAP_]TRANSFER_DEFAULT` enumerator the makes the automatic choice explicit.

lib/fabrics/include/mxl/fabrics.h
```diff
@@ -65,28 +64,43 @@ extern "C"
...
+        MXL_FABRICS_IFACE_CAP_BLOCKING_OPERATIONS = MXL_FABRICS_FLAG(0), /**< The interface supports blocking operations. */
+        MXL_FABRICS_IFACE_CAP_REMOTE_WRITE = MXL_FABRICS_FLAG(1),        /**< The interface supports remote-write (RDMA) operations. */
+        MXL_FABRICS_IFACE_CAP_SEND_RECEIVE = MXL_FABRICS_FLAG(2),        /**< The interface supports send/receive message operations. */
     } mxlFabricsInterfaceCapFlags;
 
-    /** Capabilities of an interface */
+    /** Capabilities of an interface. Returned as informational output by mxlFabricsGetInterfaces().
+     *  When used inside mxlFabricsInterfaceConfig for setup functions, only the \c flags field is
+     *  evaluated as requirements; \c maxMessageSize is ignored.
+     */
```
KimonHoffmann, line 86: Is there a true benefit of using the same structure for `mxlFabricsGetInterfaces` and `mxlFabricsInterfaceConfig`, if most fields are not used in the latter case anyway?

Could instead be replaced with a more reduced `mxlFabricsInterfaceCapReqs_t` or something along those lines.

lib/fabrics/include/mxl/fabrics.h
```diff
@@ -96,6 +110,18 @@ extern "C"
         char const* attr;
     } mxlFabricsInterfaceConfig;
 
+    /**
+     * A list of interfaces returned by mxlFabricsGetInterfaces. Should be freed with mxlFabricsFreeInterfaceList
+     */
+    typedef struct mxlFabricsInterfaceList_t mxlFabricsInterfaceList;
```
KimonHoffmann, line 116: Why separate the typedef from the `struct` definition in this particular case?

lib/fabrics/ofi/src/internal/Address.hpp
```diff
@@ -36,8 +39,10 @@ namespace mxl::lib::fabrics::ofi
 
         /**
          * \brief Parse a fabric address from a base64 encoded string
+         * \param data Base64 encoded raw binary address
+         * \param addressFormat A libfabric address format like FI_SOCKADDR_IN, FI_ADDR_EFA, etc..
          */
-        static FabricAddress fromBase64(std::string_view data);
+        static RawFabricAddress fromBase64(std::string_view data, std::uint32_t addressFormat);
```
KimonHoffmann: Should probably be `[[nodiscard]]`.

Also, since `addressFormat` is a very special implementation specific enumeration type, would it make sense to define a strong alias for that one in order to avoid accidental type confusion or implicit conversions?

```cpp
enum class FabricsAddressFormat : std::uint32_t {};
```

lib/fabrics/ofi/src/internal/Address.hpp
```diff
@@ -56,17 +61,30 @@ namespace mxl::lib::fabrics::ofi
         [[nodiscard]]
         std::size_t size() const noexcept;
 
-        bool operator==(FabricAddress const& other) const noexcept;
+        /**
+         * \brief Return the fabric specific libfaric address format, or FI_FORMAT_UNSPEC if empty.
+         */
+        [[nodiscard]]
+        std::uint32_t format() const noexcept;
```
KimonHoffmann: As above WRT having a strong alias for address formats.

lib/fabrics/ofi/src/internal/Address.hpp
```diff
@@ -56,17 +61,30 @@ namespace mxl::lib::fabrics::ofi
...
 
         /** \brief Retrieve the fabric address of an endpoint by using its fid.
          */
-        static FabricAddress retrieveFabricAddress(::fid_t);
+        static RawFabricAddress retrieveFabricAddress(::fid_t, std::uint32_t);
 
     private:
-        std::vector<std::uint8_t> _inner; /**< libfabric address represented as a buffer of bytes */
+        std::vector<std::uint8_t> _inner = {}; /**< libfabric address represented as a buffer of bytes */
+        std::uint32_t _addressFormat = FI_FORMAT_UNSPEC;
```
KimonHoffmann: Could we move those two explicit initializations to an explicitly defined default constructor?

lib/fabrics/ofi/src/internal/Address.hpp
```diff
@@ -56,17 +61,30 @@ namespace mxl::lib::fabrics::ofi
...
+        bool operator==(RawFabricAddress const& other) const noexcept;
 
     private:
-        explicit FabricAddress(std::vector<std::uint8_t> addr);
+        explicit RawFabricAddress(std::vector<std::uint8_t> addr, std::uint32_t addressFormat);
 
         /** \brief Retrieve the fabric address of an endpoint by using its fid.
          */
-        static FabricAddress retrieveFabricAddress(::fid_t);
+        static RawFabricAddress retrieveFabricAddress(::fid_t, std::uint32_t);
```
KimonHoffmann: Please keep parameter names in interface declarations.

lib/fabrics/ofi/src/internal/Address.cpp
```diff
@@ -13,51 +13,62 @@
 
 namespace mxl::lib::fabrics::ofi
 {
-    FabricAddress::FabricAddress(std::vector<std::uint8_t> addr)
+    RawFabricAddress::RawFabricAddress(std::vector<std::uint8_t> addr, std::uint32_t addressFormat)
         : _inner(std::move(addr))
+        , _addressFormat(addressFormat)
```
KimonHoffmann: Please use curly brace initializers.

lib/fabrics/ofi/src/internal/Address.cpp
```diff
@@ -71,7 +82,7 @@ namespace mxl::lib::fabrics::ofi
         std::vector<std::uint8_t> addr(addrlen);
```
KimonHoffmann: Please use AA style (2x)

lib/fabrics/ofi/src/internal/Address.cpp
```diff
@@ -71,7 +82,7 @@ namespace mxl::lib::fabrics::ofi
...
         fiCall(fi_getname, "Failed to retrieve endpoint's local address.", fid, addr.data(), &addrlen);
 
-        return FabricAddress{addr};
+        return RawFabricAddress{addr, addressFormat};
```
KimonHoffmann: Need to `std::move()` `adds` for this to be as efficient as the interface attempts to be.

lib/fabrics/ofi/src/internal/Format.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include <string>
+#include <vector>
+#include <rdma/fabric.h>
+
+namespace mxl::lib::fabrics::ofi
+{
+    std::string interfaceCapsString(std::uint64_t caps)
+    {
+        auto capStrings = std::vector<std::string>{};
+        if (caps & MXL_FABRICS_IFACE_CAP_BLOCKING_OPERATIONS)
```
KimonHoffmann: I prefer explicit boolean conditions. I.e. `((bits & MASK) != 0)`

lib/fabrics/ofi/src/internal/Format.cpp
```diff
@@ -0,0 +1,48 @@
...
+        if (caps & MXL_FABRICS_IFACE_CAP_SEND_RECEIVE)
+        {
+            capStrings.emplace_back("SEND_RECEIVE");
+        }
+        if (capStrings.empty())
+        {
+            capStrings.emplace_back("<none>");
+        }
+
+        auto result = capStrings.front();
```
KimonHoffmann, line 34: The code could sum the total number of characters with each `emplace_back` and then reserve the necessary amount once.

Also prefer `result.push_back('|')` to `append("|")`.

lib/fabrics/ofi/src/internal/ProviderConfig.cpp
```diff
@@ -0,0 +1,173 @@
...
+#include "ProviderConfig.hpp"
+#include "Exception.hpp"
+
+namespace mxl::lib::fabrics::ofi
+{
+    namespace
+    {
+        namespace
+        {
+            constexpr std::uint64_t const supportedMemoryRegistrationModes = FI_MR_VIRT_ADDR | FI_MR_LOCAL | FI_MR_ALLOCATED | FI_MR_PROV_KEY;
```
KimonHoffmann: Prefer right hand init.

lib/fabrics/ofi/src/internal/ProviderConfig.cpp
```diff
@@ -0,0 +1,173 @@
...
+            auto result = std::uint64_t{0};
+            if (capabilities->interfaceCaps & MXL_FABRICS_IFACE_CAP_REMOTE_WRITE)
+            {
+                result |= (isTarget ? FI_REMOTE_WRITE : FI_WRITE) | FI_RMA;
+            }
+            if (capabilities->interfaceCaps & MXL_FABRICS_IFACE_CAP_SEND_RECEIVE)
+            {
+                result |= FI_SEND | FI_RECV;
+            }
+            return result;
```
KimonHoffmann, line 30: More streamlined:

```cpp
auto result = std::uint64_t{0};
if (capabilities)
{
    // ...
}
return result;
```

lib/fabrics/ofi/src/internal/ProviderConfig.cpp
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        std::uint64_t libfabricCaps(std::optional<ProviderCapabilities> const& capabilities, bool isTarget)
+        {
+            if (!capabilities)
+            {
+                return 0;
+            }
+            auto result = std::uint64_t{0};
+            if (capabilities->interfaceCaps & MXL_FABRICS_IFACE_CAP_REMOTE_WRITE)
```
KimonHoffmann: As above wrt explicit boolean conditions.

lib/fabrics/ofi/src/internal/ProviderConfig.cpp
```diff
@@ -0,0 +1,173 @@
...
+            {
+                result |= FI_RMA;
+            }
+            return result;
```
KimonHoffmann, line 44: Both comments as above.

lib/fabrics/ofi/src/internal/ProviderConfig.cpp
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    std::uint64_t ProviderConfig::getCaps() const noexcept
+    {
+        return _values.caps;
+    }
+
+    ProviderConfig::ProviderConfig(ProviderConfigValues values, std::optional<ProviderCapabilities> capabilities)
+        : _values(std::move(values))
+        , _capabilities(capabilities)
```
KimonHoffmann: Please use curly brace initialization.

lib/fabrics/ofi/src/internal/ProviderConfig.hpp
```diff
@@ -0,0 +1,103 @@
...
+        /**
+         * Returns true if this is a supported fabric configuration info. Can be used to filter the output of fi_getinfo to only a list of supported
+         * fabric configs.
+         */
+        [[nodiscard]]
+        bool isSupportedFabricInfo(FabricInfoView view) const noexcept;
+
+        /** \brief Get the provider name (as expected by libfabric in prov_name) as a string. */
+        [[nodiscard]]
+        std::string getProviderName() const noexcept;
```
KimonHoffmann: Something that returns a `std::string` can not realistically be `noexcept`.... Except if the string is always empty 😉

lib/fabrics/ofi/src/internal/ProviderConfig.hpp
```diff
@@ -0,0 +1,103 @@
...
+        [[nodiscard]]
+        ::fi_ep_type getEndpointType() const noexcept;
+
+        /** \brief Return the libfabric capability flags for this provider configuration. */
+        [[nodiscard]]
+        std::uint64_t getCaps() const noexcept;
+
+    private:
+        ProviderConfig(ProviderConfigValues values, std::optional<ProviderCapabilities> capabilities);
+        ProviderConfigValues _values;
```
KimonHoffmann, line 102: Could you please separate the constructor from the members with another `private:` visibility declarator.

lib/fabrics/ofi/src/internal/AddressVector.cpp
```diff
@@ -44,22 +44,22 @@ namespace mxl::lib::fabrics::ofi
         struct MakeSharedEnabler : public AddressVector
         {
             MakeSharedEnabler(fid_av* raw, std::shared_ptr<Domain> domain)
-                : AddressVector(raw, domain)
+                : AddressVector(raw, std::move(domain))
```
KimonHoffmann, line 47: Please use curly brace initialization.

lib/fabrics/ofi/src/internal/FabricAddress.hpp
```diff
@@ -0,0 +1,133 @@
...
+        /** \brief Decode a raw libfabric address buffer of the given `addr_format`.
+         *
+         * \param addrFormat One of the libfabric FI_ADDR_* / FI_SOCKADDR* constants.
+         * \param addr       Pointer to the address bytes, or nullptr.
+         * \param len        Length of the address buffer in bytes.
+         */
+        [[nodiscard]]
+        static FabricAddress decode(std::uint32_t addrFormat, void const* addr, std::size_t len) noexcept;
+
+        FabricAddress() noexcept = default;
```
KimonHoffmann: Please explicitly implement this constructor, or at least move the defaulted implementation to the `cpp` file.

lib/fabrics/ofi/src/internal/FabricAddress.hpp
```diff
@@ -0,0 +1,133 @@
...
+        [[nodiscard]]
+        std::uint32_t addressFormat() const noexcept;
+
+        /** \brief Parse a node/service pair into a typed address for the given provider. */
+        static FabricAddress parse(Provider provider, std::optional<std::string> node, std::optional<std::string> service);
+
+    private:
+        explicit FabricAddress(Native native) noexcept
+            : _native(std::move(native))
+        {}
```
KimonHoffmann: Please use curly braces initialization and move the implementation to the `cpp` file.

lib/fabrics/ofi/src/internal/FabricInterfaceProbe.cpp
```diff
@@ -0,0 +1,181 @@
...
+            }
+
+            if (requested.empty())
+            {
+                return true;
+            }
+
+            if (*provider == Provider::SHM)
+            {
+                return requested.node() == getHostname();
```
KimonHoffmann, line 51: This will probably fail, if the memory is shared across containers!?

lib/fabrics/ofi/src/internal/FabricInterfaceProbe.cpp
```diff
@@ -0,0 +1,181 @@
...
+        auto getProviderConfig = [&](Provider provider) -> ProviderConfig const&
+        {
+            auto found = providerConfigs.find(provider);
+            if (found != providerConfigs.end())
+            {
+                return found->second;
+            }
+
+            auto [emplaced, _] = providerConfigs.emplace(provider, ProviderConfig::create(provider, false, std::nullopt));
+            return emplaced->second;
```
KimonHoffmann: Skip the find, as this will just needlessly duplicate the log(n) tree traversal in the insert case.

But of course only *iff* `ProviderConfig` is cheap to create.

lib/fabrics/ofi/src/internal/FabricInterfaceProbe.cpp
```diff
@@ -0,0 +1,181 @@
...
+            }
+
+            try
+            {
+                auto [emplaced, _] = fabricAddresses.emplace(provider,
+                    FabricAddress::parse(provider,
+                        query ? optStringFromCStr(query.value().get().address.node) : std::nullopt,
+                        query ? optStringFromCStr(query.value().get().address.service) : std::nullopt));
+
+                return emplaced->second;
```
KimonHoffmann: Same as above.
If the temporary is not cheap to create (as seems to be the case here) I'd suggest to replace the `find` + `emplace` with a `lower_bound` + hinted `emplace`.

lib/fabrics/ofi/src/internal/FabricInterfaceProbe.cpp
```diff
@@ -0,0 +1,181 @@
...
+            }
+            catch (Exception const&)
+            {
+                return fabricAddresses.emplace(provider, std::nullopt).first->second;
+            }
+        };
+
+        auto const requestedProvider = query ? providerFromAPI(query.value().get().provider) : Provider::ANY;
```
KimonHoffmann: Spreading this to two lines makes this easier to read:

```cpp
auto const requestedProvider = query
    ? providerFromAPI(query.value().get().provider)
    : Provider::ANY;
```

lib/fabrics/ofi/src/internal/FabricInterfaceProbe.cpp
```diff
@@ -0,0 +1,181 @@
...
+        if (!requestedProvider)
+        {
+            throw Exception::invalidArgument("Invalid provider");
+        }
+
+        auto list = FabricInterfaceList{};
+        for (auto info : FabricInfoList::get())
```
KimonHoffmann: I know that deep down the value type is a non-owning view and cheap to copy, but still using `auto&&` here makes this a no brainer.

lib/fabrics/ofi/src/internal/FabricInterfaceProbe.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+namespace mxl::lib::fabrics::ofi
+{
+    /**
+     * Query path: enumerate all available fabric interfaces matching the optional filter criteria.
+     * Capabilities in the query are ignored — all matching interfaces are returned with their
+     * reported capabilities as informational output. When \p query is nullopt, all interfaces
+     * across all providers are returned.
+     */
+    FabricInterfaceList probeInterfaces(std::optional<std::reference_wrapper<::mxlFabricsInterfaceConfig const>> query);
```
KimonHoffmann, line 19: Should probably be `[[nodiscard]]`.

lib/fabrics/ofi/src/internal/FabricInterfaceProbe.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * Setup path: select a single source interface for target or initiator setup.
+     * Capabilities from \p interfaceConfig are treated as requirements — if no transfer capability
+     * (REMOTE_WRITE or SEND_RECEIVE) is set, REMOTE_WRITE is applied with a warning. Returns the
+     * matched fabric info and its resolved provider configuration.
+     * \param isTarget true for target (FI_REMOTE_WRITE), false for initiator (FI_WRITE).
+     * \throws Exception::noFabric if no matching interface is found or capabilities are unsupported.
+     */
+    std::pair<FabricInfo, ProviderConfig> selectSourceInterface(::mxlFabricsInterfaceConfig const& interfaceConfig, bool isTarget);
```
KimonHoffmann, line 30: Should probably be `[[nodiscard]]`.

lib/fabrics/ofi/src/internal/FabricInterfaceList.hpp
```diff
@@ -0,0 +1,38 @@
...
+        void push(FabricInterfaceDescription iface);
+
+        /** Remove all collected interface descriptions. */
+        void clear();
+
+        /**
+         * Convert the collected descriptions into a heap-allocated mxlFabricsInterfaceList linked list.
+         * Ownership of the returned list transfers to the caller; free with freeRawLinkedList().
+         */
+        ::mxlFabricsInterfaceList* toRawLinkedList() noexcept;
```
KimonHoffmann, line 31: Should probably be `[[nodiscard]]`.

lib/fabrics/ofi/src/internal/Initiator.cpp
```diff
@@ -28,25 +29,13 @@ namespace mxl::lib::fabrics::ofi
             _inner.reset();
         }
 
-        switch (config.interface.provider)
+        auto [info, provierConfig] = selectSourceInterface(config.interface, /* target */ false);
```
KimonHoffmann: Can probably be  `const`?!

lib/fabrics/ofi/src/internal/RCInitiator.cpp
```diff
@@ -120,6 +121,8 @@ namespace mxl::lib::fabrics::ofi
                     state.ep.bind(cq, FI_TRANSMIT);
 
                     // Transition into the connecting state
+                    auto faddr = _info.fabricAddress.decode();
```
KimonHoffmann: Can probably be `const`?!

lib/fabrics/ofi/src/internal/RDMInitiator.cpp
```diff
@@ -117,40 +117,27 @@ namespace mxl::lib::fabrics::ofi
...
+        requireCapability(info, FI_WRITE, "Interface is missing required remote write capability");
 
         auto cqAttr = CompletionQueue::Attributes::defaults();
-        if (provider == Provider::EFA)
+        if (config.interface.provider == MXL_FABRICS_PROVIDER_EFA)
         {
             if (!CompletionQueue::isWaitObjectSupportedForEFA())
             {
-                MXL_WARN("Wait objects not supported in EFA provider for this libfabric version. Only non-blocking API available.");
+                if (config.interface.caps.flags & MXL_FABRICS_IFACE_CAP_BLOCKING_OPERATIONS)
```
KimonHoffmann: As above with regards to explicit boolean conditionals.

lib/fabrics/ofi/src/internal/RDMTarget.cpp
```diff
@@ -10,57 +10,46 @@
...
         auto endpoint = Endpoint::create(domain);
 
         auto cqAttr = CompletionQueue::Attributes::defaults();
-        if (provider == Provider::EFA)
+        if (config.interface.provider == MXL_FABRICS_PROVIDER_EFA)
         {
             if (!CompletionQueue::isWaitObjectSupportedForEFA())
             {
-                MXL_WARN("Wait objects not supported in EFA provider for this libfabric version. Only non-blocking API available.");
+                if (config.interface.caps.flags & MXL_FABRICS_IFACE_CAP_BLOCKING_OPERATIONS)
```
KimonHoffmann: As above.

lib/fabrics/ofi/src/internal/Target.cpp
```diff
@@ -75,33 +76,27 @@ namespace mxl::lib::fabrics::ofi
...
+
     std::unique_ptr<TargetInfo> TargetWrapper::setup(mxlFabricsTargetConfig const& config)
     {
         if (_inner)
         {
             _inner.reset();
         }
 
-        switch (config.interface.provider)
+        auto [info, providerConfig] = selectSourceInterface(config.interface, /* target */ true);
```
KimonHoffmann, line 94: Can be `const`?!

lib/fabrics/ofi/src/internal/Target.hpp
```diff
@@ -133,6 +133,13 @@ namespace mxl::lib::fabrics::ofi
         std::unique_ptr<TargetInfo> setup(mxlFabricsTargetConfig const& config);
 
     private:
+        /** \brief Set up the correct concrete target type internally and returns the target info
+         * \param config Target configuration passed by the user.
+         * \param info Fabric info already resolved from the interface config.
+         */
+        template<typename TargetT>
+        std::unique_ptr<TargetInfo> setup(mxlFabricsTargetConfig const& config, FabricInfoView info);
```
KimonHoffmann: Should probably be `[[nodiscard]]`.

lib/fabrics/ofi/src/internal/Target.hpp
```diff
@@ -133,6 +133,13 @@ namespace mxl::lib::fabrics::ofi
         std::unique_ptr<TargetInfo> setup(mxlFabricsTargetConfig const& config);
```
KimonHoffmann: Should probably be `[[nodiscard]]`.

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -25,32 +26,139 @@
...
 
 void signal_handler(int)
 {
     g_exit_requested = 1;
 }
 
+std::string capabilitiesString(std::uint64_t flags)
+{
+    auto parts = std::vector<std::string>{};
+    if (flags & MXL_FABRICS_IFACE_CAP_BLOCKING_OPERATIONS)
```
KimonHoffmann: As above with regards to explicit boolean conditions.

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -25,32 +26,139 @@
...
+    }
+    if (flags & MXL_FABRICS_IFACE_CAP_SEND_RECEIVE)
+    {
+        parts.emplace_back("SEND_RECEIVE");
+    }
+    if (parts.empty())
+    {
+        return "(none)";
+    }
+    auto result = parts.front();
```
KimonHoffmann: Similar comment as above with regards to counting characters and reserving, but might not be worth the effort here in particular.

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -25,32 +26,139 @@
...
+    {
+        case MXL_FABRICS_PROVIDER_EFA:   return 4;
+        case MXL_FABRICS_PROVIDER_VERBS: return 3;
+        case MXL_FABRICS_PROVIDER_TCP:   return 2;
+        case MXL_FABRICS_PROVIDER_SHM:   return 1;
+        default:                         return 0;
+    }
+}
+
+InterfaceSelection selectInterface(mxlFabricsInstance instance, std::optional<std::string> const& node, std::optional<std::string> const& service,
```
KimonHoffmann, line 118: All of the helper functions above should probably be `[[nodiscard]]`.

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -721,25 +829,25 @@ class AppTarget
...
-    mxlFlowWriter _writer;
-    mxlFabricsTarget _target;
-    mxlFabricsTargetInfo _targetInfo;
-    mxlFlowConfigInfo _configInfo;
+    mxlInstance _instance = nullptr;
+    mxlFabricsInstance _fabricsInstance = nullptr;
+    mxlFlowWriter _writer = nullptr;
+    mxlFabricsTarget _target = nullptr;
+    mxlFabricsTargetInfo _targetInfo = nullptr;
+    mxlFlowConfigInfo _configInfo{};
```
KimonHoffmann: Move all of these initializations to the constructor initializer list.

tools/mxl-fabrics-demo/demo.cpp
KimonHoffmann: Should use curly brace initialization.

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -417,24 +525,24 @@ class AppInitator
 private:
     mxlStatus makeProgress(std::chrono::steady_clock::duration timeout)
     {
-        if (_config.provider == MXL_FABRICS_PROVIDER_EFA)
+        if (_config.interface.caps.flags & MXL_FABRICS_IFACE_CAP_BLOCKING_OPERATIONS)
```
KimonHoffmann: As above with regards to explicit boolean conditions.

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -721,25 +829,25 @@ class AppTarget
 private:
     mxlStatus targetReadGrain(std::uint64_t* grainIndex, std::chrono::steady_clock::duration timeout)
     {
-        if (_config.provider == MXL_FABRICS_PROVIDER_EFA)
+        if (_config.interface.caps.flags & MXL_FABRICS_IFACE_CAP_BLOCKING_OPERATIONS)
```
KimonHoffmann: Same.

tools/mxl-fabrics-info/main.cpp
```diff
@@ -0,0 +1,265 @@
...
+#include <unistd.h>
+#include <CLI/CLI.hpp>
+#include <fmt/format.h>
+#include <picojson/wrapper.h>
+#include <mxl/fabrics.h>
+#include <mxl/mxl.h>
+
+namespace
+{
+    char const* providerName(mxlFabricsProvider provider) noexcept
```
KimonHoffmann: Should probably be `[[nodiscard]]`. Can be `constexpr`.

tools/mxl-fabrics-info/main.cpp
```diff
@@ -0,0 +1,265 @@
...
+            case MXL_FABRICS_PROVIDER_ANY:   return "any";
+            case MXL_FABRICS_PROVIDER_TCP:   return "tcp";
+            case MXL_FABRICS_PROVIDER_VERBS: return "verbs";
+            case MXL_FABRICS_PROVIDER_EFA:   return "efa";
+            case MXL_FABRICS_PROVIDER_SHM:   return "shm";
+        }
+        return "unknown";
+    }
+
+    std::optional<mxlFabricsProvider> providerFromName(std::string const& name) noexcept
```
KimonHoffmann: Should probably be `[[nodiscard]]`. Can probably be `constexpr`.

tools/mxl-fabrics-info/main.cpp
```diff
@@ -0,0 +1,265 @@
...
+        if ((flags & MXL_FABRICS_IFACE_CAP_SEND_RECEIVE) != 0)
+        {
+            parts.emplace_back("SEND_RECEIVE");
+        }
+        if (parts.empty())
+        {
+            return "(none)";
+        }
+
+        auto result = parts.front();
```
KimonHoffmann: As above with regards to counting and reserving.

TBH (all of those) could also use a `std::ostringstream` to construct the result on the fly.

```cpp
auto oss = std::ostringstream{};
auto first = true;
if (...)
{
    oss << (first ? "" : " | ") << "WHETEVER";
    first = false;
}
// ...
if (first)
{
    oss << "(none)";
}
return oss.str();
```

lib/fabrics/ofi/src/internal/FabricAddress.cpp
```diff
@@ -0,0 +1,385 @@
...
+                    throw Exception::invalidArgument("invalid node address");
+                }
+                return FabricAddress::Native{
+                    ::sockaddr_in{.sin_family = AF_INET, .sin_port = ::htons(port), .sin_addr = addr, .sin_zero = {}}
+                };
+            }
+        }
+
+        /** \brief Wrap `inet_ntop`, returning std::nullopt on failure (mirrors `ofi_straddr`'s `return NULL`). */
+        std::optional<std::string> ipToString(int family, void const* addr)
```
KimonHoffmann, line 53: Should probably be `[[nodiscard]]`.

lib/fabrics/ofi/src/internal/FabricAddress.cpp
```diff
@@ -0,0 +1,385 @@
...
+        {
+            char str[INET6_ADDRSTRLEN] = {};
+            if (!::inet_ntop(family, addr, str, sizeof(str)))
+            {
+                return std::nullopt;
```
KimonHoffmann, line 58: IMHO it's nicer to make the true path the likely, successful path.

lib/fabrics/ofi/src/internal/FabricAddress.cpp
```diff
@@ -0,0 +1,385 @@
...
+
+        /// fi_sockaddr_in://%s:%u
+        std::optional<std::string> toString(::sockaddr_in const& sin)
+        {
+            auto const ip = ipToString(sin.sin_family, &sin.sin_addr);
+            if (!ip)
+            {
+                return std::nullopt;
+            }
+            return fmt::format("fi_sockaddr_in://{}:{}", *ip, ntohs(sin.sin_port));
```
KimonHoffmann, line 72: As above and in this case this even allows scoping `ip` a little bit tighter:

```cpp
if (auto const ip = ipToString(sin.sin_family, &sin.sin_addr); ip)
{
    return fmt::format("fi_sockaddr_in://{}:{}", *ip, ntohs(sin.sin_port));
}
return std::nullopt;
```

lib/fabrics/ofi/src/internal/FabricAddress.cpp
```diff
@@ -0,0 +1,385 @@
...
+
+        /// fi_sockaddr_in6://[%s]:%u
+        std::optional<std::string> toString(::sockaddr_in6 const& sin6)
+        {
+            auto const ip = ipToString(sin6.sin6_family, &sin6.sin6_addr);
+            if (!ip)
+            {
+                return std::nullopt;
+            }
+            return fmt::format("fi_sockaddr_in6://[{}]:{}", *ip, ntohs(sin6.sin6_port));
```
KimonHoffmann, line 84: Same.

lib/fabrics/ofi/src/internal/FabricAddress.cpp
```diff
@@ -0,0 +1,385 @@
...
+            char str[INET6_ADDRSTRLEN] = {};
+            if (!::inet_ntop(family, addr, str, sizeof(str)))
+            {
+                return std::nullopt;
+            }
+            return std::string{str};
+        }
+
+        /// fi_sockaddr_in://%s:%u
+        std::optional<std::string> toString(::sockaddr_in const& sin)
```
KimonHoffmann, line 65: Should be `[[nodiscard]]`.

lib/fabrics/ofi/src/internal/FabricAddress.cpp
```diff
@@ -0,0 +1,385 @@
...
+            auto const ip = ipToString(sin.sin_family, &sin.sin_addr);
+            if (!ip)
+            {
+                return std::nullopt;
+            }
+            return fmt::format("fi_sockaddr_in://{}:{}", *ip, ntohs(sin.sin_port));
+        }
+
+        /// fi_sockaddr_in6://[%s]:%u
+        std::optional<std::string> toString(::sockaddr_in6 const& sin6)
```
KimonHoffmann, line 77: Should be `[[nodiscard]]`.

lib/fabrics/ofi/src/internal/FabricAddress.cpp
```diff
@@ -0,0 +1,385 @@
...
+            if (!gid)
+            {
+                return std::nullopt;
+            }
+            return fmt::format("fi_sockaddr_ib://[{}]:0x{:x}:0x{:x}:0x{:x}",
+                *gid,                                                             // GID
+                ntohs(sib.sib_pkey),                                              // P_Key
+                static_cast<std::uint16_t>((be64toh(sib.sib_sid) >> 16) & 0xfff), // port space
+                static_cast<std::uint8_t>(be64toh(sib.sib_scope_id) & 0xff));     // Scope ID
+        }
```
KimonHoffmann, line 101: Similar comments as above also for the rest of the helpers in this file.

lib/fabrics/ofi/src/internal/FabricInterfaceDescription.cpp
```diff
@@ -0,0 +1,203 @@
...
+#include "CompletionQueue.hpp"
+#include "FabricAddress.hpp"
+#include "Format.hpp"
+#include "Provider.hpp"
+
+namespace mxl::lib::fabrics::ofi
+{
+    namespace
+    {
+        std::string getHostname()
```
KimonHoffmann, line 21: Should be `[[nodiscard]]`, but see also the comment below.

lib/fabrics/ofi/src/internal/FabricInterfaceDescription.cpp
```diff
@@ -0,0 +1,203 @@
...
+        {
+            char buf[256] = {};
+            ::gethostname(buf, sizeof(buf));
+            return buf;
+        }
+
+        void copyStringProperty(picojson::object& destination, char const* value, std::string const& name)
+        {
+            if ((value != nullptr) && (::strlen(value) > 0))
```
KimonHoffmann: Might make sense to opportunistically create and move the string, instead of doing one additional iteration for strlen?

```cpp
if (value != nullptr)
{
    if (auto str = std::string{value}; !str.empty())
    {
        destination[name] = picojson::value{std::move(str)};
    }
}
```

Or, alternatively, check for `(value[0] != '\0')`.

lib/fabrics/ofi/src/internal/FabricInterfaceDescription.cpp
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    FabricInterfaceDescription::FabricInterfaceDescription(Provider provider, std::string node, std::string service, std::uint64_t rawCaps,
+        std::uint64_t maxMessageSize, picojson::object extraInfo)
+        : _provider(provider)
+        , _node(std::move(node))
+        , _service(std::move(service))
+        , _maxMessageSize(maxMessageSize)
+        , _rawCaps(rawCaps)
+        , _attr(std::move(extraInfo))
```
KimonHoffmann: Please use curly brace initialization.

lib/fabrics/ofi/src/internal/FabricInterfaceDescription.cpp
```diff
@@ -0,0 +1,203 @@
...
+        else
+        {
+            node = faddr.node().value_or("");
+        }
+
+        if (info->caps & (FI_SEND | FI_RECV))
+        {
+            caps |= MXL_FABRICS_IFACE_CAP_SEND_RECEIVE;
+        }
+        if (info->caps & (FI_REMOTE_WRITE | FI_WRITE | FI_RMA))
```
KimonHoffmann: I prefer explicit boolean conditionals. I.e. `((word & MASK) != 0)`

lib/fabrics/ofi/src/internal/FabricInterfaceDescription.hpp
```diff
@@ -0,0 +1,54 @@
...
+     */
+    class FabricInterfaceDescription
+    {
+    public:
+        /**
+         * Create a description from a libfabric fi_info entry. Translates libfabric capabilities
+         * (FI_SEND, FI_WRITE, FI_RMA, etc.) into MXL capability flags and extracts NIC, domain,
+         * and endpoint attributes. Returns nullopt if the provider is not recognized.
+         */
+        static std::optional<FabricInterfaceDescription> create(FabricInfoView info);
```
KimonHoffmann, line 26: Should probably be `[[nodiscard]]`.

lib/fabrics/ofi/src/internal/FabricInterfaceDescription.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+        /**
+         * Allocate a heap-allocated mxlFabricsInterfaceList node from this description.
+         * String fields (node, service, attr) are strdup'd and owned by the node.
+         * \param next optional pointer to the next node in the linked list.
+         */
+        ::mxlFabricsInterfaceList* toRawLinkedListNode(::mxlFabricsInterfaceList* next = nullptr);
```
KimonHoffmann, line 34: Should be `[[nodiscard]]`.

lib/fabrics/ofi/src/internal/FabricInterfaceDescription.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+        /** Free a single linked list node and its strdup'd strings. Returns the next pointer. */
+        static ::mxlFabricsInterfaceList* freeRawLinkedListNode(::mxlFabricsInterfaceList*);
```
KimonHoffmann, line 38: Should be `[[nodiscard]]`.

lib/fabrics/ofi/src/internal/FabricInterfaceDescription.hpp
```diff
@@ -0,0 +1,54 @@
...
+        [[nodiscard]]
+        std::uint64_t caps() const noexcept
+        {
+            return _rawCaps;
+        }
+
+    private:
+        explicit FabricInterfaceDescription(Provider, std::string, std::string, std::uint64_t, std::uint64_t, picojson::object);
+
+        Provider _provider;
```
KimonHoffmann, line 51: Could use another `private:` visibility declarator to separate the constructor from the members.

lib/tests/fabrics/ofi/test_Address.cpp
```diff
@@ -2,14 +2,70 @@
...
+#include <netinet/in.h>
 #include "Address.hpp"
+#include "FabricAddress.hpp"
 
 using namespace mxl::lib::fabrics::ofi;
 
+namespace
+{
+    /// Build a native sockaddr_in, exactly as libfabric would hand it to FabricAddress::decode.
+    ::sockaddr_in makeSockaddrIn(char const* ip, std::uint16_t port)
```
KimonHoffmann, line 24: Should probably be `[[nodiscard]]`.

lib/tests/fabrics/ofi/test_Address.cpp
```diff
@@ -2,14 +2,70 @@
...
+    {
+        auto sin = ::sockaddr_in{};
+        sin.sin_family = AF_INET;
+        sin.sin_port = htons(port);
+        REQUIRE(::inet_pton(AF_INET, ip, &sin.sin_addr) == 1);
+        return sin;
+    }
+
+    ::sockaddr_in6 makeSockaddrIn6(char const* ip, std::uint16_t port)
```
KimonHoffmann, line 34: Should probably be `[[nodiscard]]`.

lib/tests/fabrics/ofi/test_Address.cpp
```diff
@@ -2,14 +2,70 @@
...
+        auto sin6 = ::sockaddr_in6{};
+        sin6.sin6_family = AF_INET6;
+        sin6.sin6_port = htons(port);
+        REQUIRE(::inet_pton(AF_INET6, ip, &sin6.sin6_addr) == 1);
+        return sin6;
+    }
+
+    /// Build a native ofi_sockaddr_ib. The service id packs the port space in bits 16..31 and
+    /// the port in bits 0..15; the P_Key, scope id and service id are stored in network order.
+    OfiSockaddrIb makeSockaddrIb(char const* gid, std::uint16_t pkey, std::uint16_t portSpace, std::uint64_t scopeId, std::uint16_t port)
```
KimonHoffmann, line 46: Should probably be `[[nodiscard]]`.

lib/tests/fabrics/ofi/test_Address.cpp
```diff
@@ -2,14 +2,70 @@
...
+        sib.sib_pkey = htons(pkey);
+        REQUIRE(::inet_pton(AF_INET6, gid, sib.sib_addr) == 1);
+        sib.sib_sid = htobe64((static_cast<std::uint64_t>(portSpace) << 16) | port);
+        sib.sib_scope_id = htobe64(scopeId);
+        return sib;
+    }
+
+    /// Build the 24-byte FI_ADDR_EFA blob: a 16-byte IPv6 GID, a 16-bit QPN at offset 16 and a
+    /// 32-bit QKey at offset 20, both in host byte order.
+    std::array<std::uint8_t, 24> makeEfa(char const* gid, std::uint16_t qpn, std::uint32_t qkey)
```
KimonHoffmann, line 60: Should probably be `[[nodiscard]]`.

lib/tests/fabrics/ofi/test_Address.cpp
```diff
@@ -18,7 +74,7 @@ TEST_CASE("ofi: FabricAddress default construction", "[ofi][FabricAddress]")
 
 TEST_CASE("ofi: FabricAddress base64 decode", "[ofi][FabricAddress]")
 {
-    FabricAddress addr = FabricAddress::fromBase64("AQIDBAU="); // base64 for {1,2,3,4,5}
+    RawFabricAddress addr = RawFabricAddress::fromBase64("AQIDBAU=", FI_FORMAT_UNSPEC); // base64 for {1,2,3,4,5}
```
KimonHoffmann: `auto (const)`?!

KimonHoffmann (changes requested):
Thank you for another massive contribution!
Looks good, with just mostly stylistic comments from my end.

lib/fabrics/ofi/src/internal/Address.hpp
jonasohland: Michael though it would be better to have a proper enum class for that. 
I have replaced all occurrences of the uint32_t format with a new `FabricAddressFormat` enum and added proper conversion functions that catch and unsupported or invalid format early.

lib/fabrics/ofi/src/internal/ProviderConfig.cpp
jonasohland: Do we have a policy against early return? Because after writing a lot of golang for the last years have become a big fan of that style and vastly prefer it over multiple levels of indentation.

lib/fabrics/ofi/src/internal/FabricInterfaceProbe.cpp
jonasohland: Thanks! I will be using lower_bound + hint in both cases.

lib/fabrics/ofi/src/internal/FabricAddress.cpp
jonasohland: Same question as above regarding early return

lib/fabrics/ofi/src/internal/FabricAddress.cpp
jonasohland: Same question as above regarding early return

lib/fabrics/ofi/src/internal/FabricAddress.cpp
jonasohland: Same question as above regarding early return

lib/fabrics/ofi/src/internal/FabricInterfaceDescription.cpp
jonasohland: picojson::value{} can only accept `std::string` via const reference or `char const*`. That would always be a guaranteed copy. So I will check for `value[0] != '\0'`

jonasohland:
@KimonHoffmann Thanks for your review! I have addressed most of your comments, just one open question about early return.

lib/fabrics/include/mxl/fabrics.h
jonasohland: Yes I don't know why I wrote that, that totally does not make sense. I will resolve this by enforcing the requirement.

lib/fabrics/include/mxl/fabrics.h
jonasohland: The reason for this is ergonimics. For 99% of usecases, the user can just call the `GetInterfaces` function and pass the interface that they want to use directly to the `Setup` functions without modifications.
I am planning to support fragmentation for interfaces that have a limited message length soon. This will then use the maxMessageSize field.

lib/fabrics/include/mxl/fabrics.h
jonasohland: I am using the typedef inside the struct itself as the pointer to the next linked list node. So the typedef needs to be forward declared.

jonasohland:
> Nice work @jonasohland ! Really looking forward to integrate that in [qvest-digital/mxl-k8s](https://github.com/qvest-digital/mxl-k8s) and start wrapping my head around some early support for node-capability-aware scheduling in k8s. 🤔 Just to sanity check that I got it right, this will drop the `ANY` provider and the fallthrough to `TCP`, right? Provider auto-selection, as is wasn't really implemented in the first place, will stay a concern for downstream applications to implement? Or do you plan to ship it in libmxl-fabrics in the future with a best-effort selection/cascade?

No support for automatic provider selection is planned right now. But with the new APIs, users should be able to implement their own.

lib/fabrics/ofi/src/internal/ProviderConfig.cpp
pedro-alves-ferreira: Same here.

lib/fabrics/ofi/src/internal/FabricInterfaceProbe.cpp
jonasohland: The `shm` provider does not return any default node address, and it is mostly meaningless for SHM anyway, because by definition, all endpoints are on the same node.
With how interface selection works from the `mxl-fabrics` level, an empty node address would be kind of confusing, so on the query path behind `mxlFabricsGetInterfaces`, I am filling in the hostname for the single `shm` entry.
The user must pass the same value as the node value when binding a target or initiator, which mimics the behavior of the other providers.
This will always work the same inside one container, between containers, and on the host, as long as the user uses the correct node value, which is always queriable through `mxlFabricsGetInterfaces`.

lib/fabrics/ofi/src/internal/RDMTarget.cpp
```diff
@@ -10,43 +10,25 @@
...
-        // throws fmt::format_error("string pointer is null") on a nullptr, which MXL_INFO
-        // silently swallows. Substitute a printable sentinel for the log only — the original
-        // pointers still go to fi_getinfo so the wildcard semantics are preserved.
-        char const* const nodeStr = config.interface.address.node ? config.interface.address.node : "<unspecified>";
-        char const* const serviceStr = config.interface.address.service ? config.interface.address.service : "<unspecified>";
-
-        MXL_INFO("setting up target [endpoint = {}:{}, provider = {}]", nodeStr, serviceStr, config.interface.provider);
-
-        auto provider = providerFromAPI(config.interface.provider);
+        auto const provider = providerFromString(info->fabric_attr->prov_name);
```
k0zmo, line 27: I still don't see `requireCapability()` here

k0zmo (changes requested):

lib/fabrics/ofi/src/internal/ProviderConfig.cpp
KimonHoffmann: No, we don't I just prefer it the other way round.
Also having early returns in some cases disables NRVO and there's also some evidence that compilers assume that the true path of a conditional is the regular/likely control flow and optimize for that.

lib/fabrics/include/mxl/fabrics.h
KimonHoffmann: This comment still needs to be updated accordingly.

lib/fabrics/ofi/src/internal/FabricInterfaceProbe.cpp
```diff
@@ -0,0 +1,196 @@
...
+namespace mxl::lib::fabrics::ofi
+{
+    namespace
+    {
+        std::optional<std::string> optStringFromCStr(char const* source)
+        {
+            return source == nullptr ? std::nullopt : std::make_optional<std::string>(source);
+        }
+
+        std::string getHostname()
```
KimonHoffmann, line 28: Should probably be `[[nodiscard]]`.

lib/fabrics/ofi/src/internal/FabricInterfaceProbe.cpp
```diff
@@ -0,0 +1,196 @@
...
+#include "FabricInterfaceList.hpp"
+#include "Format.hpp" // IWYU pragma: keep; used in fmt::to_string()
+#include "Provider.hpp"
+#include "ProviderConfig.hpp"
+
+namespace mxl::lib::fabrics::ofi
+{
+    namespace
+    {
+        std::optional<std::string> optStringFromCStr(char const* source)
```
KimonHoffmann, line 22: Should probably be `[[nodiscard]]`.

lib/fabrics/ofi/src/internal/Format.cpp
```diff
@@ -0,0 +1,52 @@
...
+        }
+        if (capStrings.empty())
+        {
+            resultLength += capStrings.emplace_back("<none>").size();
+        }
+
+        resultLength += (capStrings.size() - 1); // separating '|' chars
+
+        auto result = capStrings.front();
+        result.reserve(resultLength + (capStrings.size() - 1));
```
KimonHoffmann: `(capStrings.size() - 1)` has already been added to `resultLength` above.

lib/fabrics/ofi/src/internal/Target.hpp
```diff
@@ -150,6 +150,13 @@ namespace mxl::lib::fabrics::ofi
         std::unique_ptr<TargetInfo> setup(mxlFabricsTargetConfig const& config, TargetSetupOptions const& options = {});
 
     private:
+        /** \brief Set up the correct concrete target type internally and returns the target info
+         * \param config Target configuration passed by the user.
+         * \param info Fabric info already resolved from the interface config.
+         */
+        template<typename TargetT>
+        std::unique_ptr<TargetInfo> setup(mxlFabricsTargetConfig const& config, FabricInfoView info, TargetSetupOptions const& options = {});
```
KimonHoffmann, line 160: Should probably be `[[nodiscard]]`.

lib/fabrics/ofi/src/internal/Target.hpp
```diff
@@ -150,6 +150,13 @@ namespace mxl::lib::fabrics::ofi
         std::unique_ptr<TargetInfo> setup(mxlFabricsTargetConfig const& config, TargetSetupOptions const& options = {});
```
KimonHoffmann, line 151: Should probably be `[[nodiscard]]`.

lib/tests/fabrics/ofi/test_Address.cpp
```diff
@@ -18,7 +74,7 @@ TEST_CASE("ofi: FabricAddress default construction", "[ofi][FabricAddress]")
 
 TEST_CASE("ofi: FabricAddress base64 decode", "[ofi][FabricAddress]")
 {
-    FabricAddress addr = FabricAddress::fromBase64("AQIDBAU="); // base64 for {1,2,3,4,5}
+    RawFabricAddress addr = RawFabricAddress::fromBase64("AQIDBAU=", FabricAddressFormat::Unspec); // base64 for {1,2,3,4,5}
```
KimonHoffmann: `auto (const)`?!

lib/tests/fabrics/ofi/test_Address.cpp
```diff
@@ -2,14 +2,70 @@
...
+        std::memcpy(buf.data() + 16, &qpn, sizeof(qpn));
+        std::memcpy(buf.data() + 20, &qkey, sizeof(qkey));
+        return buf;
+    }
+}
+
 TEST_CASE("ofi: FabricAddress default construction", "[ofi][FabricAddress]")
 {
-    FabricAddress empty;
+    RawFabricAddress empty;
```
KimonHoffmann: AA style

lib/tests/fabrics/test_interfaces.cpp
```diff
@@ -0,0 +1,283 @@
...
+namespace
+{
+    struct InterfaceListStats
+    {
+        int total = 0;
+        int tcp = 0;
+        int shm = 0;
+    };
+
+    InterfaceListStats countInterfaces(mxlFabricsInterfaceList const* list)
```
KimonHoffmann, line 22: `[[nodiscard]]`

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -25,37 +26,156 @@
...
+    }
+    if (capStrings.empty())
+    {
+        resultLength += capStrings.emplace_back("<none>").size();
+    }
+
+    resultLength += (capStrings.size() - 1); // separating '|' chars
+
+    auto result = capStrings.front();
+    result.reserve(resultLength + (capStrings.size() - 1));
```
KimonHoffmann: `(capStrings.size() - 1)` has already been added to `resultLength` above.

tools/mxl-fabrics-info/main.cpp
```diff
@@ -0,0 +1,272 @@
...
+        }
+        if (capStrings.empty())
+        {
+            resultLength += capStrings.emplace_back("<none>").size();
+        }
+
+        resultLength += (capStrings.size() - 1); // separating '|' chars
+
+        auto result = capStrings.front();
+        result.reserve(resultLength + (capStrings.size() - 1));
```
KimonHoffmann: As above `(capStrings.size() - 1)` has already been added.

KimonHoffmann (changes requested):
A few more things. I also unresolved a few of the old ones that remained unaddressed.

jonasohland:
@KimonHoffmann Thanks again for the thorough review! I hope that I have addressed everything this time. I greatly appreciate the conscientiousness.

k0zmo (approved):

vt-tv (approved):
Approving on behalf of @k0zmo

KimonHoffmann (approved):
Thank you for making all the changes!
There a re a few potential opportunities for `[[nodiscard]]` annotations left in `demo.cpp`, but I'll stop bothering you now. 😉

PR #566, opened 2026-06-16
Pack audio RMA samples in channel-major order

### LINKED ISSUE #565: Audio RMA sample transfers can corrupt multi-channel audio
Audio RMA sample transfers can corrupt multi-channel audio when the source and target ring buffers wrap at different fragment boundaries.

The sender builds the payload from a scatter/gather list in source-ring order: fragment first, channel second. The receiver rebuilds the samples in `AudioBounceBuffer::unpack()`, but it walks the target ring instead. If the target wraps at a different point, `unpack()` splits the stream using different per-channel sample counts than the sender used.

This usually stays hidden when the buffers have the same length/layout or when the test data is single-channel. It shows up with multi-channel audio when both sides wrap differently, for example:

```
source per-channel split: 2 + 3 samples
target per-channel split: 3 + 2 samples
```

The old sender payload is grouped like this:

```
src fragment 0: ch0 first 2, ch1 first 2
src fragment 1: ch0 next 3, ch1 next 3
```

but the receiver unpacks it like this:

```
target fragment 0 expects 3 per channel
target fragment 1 expects 2 per channel
```

The byte count and transfer completion still look correct. The payload is the right size but the samples are read back in the wrong order.

k0zmo:
# Details

Fixes #565. Adds a test testing only the `unpack` part as `makeScatterGatherList` helper function is private.

# Backport requirements

Should be applied to v1.1 as well.

### DIFF
--- a/lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
+++ b/lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
@@ -106,6 +106,8 @@ namespace mxl::lib::fabrics::ofi
             slices);
 
         auto srcAddr = entry.samples();
+        // The egress side writes the payload in channel-major order: all fragments for one channel are contiguous before the next channel.
+        // Consume fragments in that same order so wrapped slices do not interleave channel data.
         for (auto chan = std::size_t{0}; chan < slices.count; chan++)
         {
             for (auto const& fragment : slices.base.fragments)
--- a/lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
+++ b/lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
@@ -106,12 +106,12 @@ namespace mxl::lib::fabrics::ofi
             slices);
 
         auto srcAddr = entry.samples();
-        for (auto& fragment : slices.base.fragments)
+        for (auto chan = std::size_t{0}; chan < slices.count; chan++)
         {
-            // check if the fragment present
-            if (fragment.size > 0)
+            for (auto const& fragment : slices.base.fragments)
             {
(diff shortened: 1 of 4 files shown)


vt-tv (approved):
LGTM. Thanks for your contribution.

lib/fabrics/ofi/src/internal/AudioBounceBuffer.cpp
```diff
@@ -106,12 +106,12 @@ namespace mxl::lib::fabrics::ofi
             slices);
 
         auto srcAddr = entry.samples();
-        for (auto& fragment : slices.base.fragments)
```
jonasohland, line 109: It would be great if you could leave a comment here to explain this issue and why we need to pack by channel first.

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
```diff
@@ -204,12 +204,12 @@ namespace mxl::lib::fabrics::ofi
         // additional one per channel if 2 fragments are present (wrap-around). When a fragment is not present its size will be 0.
         auto sgList = std::vector<LocalRegion>{};
         sgList.reserve(sgListLen);
-        for (auto& fragment : slice.base.fragments)
+        for (auto chan = std::size_t{0}; chan < slice.count; chan++)
```
jonasohland, line 211: And the exact same comment here.

mlefebvre1 (approved):
Thank you for this fix.

jonasohland (approved):
Only the last commit is missing a sign-off.

jonasohland:
@k0zmo Thanks a lot for this fix! That would probably have taken a while for me to figure out...

PR #569, opened 2026-06-18
Fix audio testsrc busy-spinning at 100% CPU

jonasohland:
When pull() returned no sample and the next batch time had already passed, mxlGetNsUntilIndex() returned 0, making the next pull() non-blocking. With no sleep on the empty-pull path this caused the loop to spin. Clamp the pull timeout to at least one batch period so the thread always blocks between retries.

# Backport requirements
Would make sense to backport to v1.1


vt-tv (approved):
Thanks for fixing this.  it was driving me nuts!

pedro-alves-ferreira (approved):
LGTM

PR #574, opened 2026-06-20
fix(fabrics/ofi): make endpoint ids collision-free

dheinrich-qvest-digital:
## What

Make `Endpoint::randomId()` (Endpoint.cpp) collision-free under concurrent setup: seed one 64-bit engine **once** from full-width entropy and serialize draws, instead of reseeding a fresh `std::mt19937` from a single 32-bit `std::random_device` draw on every call.

## Why (root cause)

`Endpoint::Id` is a process-local dispatch key: an initiator routes every connection/completion event to the correct target by this id (RCInitiator's id-keyed `_targets` map — `find(idFromToken(...))` / `find(idFromFID(...))`). The id is published in `TargetInfo` (`RCTarget::setup`) and reused as the accepted endpoint's id.

The old `randomId()`:

```cpp
std::uniform_int_distribution<Endpoint::Id> dist;
std::random_device rd;
std::mt19937 eng{rd()};   // fresh engine, ONE 32-bit time-correlated seed, per call
return dist(eng);
```

When several targets are set up in the same tight window (a gateway bringing up many mirrors at once), two `randomId()` calls can draw the same 32-bit, time-correlated seed → the same id. The initiator's id-keyed map then collapses the two endpoints, and one target's connection events are delivered to the wrong endpoint → that mirror never connects and stays stuck forever.

This is a birthday collision on the 32-bit **seed** (not the 64-bit output): with 1 target it never happens; with ~9 concurrent setups it happens ~1/3 of the time — exactly the observed intermittency.

## Impact

Fixes an intermittent MxlFlowMirror 0/N "stuck forever" wedge observed in the mxl-k8s gateway when many cross-node mirrors are established concurrently (TCP provider). 1 mirror = always works; 9 mirrors = ~1/3 wedge.

## Notes

`randomId()` runs only at setup/reconnect, never on the steady-state hot path — the static `std::mutex` guards concurrent generation on the shared engine and is uncontended in normal operation.

> Earlier revisions also serialized `fi_getinfo`; **dropped after review** — `fi_ini` is already guarded by libfabric's `common_locks.ini_lock` and `fi_getinfo` is documented thread-safe, so that change was removed. This PR is now scoped to the endpoint-id fix only.

### DIFF
--- a/lib/fabrics/ofi/src/internal/Endpoint.cpp
+++ b/lib/fabrics/ofi/src/internal/Endpoint.cpp
@@ -33,12 +33,7 @@ namespace mxl::lib::fabrics::ofi
     {
         // Use a global static random engine to prevent endpoint id collisions.
         static auto mu = std::mutex{};
-        static auto engine = []
-        {
-            std::random_device rd;
-            std::seed_seq seq{rd(), rd(), rd(), rd(), rd(), rd(), rd(), rd()};
-            return std::mt19937_64{seq};
-        }();
+        static auto engine = std::mt19937_64{std::random_device{}()};
         auto const lock = std::scoped_lock{mu};
         return std::uniform_int_distribution<Endpoint::Id>{}(engine);
     }
--- a/lib/fabrics/ofi/src/internal/Endpoint.cpp
+++ b/lib/fabrics/ofi/src/internal/Endpoint.cpp
@@ -31,30 +31,15 @@ namespace mxl::lib::fabrics::ofi
 {
     Endpoint::Id Endpoint::randomId() noexcept
     {
-        // Endpoint::Id is a process-local dispatch key: an initiator routes
-        // connection and completion events to the correct target by this id
-        // (see RCInitiator's id-keyed _targets map). It must be unique across
-        // every endpoint a single initiator talks to; it is not a wire value.
-        //
-        // The previous implementation reseeded a fresh std::mt19937 from ONE
-        // 32-bit, time-correlated std::random_device draw on every call. Targets
-        // set up in the same tight window (e.g. a gateway bringing up many
-        // mirrors at once) could draw the same 32-bit seed -> the same id; the
-        // initiator's id-keyed map then collapsed two endpoints and one mirror
-        // never connected -- an intermittent wedge that scaled with the number
-        // of concurrent setups.
-        //
-        // Seed one 64-bit engine ONCE from full-width entropy and serialize the
-        // draws. randomId() runs only at setup/reconnect, never on the hot path,
-        // so the lock is uncontended in steady state.
-        static std::mutex mu;
-        static std::mt19937_64 engine = []
+        // Use a global static random engine to prevent endpoint id collisions.
+        static auto mu = std::mutex{};
+        static auto engine = []
         {
             std::random_device rd;
             std::seed_seq seq{rd(), rd(), rd(), rd(), rd(), rd(), rd(), rd()};
             return std::mt19937_64{seq};
         }();
-        std::scoped_lock const lock{mu};
+        auto const lock = std::scoped_lock{mu};
         return std::uniform_int_distribution<Endpoint::Id>{}(engine);
     }
 
--- a/lib/fabrics/ofi/src/internal/Endpoint.cpp
+++ b/lib/fabrics/ofi/src/internal/Endpoint.cpp
@@ -6,7 +6,9 @@
 #include <cstddef>
 #include <cstdint>
 #include <memory>
+#include <mutex>
 #include <optional>
+#include <random>
 #include <utility>
 #include <uuid.h>
 #include <sys/uio.h>
@@ -29,11 +31,31 @@ namespace mxl::lib::fabrics::ofi
 {
     Endpoint::Id Endpoint::randomId() noexcept
     {
-        std::uniform_int_distribution<Endpoint::Id> dist;
-        std::random_device rd;
-        std::mt19937 eng{rd()};
-
-        return dist(eng);
+        // Endpoint::Id is a process-local dispatch key: an initiator routes
+        // connection and completion events to the correct target by this id
+        // (see RCInitiator's id-keyed _targets map). It must be unique across
+        // every endpoint a single initiator talks to; it is not a wire value.
+        //
+        // The previous implementation reseeded a fresh std::mt19937 from ONE
+        // 32-bit, time-correlated std::random_device draw on every call. Targets
+        // set up in the same tight window (e.g. a gateway bringing up many
+        // mirrors at once) could draw the same 32-bit seed -> the same id; the
+        // initiator's id-keyed map then collapsed two endpoints and one mirror
+        // never connected -- an intermittent wedge that scaled with the number
+        // of concurrent setups.
+        //
+        // Seed one 64-bit engine ONCE from full-width entropy and serialize the
+        // draws. randomId() runs only at setup/reconnect, never on the hot path,
+        // so the lock is uncontended in steady state.
+        static std::mutex mu;
+        static std::mt19937_64 engine = []
+        {
+            std::random_device rd;
+            std::seed_seq seq{rd(), rd(), rd(), rd(), rd(), rd(), rd(), rd()};
+            return std::mt19937_64{seq};
+        }();
+        std::scoped_lock const lock{mu};
+        return std::uniform_int_distribution<Endpoint::Id>{}(engine);
     }
 
     Endpoint::Id Endpoint::idFromFID(::fid_t fid) noexcept
--- a/lib/fabrics/ofi/src/internal/FabricInfo.cpp
+++ b/lib/fabrics/ofi/src/internal/FabricInfo.cpp
@@ -174,13 +174,10 @@ namespace mxl::lib::fabrics::ofi
 
         // hints: add condition to append FI_HMEM capability if needed!
 
-        // libfabric performs lazy, process-global provider initialization on the
-        // first fi_getinfo. Concurrent first calls from several target/initiator
-        // setups can race that init and intermittently fail (or hand back a
-        // malformed fi_info) for one caller. fi_getinfo only runs at setup, so
-        // serialize it -- the steady-state hot path never touches this.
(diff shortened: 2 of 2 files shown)


lib/fabrics/ofi/src/internal/Endpoint.cpp
```diff
@@ -29,11 +31,31 @@ namespace mxl::lib::fabrics::ofi
 {
     Endpoint::Id Endpoint::randomId() noexcept
     {
-        std::uniform_int_distribution<Endpoint::Id> dist;
-        std::random_device rd;
-        std::mt19937 eng{rd()};
-
-        return dist(eng);
+        // Endpoint::Id is a process-local dispatch key: an initiator routes
```
jonasohland: Please condense this comment into what is really important: Use global static random engine to prevent endpoint id collisions. We don't need this LLM wall of text in the codebase.

lib/fabrics/ofi/src/internal/Endpoint.cpp
```diff
@@ -29,11 +31,31 @@ namespace mxl::lib::fabrics::ofi
...
+        // set up in the same tight window (e.g. a gateway bringing up many
+        // mirrors at once) could draw the same 32-bit seed -> the same id; the
+        // initiator's id-keyed map then collapsed two endpoints and one mirror
+        // never connected -- an intermittent wedge that scaled with the number
+        // of concurrent setups.
+        //
+        // Seed one 64-bit engine ONCE from full-width entropy and serialize the
+        // draws. randomId() runs only at setup/reconnect, never on the hot path,
+        // so the lock is uncontended in steady state.
+        static std::mutex mu;
```
jonasohland: Some of this code previously did not adhere to the "auto left, type right" style that we are trying to adopt across the whole codebase.
This would be great occasion to change that.
So to be consistent let's change this to: `auto mu = std::mutex{}`.

lib/fabrics/ofi/src/internal/Endpoint.cpp
```diff
@@ -29,11 +31,31 @@ namespace mxl::lib::fabrics::ofi
...
+        static std::mt19937_64 engine = []
+        {
+            std::random_device rd;
+            std::seed_seq seq{rd(), rd(), rd(), rd(), rd(), rd(), rd(), rd()};
+            return std::mt19937_64{seq};
+        }();
+        std::scoped_lock const lock{mu};
```
jonasohland:
```suggestion
        auto const lock = std::scoped_lock{mu};
```

lib/fabrics/ofi/src/internal/FabricInfo.cpp
```diff
@@ -173,6 +174,14 @@ namespace mxl::lib::fabrics::ofi
 
         // hints: add condition to append FI_HMEM capability if needed!
 
+        // libfabric performs lazy, process-global provider initialization on the
+        // first fi_getinfo. Concurrent first calls from several target/initiator
+        // setups can race that init and intermittently fail (or hand back a
+        // malformed fi_info) for one caller. fi_getinfo only runs at setup, so
+        // serialize it -- the steady-state hot path never touches this.
+        static std::mutex getInfoMutex;
```
jonasohland:
```suggestion
        static auto getInfoMutex = std::mutex{};
```

lib/fabrics/ofi/src/internal/FabricInfo.cpp
```diff
@@ -173,6 +174,14 @@ namespace mxl::lib::fabrics::ofi
...
+        std::scoped_lock const getInfoLock{getInfoMutex};
```
jonasohland:
```suggestion
        auto const getInfoLock = std::scoped_lock{getInfoMutex};
```

lib/fabrics/ofi/src/internal/FabricInfo.cpp
```diff
@@ -173,6 +174,14 @@ namespace mxl::lib::fabrics::ofi
 
         // hints: add condition to append FI_HMEM capability if needed!
 
+        // libfabric performs lazy, process-global provider initialization on the
```
jonasohland: Please condense to something like: "Protect libfabric internal process-global initialization in ::fi_getinfo"

jonasohland (changes requested):
Hello! Thanks for your PR. Only style-related comments.

jonasohland:
@dheinrich-qvest-digital Can you point me to where that unprotected global init happens in libfabric? I had a quick look through the libfabric code and it looks like `fi_ini()` which is called right at the beginning of `fi_getinfo()` and seems to be responsible for the provider init, is protected by a mutex `common_locks.ini_lock`.
So I think `fi_getinfo` is intended to be safe to be called concurrently, and either this is a libfabric bug or not and issue.

`fi_ini`: https://github.com/ofiwg/libfabric/blob/24d63193433d975a264fa00476e4eb9f853e429f/src/fabric.c#L923
`fi_getinfo`: https://github.com/ofiwg/libfabric/blob/24d63193433d975a264fa00476e4eb9f853e429f/src/fabric.c#L1329

dheinrich-qvest-digital:
@jonasohland You're right that `fi_ini()` is guarded by `common_locks.ini_lock`, so the one-time provider init is safe — thanks, I've corrected the comment.

My concern is the step after: once `fi_ini()` returns, `fi_getinfo` iterates `prov_head` calling each `prov->provider->getinfo()` **without holding any lock**, and that per-provider path (EFA in our case) isn't guaranteed reentrant. The mutex only runs at setup/reconnect, so it's uncontended — I kept it as cheap defensive serialization, not a claim of a libfabric core bug. Happy to drop it if you'd prefer we not guard provider-specific paths here.

I've also pushed the style fixes (condensed comments + auto-left declarations). PTAL.

jonasohland:
> @jonasohland You're right that `fi_ini()` is guarded by `common_locks.ini_lock`, so the one-time provider init is safe — thanks, I've corrected the comment.
> 
> My concern is the step after: once `fi_ini()` returns, `fi_getinfo` iterates `prov_head` calling each `prov->provider->getinfo()` **without holding any lock**, and that per-provider path (EFA in our case) isn't guaranteed reentrant. The mutex only runs at setup/reconnect, so it's uncontended — I kept it as cheap defensive serialization, not a claim of a libfabric core bug. Happy to drop it if you'd prefer we not guard provider-specific paths here.
> 
> I've also pushed the style fixes (condensed comments + auto-left declarations). PTAL.

I mean either calling `fi_getinfo` is reentrant or it is not. And in my opinion the interface and docs suggest that it is. So if this contract breaks when a specific provider is compiled in then it is a bug.

Also: I can't find anything that mutates global state in `efa_getinfo` except for a section in `efa_get_userinfo` that is guarded by a mutex.

lib/fabrics/ofi/src/internal/Endpoint.cpp
```diff
@@ -29,11 +31,16 @@ namespace mxl::lib::fabrics::ofi
...
     Endpoint::Id Endpoint::randomId() noexcept
     {
-        std::uniform_int_distribution<Endpoint::Id> dist;
-        std::random_device rd;
-        std::mt19937 eng{rd()};
-
-        return dist(eng);
+        // Use a global static random engine to prevent endpoint id collisions.
+        static auto mu = std::mutex{};
+        static auto engine = []
```
jonasohland: For some reason this comment did not make it through in my last review:

With a global singleton random engine, this section can probably be simplified to:
```
static auto engine = std::mt19937_64{std::random_device{}()};
```

jonasohland (changes requested):

dheinrich-qvest-digital:
You're right, @jonasohland — I've dropped the `fi_getinfo` serialization.

`fi_ini()` is guarded by `common_locks.ini_lock` and `fi_getinfo` is documented as thread-safe, and I couldn't substantiate an unguarded global-state path in `efa_getinfo` beyond the `efa_get_userinfo` section that's already mutex-protected. So the extra lock was guarding a non-bug. If a provider ever broke that documented reentrancy contract, that's a libfabric bug to report upstream rather than something to paper over here. `FabricInfo.cpp` is back to baseline.

I also applied the rng simplification — `static auto engine = std::mt19937_64{std::random_device{}()};` (the mutex stays; it guards concurrent generation). The PR is now just the endpoint-id collision fix. PTAL.

lib/fabrics/ofi/src/internal/Endpoint.cpp
dheinrich-qvest-digital: fixed

dheinrich-qvest-digital:
@jonasohland Addressed all style feedback in 6dc0013 + c551275 — comment condensed, auto-left style, rng simplified, fi_getinfo serialization dropped. Ready for another look.

jonasohland (approved):
Approved.

A note for the future: Our contribution guidelines outline the requirements contributors have to adhere to when it comes to AI assisted contributions:

> A contribution to the project must be reviewed and understood in its entirety by the contributor. AI assisted contributions are allowed but the rule remains the same: the contributor must understand every line of the contribution and should be able to justify all of them.

vt-tv (approved):
lgtm

vt-tv:
> Backport failed for `release/v1.1`, because it was unable to cherry-pick the commit(s).
> 
> Please cherry-pick the changes locally and resolve any conflicts.
> 
> ```shell
> git fetch origin release/v1.1
> git worktree add -d .worktree/backport/574/-/release/v1.1 origin/release/v1.1
> cd .worktree/backport/574/-/release/v1.1
> git switch --create backport/574/-/release/v1.1
> git cherry-pick -x d3771a47a533ed579be023b659ecedc7ae82d711 93d62c0ab6ac0952978c9c33a99ae718d396f29b 7234f9a68bafe6743c5e1f6aa9c431c0ad4a0bdf 45f49fe7148aef7e721862b500030c1e2a6d0d9c
> ```

@dheinrich-qvest-digital. could you please cherry pick that PR onto the 1.1 release branch ?  Thanks!

dheinrich-qvest-digital:
@vt-tv Thanks for checking. Done.

https://github.com/dmf-mxl/mxl/pull/584

Please let me know if you need anything else.

PR #579, opened 2026-06-24
Configurable target CQ depth + batch grain drain for Fabrics

moleksy:
Two related improvements to the MXL Fabrics receive path that make draining the completion queue (CQ) safer and easier for library users:

1) Configurable target CQ depth - the receiver's completion queue size is now a tunable option instead of a fixed default, so applications that burst many grains can size the CQ to avoid overflow.

2) Batch grain drain primitive - a new mxlFabricsTargetReadGrainsNonBlocking drains up to a caller-specified number of available grain completions in a single call, which is the recommended building block for a two-phase (drain -> process) receive loop.

Together these changes give users control over CQ capacity and a one-call drain that keeps the CQ empty even when per-grain processing temporarily falls behind - the backlog then moves into the application's own (unbounded) queue rather than overflowing the completion queue.


lib/fabrics/ofi/src/fabrics.cpp
```diff
@@ -23,6 +24,47 @@
 #include "mxl/flow.h"
 #include "mxl/platform.h"
 
+namespace
+{
+    /// Parse the optional target setup options JSON and return the requested completion
+    /// queue depth, or 0 (use the implementation default) when not specified.
+    ///
+    /// Recognized form: {"cqDepth": <positive integer>}
+    std::size_t parseCqDepthOption(char const* options)
```
mlefebvre1: Please return an optional instead of 0 when the option is not present.

lib/fabrics/ofi/src/fabrics.cpp
```diff
@@ -207,6 +250,37 @@ mxlStatus mxlFabricsTargetReadGrain(mxlFabricsTarget in_target, uint16_t in_time
         "Failed to wait for new grain");
 }
 
+extern "C" MXL_EXPORT
+mxlStatus mxlFabricsTargetReadGrainsNonBlocking(mxlFabricsTarget in_target, uint64_t* out_grainIndices, size_t in_maxCount, size_t* out_count)
```
mlefebvre1: I don't see the point of adding this new function as this can easily be implemented using the existing API. We should strive to make the API as minimal and simple as possible.

This function could make sense if the `count` argument of `fi_cq_read` and `fi_cq_sread` was used.

lib/fabrics/ofi/src/internal/RCTarget.hpp
```diff
@@ -29,7 +29,7 @@ namespace mxl::lib::fabrics::ofi
          * \return A pair consisting of the newly setup RCTarget and its associated TargetInfo.
          */
         [[nodiscard]]
-        static std::pair<std::unique_ptr<RCTarget>, std::unique_ptr<TargetInfo>> setup(mxlFabricsTargetConfig const& config);
+        static std::pair<std::unique_ptr<RCTarget>, std::unique_ptr<TargetInfo>> setup(mxlFabricsTargetConfig const& config, std::size_t cqDepth = 0);
```
mlefebvre1: I would make `cqDepth` part of a config struct (to establish the pattern). Otherwise, before we know it, this function will have 10 arguments. Same applies to others.

lib/fabrics/ofi/src/internal/RCTarget.cpp
```diff
@@ -63,21 +63,24 @@ namespace mxl::lib::fabrics::ofi
         auto proto = selectIngressProtocol(mxlRegions.dataLayout(), mxlRegions.regions(), mxlRegions.maxSyncBatchSize());
         auto targetInfo = std::make_unique<TargetInfo>(pep.id(), pep.localAddress(), proto->registerMemory(domain), proto->bounceBufferInfo());
 
+        auto const cqSize = (cqDepth != 0) ? cqDepth : CompletionQueue::Attributes::DEFAULT_SIZE;
```
mlefebvre1: use `std::optional` instead (and everywhere it applies)

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -612,54 +612,85 @@ class AppTarget
     mxlStatus runDiscrete()
     {
         mxlGrainInfo grainInfo;
-        std::uint64_t grainIndex = 0;
         std::uint8_t* dummyPayload;
-        mxlStatus status;
+
+        // Phase 1 drains up to this many completions per iteration before doing the (potentially slower) per-grain commit work in Phase 2.
+        // The drained indices are buffered here so the completion queue is emptied promptly even when Phase 2 temporarily falls behind.
+        constexpr std::size_t maxDrainPerIteration = 64;
```
mlefebvre1: Why does the demo code needs to do this now ? It is just a proxy, it does not do any processing. If it is only for feature demonstration purposes, then I would add a comment mentioning that it is not required for this specific example, but could be useful for other applications (and refer to the docs).

mlefebvre1 (changes requested):

lib/fabrics/ofi/src/fabrics.cpp
```diff
@@ -23,6 +25,47 @@
...
 
+namespace
+{
+    /// Parse the optional target setup options JSON and return the requested completion
+    /// queue depth, or std::nullopt (use the implementation default) when not specified.
+    ///
+    /// Recognized form: {"cqDepth": <positive integer>}
+    std::optional<std::size_t> parseCqDepthOption(char const* options)
+    {
+        using mxl::lib::fabrics::ofi::Exception;
```
jonasohland: Please move this function into the existing anonymous namespace a few lines down that also contains the `try_run` function.

jonasohland (approved):

lib/fabrics/ofi/src/internal/RCTarget.hpp
```diff
@@ -105,6 +108,7 @@ namespace mxl::lib::fabrics::ofi
     private:
         std::unique_ptr<IngressProtocol> _proto;
         std::shared_ptr<Domain> _domain;
-        State _state; /**< The current state of the RCTarget. */
+        std::size_t _cqDepth; /**< Completion queue depth for accepted connections. */
```
mlefebvre1: why not use `TargetSetupOptions` all the way through ?

lib/fabrics/include/mxl/fabrics.h
```diff
@@ -156,7 +156,10 @@ extern "C"
      * \param in_target A valid fabrics target
      * \param in_config The target configuration. This will be used to create an endpoint and register a memory region. The memory region
      * corresponds to the one that will be written to by the initiator.
-     * \param options A json-formatted string with options. Currently not used.
+     * \param options An optional json-formatted string with options. May be NULL or empty. Recognized fields:
```
mlefebvre1: can we fix the typo while we're here, it should be `\param in_options ...`

moleksy:
I think that this build error needs a repo-wide fix.

crossbeam-epoch 0.9.18 -> 0.9.20 in rust/Cargo.lock

Should I apply it here?

mlefebvre1:
> I think that this build error needs a repo-wide fix.
> 
> crossbeam-epoch 0.9.18 -> 0.9.20 in rust/Cargo.lock
> 
> Should I apply it here?

I would probably create a separate PR to fix this first. The lockfile can be updated with `cargo update -p crossbeam-epoch`.
EDIT: https://github.com/dmf-mxl/mxl/pull/589

mlefebvre1 (approved):

PR #585, opened 2026-07-04
FlowParser: support a video/H264 coded DATA flow (large grains, page slices)

Ali-Al-Abedi:
## Problem

`MXL_DATA_FORMAT_DATA` currently only accepts `media_type == "video/smpte291"` and fixes the grain at `MXL_DATA_FORMAT_GRAIN_SIZE` (4096 B) with 1-byte slices. That is exactly right for ST 2110-40 ANC, but ~35× too small for an H.264 access unit (a 1080p59.94 keyframe is 100–140 KiB), so MXL cannot carry **coded** essence today — only raw video, audio and ANC.

## Change

Adds a second DATA profile for `media_type == "video/H264"`, touching **only** the three size methods (`getPayloadSize`, `getPayloadSliceLengths`, `getTotalPayloadSlices`):

- **Page-sized (4096 B) slices** instead of 1 byte, so a large access-unit grain stays well within the `uint16 validSlices` count (a 512 KiB grain = 128 slices; 1 MiB = 256).
- **Grain payload size read from the descriptor's optional `"grain_size"`** (bytes) — the producer derives it from its resolution/bitrate — rounded up to a whole number of slices, defaulting to 1 MiB. Nothing is hardcoded per stream.

## Backward compatibility

`video/smpte291` and all `video`/`audio` paths are unchanged. The change is purely additive (a new `else if` branch + a small helper), so existing flows behave exactly as before.

### Suggested grain wire format (`video/H264`)

```
[ uint32 LE au_len ][ au_len bytes H.264 Annex-B, alignment=au ][ pad to slice ]
validSlices = ceil((4 + au_len) / 4096)
```

The 4-byte length prefix lets readers recover the exact AU length after the page-granular slice count overshoots it (readers use `mxlFlowReaderGetGrainSlice(minValidSlices=1)` then trust the prefix — the same pattern the ANC reader uses).

## Testing

- Builds against current `main` (`libmxl.so.1.2`).
- Drives a live NVIDIA L4 pipeline: one encoder writes H.264 AUs into a `video/H264` flow and N WebRTC viewers fan out from shared memory with **no per-viewer re-encode**. Measured **16 concurrent 1080p59.94 producers at 82% NVENC / 4.1 GB VRAM with 0 dropped grains**; AU avg ~14 KiB, max ~106 KiB.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/lib/internal/src/FlowParser.cpp
+++ b/lib/internal/src/FlowParser.cpp
@@ -29,6 +29,33 @@ namespace mxl::lib
         constexpr auto MAX_WIDTH = 7680u;  // 8K UHD
         constexpr auto MAX_HEIGHT = 4320u; // 8K UHD
 
+        // Coded/compressed "data" flows (e.g. video/H264 access units) carry a whole
+        // access unit per grain. Unlike ANC (1-byte slices, capped at
+        // MXL_DATA_FORMAT_GRAIN_SIZE), a coded grain must hold a full keyframe (100s of
+        // KB), so:
+        //   - the slice length is a VFS page (4096 B), not a single byte, so a large
+        //     grain stays well within the uint16 validSlices count, and
+        //   - the grain payload size is taken from the flow descriptor's optional
+        //     "grain_size" field (bytes) set by the producer from its resolution /
+        //     bitrate, defaulting to 1 MiB. Nothing is hardcoded per stream.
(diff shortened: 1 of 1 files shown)


garethsb:
See https://github.com/dmf-mxl/mxl/pull/580#issuecomment-4793066926

PR #588, opened 2026-07-07
gst-mxl-rs: video + audio + anc data flow synchronisation

### LINKED ISSUE #240: [Feature] Add ANC support in Gstreamer plugin
### Requirement Issue Link

https://github.com/dmf-mxl/mxl-requirements/issues/129

### Description

To be able to test the support for standard Anciliary Data

### Additional context

_No response_

garethsb:
# MXL video + audio + anc data flow synchronisation

**Status:** implemented
**Scope:** `gst-mxl-rs` (`mxlsrc`, `mxlsink`, `clock`, integration tests), `gst-plugins-rs` (`st2038combiner`)

## 1. Summary

MXL stores **video** (`video/x-raw` / v210) and **ancillary data** (`meta/x-st-2038` on a
separate **data flow**) as discrete grains in a shared index space. GStreamer pipelines that
**split** ancillary from video (`st2038extractor` → two `mxlsink`s) and **rejoin** them (two
`mxlsrc`s → `st2038combiner`) must attach each video frame's own ST-2038 packets — no
off-by-one, no smear.

Correct pairing needs two things, both now in place:

1. **Identical PTS for the same frame on both flows.** `mxlsink` maps a buffer's PTS to an
   **absolute** MXL grain index; `mxlsrc` maps a grain's absolute index back to the **same**
   PTS. Both use one shared MXL-time offset `D` (§4.2) so the same producer frame lands on
   the same absolute index on every flow.
2. **A combiner that waits for the frame's window.** `st2038combiner` holds each picture until
   the end of its ST-2038 running-time window and completes it on the first in-window ANC,
   instead of exiting early on an empty peek (§4.3).

The disjoint and combiner round-trips, the discrete data round-trips, the audio round-trip,
and the timing-contract tests all pass. Producer and consumer sides map every flow to the same
absolute grain index through a single shared MXL-time offset `D`, sampled once against whatever
clock the pipeline selects (§4.2, §6).

### 2.1 Producer (common to the video/data tests)

```text
appsrc (v210 + GstAncillaryMeta)
  ! st2038extractor name=ext remove-ancillary-meta=true
ext.src    ! queue ! mxlsink (video flow)
ext.st2038 ! queue ! mxlsink (data flow)
```

Each frame index is written into the v210 payload and every ancillary payload so tests verify
alignment without parsing ANC semantics.

### 2.2 Consumer patterns

| Test | Pipeline | Purpose |
|------|----------|---------|
| **Disjoint** (`v210_with_meta_to_v210_and_st2038_via_mxl`) | `mxlsrc` (video) → `appsink`; `mxlsrc` (data) → `appsink` | Compare frame index + PTS per grain across flows |
| **Combiner** (`v210_with_meta_to_v210_with_meta_via_mxl`) | `mxlsrc` (video) → `st2038combiner.sink`; `mxlsrc` (data) → `st2038combiner.st2038` → `appsink` | Round-trip v210 + `GstAncillaryMeta` |

The disjoint test is the regression guard for reader timing; the combiner test is the
end-to-end contract that also depends on the `st2038combiner` fix.

## 3. Original root cause

Two failure modes had to be separated:

### 3.1 Timing: relative vs absolute grain index

`mxlsink` writes each buffer to an **absolute** grain index derived from the MXL media
clock, not to indices 0, 1, 2…. The reader anchored at whatever the flow head happened to be
when it attached (`get_current_index()`), read `anchor + frame_counter`, and derived PTS from
that per-reader `frame_counter` plus the pipeline time sampled at attach — jumping back to the
head whenever it fell behind. Two `mxlsrc`s attaching at slightly different instants therefore
anchored at different indices and captured different start times, so for the *same* producer
frame they selected different grains and stamped different PTS — which the combiner, pairing on
running time, cannot reconcile. Anchoring in the shared absolute index space and deriving PTS
from the absolute index (§4.1, §4.2) is what makes the two flows agree per frame.

### 3.2 Combiner pull order and time windows

`st2038combiner` is a stateful aggregator that pairs ANC to video by **running time**, not by
payload or grain index. Its earlier `collect_st2038()` could exit early on an empty ST-2038
peek and output picture *N* with stale ANC, and (once the video end time was known) had no
lower bound so it could also collect ANC from before the picture's window. Matching PTS per
grain is therefore **necessary but not sufficient**: the combiner itself had to hold the
picture until its window closed (§4.3).

### 4.1 Absolute grain-index timing (`mxlsrc/timing.rs`, `create_*`)

- **Anchor** at the flow head (`get_current_index()`) on attach — the same index space
  `mxlsink` writes into.
- Read an **absolute** `read_index`; wait when it is ahead of the flow head.
- **Reader guard:** each read exposes the returned `GrainInfo.index`. If it is *behind* the
  request the slot is a hole → report `NoDataCreated` and retry; if it is *ahead* the ring has
  evicted the request → catch up to the live edge and mark the next buffer `DISCONT`.
- **PTS** comes straight from `pts_for_index(index, D)`, so two readers on the same absolute
  index expose identical timestamps. Audio uses the same rule with the sample index and the
  flow's sample rate.

### 4.2 Offset `D` against the pipeline clock (`clock.rs`)

The MXL elements do **not** provide a pipeline clock. MXL time is `clock_gettime(CLOCK_TAI)`,
a wall clock that can step (NTP, leap seconds, or a hypervisor time sync); pacing `GstBaseSink`
on it stalls the pipeline and then bursts, so a continuous reader overruns the ring. The
pipeline therefore runs on whatever clock it selects — the default monotonic system clock — and
each element only tracks the MXL-time → pipeline-clock offset.

`clock_offset(clock, mxl_now)` returns `D = mxl_now − clock_now`, the constant offset between
MXL (TAI) time and the pipeline clock. `D` is sampled once, recached when the pipeline hands the
element a different clock (`set_clock`), and shared between sibling MXL elements through a
`GstContext` — so two elements on the same clock use one identical `D` instead of two
independently sampled values that could straddle a grain-rounding boundary.

A reader subtracts `D + base_time` from a grain's absolute MXL timestamp to get its PTS; a
writer adds the same to a buffer's PTS to get the MXL timestamp it converts to a grain
index. A grain whose resulting running time would be negative — it predates the consumer joining
the flow — is dropped rather than clamped. Sharing one `D` is what keeps two flows mapping the
same frame to the same index.

### 4.3 `st2038combiner` fix (`gst-plugins-rs` [MR !3125](https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/merge_requests/3125))

- **Hold each picture until the end of its ST-2038 window**; complete it on the first in-window
  ANC rather than breaking early on an empty peek.
- **`drop-late-st2038`** property: drop ANC that misses its picture's window instead of smearing
  it onto a later frame — so each frame carries its own ANC or none.
- The aggregator advances its own segment position with `self.obj().set_position(...)`, so
  `simple_get_next_time()` yields a live deadline instead of a stale `0` (the residual that
  otherwise starved pairing). This is the core fix; no ANC/video duration or frame-rate
  tolerance is needed.

### 4.4 Grain commit (`mxlsink/render_*.rs`)

`mxlsink` is an ordinary `GstBaseSink` with `sync=true`: the base sink waits for each buffer's
running time against the pipeline clock, then `render()` opens the grain at the absolute index
derived from the buffer's MXL timestamp and commits it synchronously. There is no separate
writer thread or pacing loop — the base sink is the single scheduler, so an on-time buffer
commits at its deadline and chaining MXL elements does not accrue a grain period of latency per
hop.

## 5. Approaches ruled out

| Approach | Decision | Why |
|----------|----------|-----|
| `queue` per branch as a sync fix | Ruled out | Decouples threads but does not lock grain index between two `mxlsrc`s; skew remains. Queues stay for thread decoupling only. |
| `multiqueue` before `st2038combiner` | Ruled out | Designed for demuxer fan-out during PAUSED preroll; two independent **live** `PushSrc`s return `NO_PREROLL` and never fill it → deadlock. |
| Avoid split/rejoin in MXL | Not possible | MXL stores ANC on a separate data flow; rejoin via `st2038combiner` is required. |
| MXL-native combiner element | Declined | Duplicates `st2038combiner` with grain-index matching; more code than fixing the upstream combiner. |
| Application-level merge (dual `appsink`) | Declined for production | Useful only as the disjoint **test** pattern. |
| ANC/video duration tolerance in the combiner | Declined | The `set_position` fix (§4.3) removes the residual without needing durations, which are not always known. |

## 6. Producer-side determinism

The producer pipeline has two `mxlsink`s that must map a frame to the **same** absolute grain
index on both flows. Both sinks derive the index from the buffer's PTS plus the shared offset
`D`. Without sharing, each sink would sample `D` at a slightly different instant and could
disagree by one grain on the **first** frame when the difference straddles a grain-rounding
boundary. Sharing `D` through a `GstContext` (§4.2) gives both sinks one identical offset,
removing that edge.

The same context is shared with the consumer-side `mxlsrc`s, so every MXL element in a pipeline
agrees on the mapping regardless of which clock the application selects.

## 7. Tests

All integration tests are **always compiled** (no `#[ignore]`). Each checks its prerequisites
at runtime and prints a libtest-style `... skipped, <reason>` line (via `tests/common`'s
`skip_reason` + `skip!`) when MXL `/dev/shm` domains or required element factories are absent,
so CI surfaces them as notices instead of hiding them.

| Test file | Cases | Covers |
|-----------|-------|--------|
| `video_data_sync.rs` | disjoint, combiner | Per-frame PTS/content alignment across flows; combiner round-trip |
| `data_round_trip.rs` | `st2038_round_trip`, `cea608_round_trip` | Discrete ST-2038 byte round-trip; CEA-608 caption round-trip |
| `audio_round_trip.rs` | `f32le_tone_round_trip` | Continuous audio path (clock/offset) via a 440 Hz tone |
| `mxlsrc_timing.rs` | lag catch-up | Reader contract: across a lag catch-up, a one-frame PTS step ⇒ a one-grain content step |

Run (from `mxl/rust`, following `~/.cursor/rules/mxl-testing.mdc`):

```bash
export CARGO_TARGET_DIR=/home/garethsb/git/mxl/rust/target
export GST_PLUGIN_PATH=/home/garethsb/git/mxl/rust/target/release:/home/garethsb/git/gst-plugins-rs/target/x86_64-unknown-linux-gnu/debug
unset LD_LIBRARY_PATH   # the test binaries find libmxl.so via rpath
cargo test --release -p gst-mxl-rs --test video_data_sync -- --test-threads=1
```

The combiner cases additionally need `st2038combiner` (and CEA-608 needs the closed-caption
chain) from `gst-plugin-closedcaption` on `GST_PLUGIN_PATH`; missing factories skip cleanly.

## 8. References

| Resource | Location |
|----------|----------|
| Reader timing | `mxl/rust/gst-mxl-rs/src/mxlsrc/timing.rs`, `create_video.rs`, `create_data.rs`, `create_audio.rs` |
| Writer grain commit | `mxl/rust/gst-mxl-rs/src/mxlsink/render_*.rs`, `state.rs` |
| MXL-time offset `D` | `mxl/rust/gst-mxl-rs/src/clock.rs` |
| `st2038combiner` | `gst-plugins-rs/video/closedcaption/src/st2038combiner/imp.rs` |
| Integration tests + shared helpers | `mxl/rust/gst-mxl-rs/tests/` (`common/mod.rs`) |
| GStreamer live sources | https://gstreamer.freedesktop.org/documentation/additional/design/live-source.html |


garethsb:
Note: assuming it happens in the next day or two, I'd prefer to wait until https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/merge_requests/3125 is merged and we can bump the gst-plugins-rs dependency version to use it (already well-tested locally).

SimaoPFonseca:
Hi! I have a bone to pick with the use of `unwrap()` and `expect()` in this PR. We generally avoid using them in production code and prefer to propagate errors with `?` instead, as we do throughout the rest of the codebase. Could we do the same here? Test code can be the exception, where using `unwrap()`/`expect()` is acceptable.

pedro-alves-ferreira:
> Hi! I have a bone to pick with the use of `unwrap()` and `expect()` in this PR. We generally avoid using them in production code and prefer to propagate errors with `?` instead, as we do throughout the rest of the codebase. Could we do the same here? Test code can be the exception, where using `unwrap()`/`expect()` is acceptable.

Yes, excellent point. I think one of the Rust community's biggest missteps was normalizing `unwrap()`/`expect()` in examples. AI tooling has certainly picked up that pattern.

garethsb:
@pedro-alves-ferreira @SimaoPFonseca sorry, yes, definitely makes sense to avoid `unwrap`/`expect` when validating arguments etc. I assume it is OK to use them for internal/other library contract invariants that we couldn't recover from - feels like those are better handled at a high level with `catch_unwind` if at all?

pedro-alves-ferreira:
> @pedro-alves-ferreira @SimaoPFonseca sorry, yes, definitely makes sense to avoid `unwrap`/`expect` when validating arguments etc. I assume it is OK to use them for internal/other library contract invariants that we couldn't recover from - feels like those are better handled at a high level with `catch_unwind` if at all?

For internal invariants, I would say yes, but... 
I did that a few times and I think I ended up always going back to another approach, like transitioning to an error state. For example, if you find such a case, you could stop processing data and send an error to the pipeline.

Otherwise, it is very annoying to the application writer. As an example, there's a library we use in several projects that insists in panicking if they can't initialize a piece of hardware, or if there is an inconsistency between their expected state and the hardware's. I ended up patching their code not to do that, because we certainly can recover from that at the application level, by restarting it, rather than just crashing the application. But I had to change hundreds of lines of their code, which they wouldn't upstream, so it's a maintenance nightmare.

The pattern we use is that anything that can fail, even if it is only because of an invariant check, returns a `Result<>`. Then, we let that bubble up to the caller. That could be the application, which is fine, or a library callback. If it's the latter and it doesn't accept an error, then we transition to an error state.

garethsb:
@pedro-alves-ferreira hopefully that's better - as an extra commit to make review easier

pedro-alves-ferreira:
> @pedro-alves-ferreira hopefully that's better - as an extra commit to make review easier

Much better. Thanks!

garethsb:
Do you think the ones that are just protection against a poisoned lock are necessary?

```rust
self.state.lock().unwrap()
```

pedro-alves-ferreira:
I think we can get away with that as it is. There's a simple way to work around that. I'll explain later.

rust/gst-mxl-rs/src/mxlsrc/create_audio.rs
```diff
@@ -283,19 +229,63 @@ fn interleave_audio(samples: &SamplesData<'_>) -> Result<Vec<u8>, gst::FlowError
...
+    }
+}
+
+#[cfg(test)]
+mod pts_tests {
+    use super::define_cushion;
+
+    /// Correct audio PTS uses the sample index read, not `batch_counter`.
+    #[test]
+    fn correct_audio_pts_uses_sample_index_not_batch_counter() {
```
pedro-alves-ferreira: What is the goal of this test? Was it just used to test the formula? Should it be left here as documentation?

rust/gst-mxl-rs/src/mxlsrc/create_audio.rs
```diff
@@ -283,19 +229,63 @@ fn interleave_audio(samples: &SamplesData<'_>) -> Result<Vec<u8>, gst::FlowError
...
+        assert_eq!(correct_pts, 100_000_000);
+    }
+
+    #[test]
+    fn catch_up_jumps_sample_index_and_requires_discont() {
+        let head = 500u64;
+        let batch = 48u64;
+        let target = define_cushion(head, batch);
+        assert_eq!(target, 404);
+        assert!(
```
pedro-alves-ferreira: Given the previous assertion, this is always true.

rust/gst-mxl-rs/src/mxlsrc/imp.rs
```diff
@@ -429,9 +474,28 @@ impl BaseSrcImpl for MxlSrc {
...
+        // instance never waits for the flow, so it is safe on the state-change
+        // thread. The reader attach still runs from `negotiate()`. If the domain
+        // is not set yet, `negotiate()` creates the instance instead.
+        let have_domain = {
+            let settings = self.settings.lock().map_err(|e| {
+                gst::error_msg!(gst::CoreError::Failed, ["Failed to lock settings: {}", e])
+            })?;
+            !settings.domain.is_empty()
+        };
+        if have_domain {
```
pedro-alves-ferreira, line 480: I guess this isn't a race condition because `ensure_instance` is race-free, and if `have_domain` is false while we were holding the lock, then `negotiate` will sort it out, correct?

rust/gst-mxl-rs/src/mxlsrc/create_audio.rs
```diff
@@ -283,19 +233,63 @@ fn interleave_audio(samples: &SamplesData<'_>) -> Result<Vec<u8>, gst::FlowError
...
+    }
+
+    #[test]
+    fn oldest_valid_sample_index_formula() {
+        let ring = 480u64;
+        let batch = 48u64;
+        let head = 500u64;
+        let oldest_valid = head.saturating_sub(ring.saturating_sub(batch));
+        assert_eq!(oldest_valid, 68);
+        assert!(10 < oldest_valid, "index 10 should trigger catch-up");
```
pedro-alves-ferreira: Given the previous assertion, this is always true.

pedro-alves-ferreira (approved):
LGTM. I think this is a very clean and robust approach. I just have a few comments/questions.

rust/gst-mxl-rs/src/mxlsrc/imp.rs
garethsb: Yes, added additional comment in 13c2b8591014d6f20744f5b8803e4e57b7fa2e22.

rust/gst-mxl-rs/src/mxlsrc/create_audio.rs
garethsb: True! Resolved in 13c2b8591014d6f20744f5b8803e4e57b7fa2e22.

rust/gst-mxl-rs/src/mxlsrc/create_audio.rs
garethsb: True! Resolved in 13c2b8591014d6f20744f5b8803e4e57b7fa2e22.

rust/gst-mxl-rs/src/mxlsrc/create_audio.rs
garethsb: Yeah, should probably have been temporary, removed in 13c2b8591014d6f20744f5b8803e4e57b7fa2e22.

garethsb:
⬇️ I briefly experimented in bca545d619e4264da2668ded2fa0e5de7f2546df with adjusting the PTS of an initial buffer that would have been skipped because it straddled running-time `0` so that it starts at `0` with a correspondingly shortened duration. However I think it's complication for no practical benefit in the end, so I've reset to the previous commit.

rust/gst-mxl-rs/src/clock.rs
```diff
@@ -0,0 +1,191 @@
...
+/// The constant offset between MXL (TAI) time and the pipeline clock, sampled
+/// once. `None` when the element has no clock yet.
+///
+/// A reader subtracts `D + base_time` from a grain's absolute MXL timestamp to
+/// get its PTS; a writer adds the same to a buffer's PTS to get the MXL
+/// timestamp it maps to a grain index. Sampling `D` once (and reusing it) is
+/// what keeps two flows mapping the same frame to the same absolute index.
+pub(crate) fn clock_offset(clock: Option<gst::Clock>, mxl_now: u64) -> Option<u64> {
+    let clock = clock?;
+    Some(mxl_now.saturating_sub(clock.time().nseconds()))
```
mlefebvre1: Could be `Some(mxl_now.saturating_sub(clock?.time().nseconds()))` with the previous line removed

rust/gst-mxl-rs/src/mxlsink/imp.rs
```diff
@@ -243,14 +228,63 @@ impl ElementImpl for MxlSink {
...
+        self.obj().clone().upcast()
+    }
+
+    fn mxl_now(&self) -> Option<u64> {
+        let context = self.context.lock().ok()?;
+        Some(context.state.as_ref()?.instance.get_time())
+    }
+
+    fn cached_clock_offset(&self) -> Option<crate::clock::SharedClockOffset> {
+        self.context.lock().ok()?.shared_offset.clone()
```
mlefebvre1: If we receive None, we can't distinguish between a poisoned mutex or an empty SharedClockOffset.

rust/gst-mxl-rs/src/mxlsink/imp.rs
```diff
@@ -243,14 +228,63 @@ impl ElementImpl for MxlSink {
     ) -> Result<gst::StateChangeSuccess, gst::StateChangeError> {
         self.parent_change_state(transition)
     }
+
+    fn set_clock(&self, clock: Option<&gst::Clock>) -> bool {
+        // Re-sample `D` against the newly selected clock.
+        if crate::clock::ClockOffsetExt::invalidate_clock_offset(self).is_err() {
```
mlefebvre1: `self.invalidate_clock_offset().is_err()`

rust/gst-mxl-rs/src/mxlsink/imp.rs
```diff
@@ -243,14 +228,63 @@ impl ElementImpl for MxlSink {
...
+                ["Internal clock offset error"]
+            );
+            return false;
+        }
+        self.parent_set_clock(clock)
+    }
+
+    fn set_context(&self, context: &gst::Context) {
+        if let Some(cell) = crate::clock::clock_offset_from_context(context) {
+            crate::clock::ClockOffsetExt::store_clock_offset(self, cell);
```
mlefebvre1: This pattern seems to be used at multiple places and I prefer to just add the trait in scope and replace with `self.` instead.

rust/gst-mxl-rs/src/clock.rs
```diff
@@ -0,0 +1,191 @@
...
+                && let Some(context) = query.context_owned()
+                && let Some(cell) = clock_offset_from_context(&context)
+            {
+                self.store_clock_offset(cell.clone());
+                return Ok(cell);
+            }
+        }
+
+        // Ask the pipeline; a sibling may answer synchronously via set_context.
+        let _ = element.post_message(
```
mlefebvre1: Handle that potential error.

rust/gst-mxl-rs/src/clock.rs
```diff
@@ -0,0 +1,191 @@
...
+        // Nobody has one yet: create ours and share it with the pipeline.
+        let cell = SharedClockOffset::default();
+        self.store_clock_offset(cell.clone());
+        let mut context = gst::Context::new(CLOCK_OFFSET_CONTEXT, true);
+        let Some(s) = context.get_mut() else {
+            return Err(ClockOffsetError::Failed);
+        };
+        s.structure_mut().set("offset", &cell);
+        element.set_context(&context);
+        let _ = element.post_message(
```
mlefebvre1: Handle that potential error

rust/gst-mxl-rs/src/mxlsink/render_audio.rs
```diff
@@ -1,131 +1,118 @@
...
 
-    let num_channels = audio_state.flow_def.channel_count as usize;
     let mut remaining = samples_per_buffer;
     let mut src_offset_samples = 0;
-    let mut base_pts = gst_pts.nseconds() + initial_pts_offset;
+    // First chunk's MXL timestamp; each chunk advances by its own duration so
+    // consecutive chunks map to consecutive sample indices.
+    let mut base_mxl_ts = gst_pts
+        .nseconds()
+        .saturating_add(base_time.nseconds())
```
mlefebvre1: Very unlikely, but we should probably use `checked_add` (or wrapping) instead? Otherwise if this truly materialize we get a stale sink and we won't know why. Same comments for render_data and render_video

rust/gst-mxl-rs/src/mxlsink/imp.rs
```diff
@@ -243,14 +228,63 @@ impl ElementImpl for MxlSink {
...
+    fn mxl_now(&self) -> Option<u64> {
+        let context = self.context.lock().ok()?;
+        Some(context.state.as_ref()?.instance.get_time())
+    }
+
+    fn cached_clock_offset(&self) -> Option<crate::clock::SharedClockOffset> {
+        self.context.lock().ok()?.shared_offset.clone()
+    }
+
+    fn store_clock_offset(&self, cell: crate::clock::SharedClockOffset) {
```
mlefebvre1: Again, we have no feedback if the Mutex was poisoned or not.

rust/gst-mxl-rs/src/mxlsrc/create_data.rs
```diff
@@ -1,96 +1,107 @@
...
+        // running time (before the consumer joined); a live source must not emit
+        // it. Skip forward rather than clamp its PTS to 0 (which would corrupt
+        // the first inter-frame interval). Both readers share `subtrahend`, so
+        // they skip the same grains and stay index-aligned.
+        trace!("Skipping pre-start grain {read_index} (running time would be negative)");
+        data_state.next_discont |= jumped || slot_discont;
+        data_state.index = read_index + 1;
+        return Ok(CreateState::NoDataCreated);
+    };
+    let is_discont = jumped || slot_discont || std::mem::take(&mut data_state.next_discont);
```
mlefebvre1: Is `data_state.next_discont` meant to be a one-shot deferred flag? If yes, then std::mem::take should be done in the previous line, otherwise the if could shortcuit before this gets evaluated.

Same pattern in create_video.

rust/gst-mxl-rs/src/mxlsrc/create_data.rs
```diff
@@ -1,96 +1,107 @@
...
+    ReadStep, discrete_grain_count, flow_head_index, index_period, pts_for_index, resolve_read_step,
+};
 use gstreamer as gst;
 use tracing::trace;
 
 const GET_GRAIN_TIMEOUT: Duration = Duration::from_secs(5);
 pub(super) const MXL_GRAIN_FLAG_INVALID: u32 = 0x00000001;
 
-pub(crate) fn create_data(src: &MxlSrc, state: &mut State) -> Result<CreateState, gst::FlowError> {
+pub(crate) fn create_data(
```
mlefebvre1: This function shares the same logic with `create_video` and could be merged as one with very little effort.

mlefebvre1 (changes requested):
Thank you for this. I've only reviewed `gst-mxl-rs` so far, so here's my comments.

rust/gst-mxl-rs/src/mxlsink/imp.rs
garethsb: Fixed in de140d5.

rust/gst-mxl-rs/src/mxlsink/imp.rs
garethsb: Fixed in de140d5.

rust/gst-mxl-rs/src/mxlsink/imp.rs
garethsb: Fixed in de140d5.

rust/gst-mxl-rs/src/mxlsink/imp.rs
garethsb: Fixed in de140d5.

rust/gst-mxl-rs/src/mxlsink/render_audio.rs
garethsb: Fixed in de140d5.

rust/gst-mxl-rs/src/mxlsrc/create_data.rs
garethsb: This was a big enough refactor to address as a separate commit. It did save ~150 lines of code in total and align the lower layers of gst-mxl-rs with the MXL SDK. See if you like 6ad1879, @mlefebvre1, @pedro-alves-ferreira.

rust/gst-mxl-rs/src/mxlsrc/create_data.rs
garethsb: Good spot. Fixed in de140d5.

rust/gst-mxl-rs/src/clock.rs
garethsb: Fixed in de140d5.

rust/gst-mxl-rs/src/clock.rs
garethsb: Fixed in de140d5.

rust/gst-mxl-rs/src/clock.rs
garethsb: Fixed in de140d5.

rust/gst-mxl-rs/src/mxlsink/state.rs
```diff
@@ -40,24 +40,26 @@ impl Default for Settings {
 pub(crate) struct State {
     pub instance: MxlInstance,
     pub flow: Option<FlowConfigInfo>,
-    pub video: Option<VideoState>,
-    pub audio: Option<AudioState>,
-    pub data: Option<DataState>,
+    pub discrete: Option<DiscreteState>,
```
pedro-alves-ferreira: If you can only have one (discrete or continuous), that would be better modeled as an enum:

```rust
enum InnerState {
    Discrete(DiscreteState),
    Continuous(ContinuousState)
}

pub(crate) struct State {
    pub instance: MxlInstance,
    pub flow: Option<FlowConfigInfo>,
    pub inner: Option<InnerState>,
}
```

garethsb: Yep! Fixed in ~[d60b6ee](https://github.com/dmf-mxl/mxl/compare/96fa41ca306dcd920c5fb3a9a6dbccea3bf7389c..d60b6eea983cc5058a9aede0c9534490a5c5fa6c)~ [7cc9f91](https://github.com/dmf-mxl/mxl/compare/96fa41ca306dcd920c5fb3a9a6dbccea3bf7389c..7cc9f91181870a235447adda1647620c56d0d2df).

pedro-alves-ferreira: Thanks!

garethsb:
> Thank you for this. I've only reviewed `gst-mxl-rs` so far, so here's my comments.

@mlefebvre1 - I've now made consistent changes to **gst-avsynctest-rs** based on your review of **gst-mxl-rs** and other things that @pedro-alves-ferreira and @SimaoPFonseca raised - hoping we can get this merged 😄

mlefebvre1 (commented):
Thank you again, LGTM

mlefebvre1 (approved):

Issue #563, opened 2026-06-16
[Task] Include Doxygen API in GitHub Pages documentation


peterbrightwell:
### Description

As well as the markdown docs/ it would be useful to see the API information rendered on GitHub pages (currently at https://dmf-mxl.github.io/mxl).  This will be a separate render using Doxygen rather than Zensical.  Should also be versioned.

peterbrightwell:
https://peterbrightwell.github.io/mxl/api/latest/ etc show API info for main and some selected tags.  Currently this is a separate GitHub workflow (doxy-docs.yml) to the Zensical/mike one (docs.yml) and has a Copilot-generated javascript version selector.

peterbrightwell:
This is now working (see #604 and https://dmf-mxl.github.io/mxl/api/latest/ etc).  I also added links doxygen -> docs -> repo.

PR #604, opened 2026-07-17
Add Doxygen rendering of API documentation

peterbrightwell:
# Details
Adds a doxy-docs.yml action to build Doxygen pages. Separate to the Zensical build but both happen via gh-pages and are consistent. There are links between.  Requires settings change for Pages to build from a new pages-depoly.yml action rather than GitHub's legacy builder, which uses deprecated node version.

Tested ok at https://peterbrightwell.github.io/mxl/api/latest/ etc

Uses some GitHub Copilot code.

Addresses #563

# Backport requirements
Maybe not because it runs from main by default.

### DIFF
--- a/.github/doxygen/github-repo-link.css
+++ b/.github/doxygen/github-repo-link.css
@@ -1,0 +1,28 @@
+/* SPDX-FileCopyrightText: 2025 Contributors to the Media eXchange Layer project.
+ * SPDX-License-Identifier: CC-BY-4.0
+ *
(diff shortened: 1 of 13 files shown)


vt-tv (approved):
LGTM

peterbrightwell:
Confirmed working at https://dmf-mxl.github.io/mxl/api/latest/ etc (after manually running workflows to get things up to date-- this shouldn't be necessary in future though)

PR #613, opened 2026-07-22
Fix devcontainer's dockerfile for use in multiarch environments

lapointejp:
# Details
As we package MXL in our builds, we use the devcontainer in a multiarch environment to produce builds for x64 as well as ARM (aarch64). Unfortunately, the sudo command doesn't work within the multiarch environment for ARM. This fixes that workflow for us.

# Backport requirements
Yes: release/1.1


vt-tv (approved):
LGTM

KimonHoffmann (approved):
Looks good to me!

jonasohland (approved):

PR #620, opened 2026-07-23
Fix slice transfer for flows with alpha channel

### LINKED ISSUE #592: [Feature]  Open Fabrics: partial-grain transfer support for sliced alpha
### Description

complete the partial-grain transfer with sliced alpha

jonasohland:
Fix the RMA based protocol (the only one we have) to correctly handle alpha channel when transferring slices.

Excluding the SHM provider from the tests for now, because the writer cannot handle multiple writes without completion queue reads in between. We intent to fix this later together with some other possible optimizations.

Fixes #592

### DIFF
--- a/lib/fabrics/ofi/src/internal/DataLayout.cpp
+++ b/lib/fabrics/ofi/src/internal/DataLayout.cpp
@@ -9,10 +9,56 @@
 
 namespace mxl::lib::fabrics::ofi
 {
-    DataLayout DataLayout::fromDiscrete(std::array<std::uint32_t, MXL_MAX_PLANES_PER_GRAIN> const& sliceSizes) noexcept
+    DataLayout DataLayout::fromDiscrete(std::array<std::uint32_t, MXL_MAX_PLANES_PER_GRAIN> const& sliceSizes, std::uint16_t totalSlices) noexcept
     {
-        return DataLayout{DataLayout::Discrete{.sliceSizes = sliceSizes}};
+        return DataLayout{
+            DataLayout::Discrete{.sliceSizes = sliceSizes, .totalSlices = totalSlices}
+        };
     };
+
+    std::size_t DataLayout::Discrete::totalLength() const noexcept
+    {
+        auto totalLength = std::size_t{0};
+        for (auto const s : sliceSizes)
+        {
+            if (s == 0)
+            {
+                return totalLength;
+            }
+
+            totalLength += s;
+        }
+
+        return totalLength;
+    }
+
+    std::size_t DataLayout::Discrete::activePlaneCount() const noexcept
+    {
+        auto count = std::size_t{0};
+        for (auto s : sliceSizes)
+        {
+            if (s == 0)
+            {
+                break;
+            }
+            ++count;
+        }
+        return count;
+    }
+
+    std::uint32_t DataLayout::Discrete::planePayloadOffset(std::size_t planeIndex, std::uint32_t grainPayloadOffset) const noexcept
+    {
+        auto offset = grainPayloadOffset;
+        for (std::size_t i = 0; i < planeIndex && i < MXL_MAX_PLANES_PER_GRAIN; ++i)
+        {
+            if (sliceSizes[i] == 0)
+            {
+                break;
+            }
+            offset += static_cast<std::uint32_t>(totalSlices) * sliceSizes[i];
+        }
+        return offset;
+    }
 
     DataLayout DataLayout::fromContinuous(std::size_t sampleSize, std::size_t channelCount, std::size_t bufferLength) noexcept
     {
--- a/lib/fabrics/ofi/src/internal/DataLayout.cpp
+++ b/lib/fabrics/ofi/src/internal/DataLayout.cpp
@@ -4,8 +4,11 @@
 
 #include "DataLayout.hpp"
 #include <cassert>
+#include <algorithm>
 #include <array>
+#include <numeric>
 #include "mxl/flowinfo.h"
+#include "Exception.hpp"
 
 namespace mxl::lib::fabrics::ofi
 {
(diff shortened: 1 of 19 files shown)


lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
```diff
@@ -28,16 +28,45 @@ namespace mxl::lib::fabrics::ofi
...
+
+            if (plane == 0)
+            {
+                offset = sliceRange.transferOffset(payloadOffset, sliceSize);
+                size = sliceRange.transferSize(payloadOffset, sliceSize);
+            }
+            else
+            {
+                auto const planeBase = _layout.planePayloadOffset(plane, payloadOffset);
+                offset = planeBase + (sliceRange.start() * sliceSize);
```
mlefebvre1: any reason you can't use the following ?

```c++
offset = sliceRange.transferOffset(planeBase, sliceSize);
```

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
```diff
@@ -28,16 +28,45 @@ namespace mxl::lib::fabrics::ofi
...
+                size = (numSlices * sliceSize);
```
mlefebvre1: same question for this one

mlefebvre1 (commented):
Looks good, just 2 questions.

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
jonasohland: you are totally right!

mlefebvre1 (approved):
LGTM

lib/tests/fabrics/test_sliced.cpp
```diff
@@ -0,0 +1,547 @@
...
+
+                // generate a value for fill and key buffer
+                auto fillValue = static_cast<std::uint8_t>(0xAC & ~middleSlice);
+                auto keyValue = static_cast<std::uint8_t>(0xAB & ~middleSlice);
+
+                // write to a slice in the middle of the block
+                this->writeInitiatorGrainSlice(grainIndex, middleSlice, fillValue, keyValue);
+
+                // transfer the slices
+                REQUIRE(this->transferGrainSlices(grainIndex, lastSlice, endSlice) == grainIndex);
```
k0zmo, line 525: While `endSlice` is incremented during the test, the `lastSlice` is not. Is it intentional?

lib/fabrics/ofi/src/internal/DataLayout.cpp
```diff
@@ -9,11 +9,57 @@
...
     {
-        return DataLayout{DataLayout::Discrete{.sliceSizes = sliceSizes}};
+        return DataLayout{
+            DataLayout::Discrete{.sliceSizes = sliceSizes, .totalSlices = totalSlices}
+        };
     };
 
+    std::size_t DataLayout::Discrete::totalLength() const noexcept
+    {
+        auto totalLength = std::size_t{0};
```
k0zmo: Could use `std::accumulate`

lib/fabrics/ofi/src/internal/DataLayout.cpp
```diff
@@ -9,11 +9,57 @@
...
+            totalLength += s;
+        }
+
+        return totalLength;
+    }
+
+    std::size_t DataLayout::Discrete::activePlaneCount() const noexcept
+    {
+        auto count = std::size_t{0};
+        for (auto s : sliceSizes)
```
k0zmo: Could use `count_if`

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
```diff
@@ -28,16 +28,44 @@ namespace mxl::lib::fabrics::ofi
...
-        auto const remoteRegion = _remoteInfo.remoteRegions[remoteIndex % _remoteInfo.remoteRegions.size()].sub(remoteOffset, remoteSize);
+        auto const localGrain = _localRegions[localIndex % _localRegions.size()];
+        auto const remoteGrain = _remoteInfo.remoteRegions[remoteIndex % _remoteInfo.remoteRegions.size()];
         auto const remoteSlot = remoteIndex % _remoteInfo.remoteRegions.size();
 
-        _pending += ep.write(_token, localRegion, remoteRegion, destAddr, ImmDataGrain{remoteSlot, sliceRange.end()}.data());
+        // Fast path: a full grain is transferred, can do one write no matter how many planes
+        if ((sliceRange.start() == 0) && (sliceRange.end() == _layout.totalSlices))
+        {
+            _pending += ep.write(_token, localGrain, remoteGrain, destAddr, std::make_optional(ImmDataGrain{remoteSlot, _layout.totalSlices}.data()));
```
k0zmo, line 38: I don't think the `make_optional` is needed here, unless we care about the explicitness of it

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
```diff
@@ -28,16 +28,44 @@ namespace mxl::lib::fabrics::ofi
...
+            auto size = std::uint32_t{0};
+
+            if (plane == 0)
+            {
+                offset = sliceRange.transferOffset(payloadOffset, sliceSize);
+                size = sliceRange.transferSize(payloadOffset, sliceSize);
+            }
+            else
+            {
+                auto const planeBase = _layout.planePayloadOffset(plane, payloadOffset);
```
k0zmo, line 65: I'm not entirely sure these calculations are correct. When `sliceRange`'s `_start` is 0 then the `transferOffset` returns 0 and includes the plane 0/fill data. Proper implementation would be:

```cpp
offset = planeBase + sliceRange.start() * sliceSize;
size = (sliceRange.end() - sliceRange.start()) * sliceSize;
```

which is what you had in the initial version

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
k0zmo: The suggested implementation is not correct

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
jonasohland: I will have a look

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
jonasohland: I don't like implicit conversions :smiley:

lib/fabrics/ofi/src/internal/DataLayout.cpp
jonasohland: Completely irrelevant nitpick, but: std::accumulate is slightly less performant because the loop can stop at the first `sliceSize == 0` and std::accumulate can't.
But it is a std::array and the compiler will unroll the loop anyway plus std::accumulate expresses the intent better, so I am replacing it here.

lib/tests/fabrics/test_sliced.cpp
jonasohland: `endSlice` is always set to `startSlice+blockSize` at the start of the loop

lib/fabrics/ofi/src/internal/ProtocolEgressRMA.cpp
jonasohland: You are right. There was some behavior hidden in `SliceRange` to make sure that the grain header is transferred with the range that includes slice 0.
I have moved that out of the slice range class and into the code in the loop directly.

vt-tv (approved):
Approving on behalf of @k0zmo

PR #621, opened 2026-07-23
Add Rust Bindings for Fabrics API

### LINKED ISSUE #182: Fabrics: Rust bindings
Create rust bindings for the fabrics API

mlefebvre1:
# Details
As the title describes, this adds Rust Bindings to the new Fabrics API. It also includes a demo app written in Rust that behaves similarly to demo.cpp.

This PR will close https://github.com/dmf-mxl/mxl/issues/182.

# Backport requirements
v1.1

### DIFF
--- a/rust/mxl-sys/build.rs
+++ b/rust/mxl-sys/build.rs
@@ -124,8 +124,6 @@ fn main() {
         println!("cargo:include={include_dir}");
     }
 
-    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
-
     let bindings = bindgen::builder()
         .clang_args(
             bindgen_specs
@@ -143,6 +141,7 @@ fn main() {
         .generate()
         .unwrap();
 
+    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
     bindings
         .write_to_file(out_path.join("bindings.rs"))
         .expect("Could not write bindings");
--- a/rust/mxl-sys/build.rs
+++ b/rust/mxl-sys/build.rs
@@ -63,11 +63,68 @@ fn get_bindgen_specs() -> BindgenSpecs {
     }
 }
 
+fn get_bindgen_specs_fabrics() -> BindgenSpecs {
+    let header = "wrapper-fabrics.h".to_string();
+
+    let manifest_dir =
+        PathBuf::from(env::var("CARGO_MANIFEST_DIR").expect("failed to get current directory"));
+    let repo_root = manifest_dir.parent().unwrap().parent().unwrap();
+    let mut includes_dirs = vec![
+        repo_root
+            .join("lib")
+            .join("include")
+            .to_string_lossy()
+            .to_string(),
+        repo_root
+            .join("lib")
+            .join("fabrics")
+            .join("include")
+            .to_string_lossy()
+            .to_string(),
+    ];
+    if cfg!(not(feature = "mxl-not-built")) {
+        let out_dir = PathBuf::from(std::env::var("OUT_DIR").unwrap());
+        let build_version_dir = out_dir.join("include").to_string_lossy().to_string();
+
+        includes_dirs.push(build_version_dir);
+
+        // Rebuild if any file in lib/ changes
+        let lib_root = repo_root.join("lib");
+        println!("cargo:rerun-if-changed={}", lib_root.display());
+
+        let dst = cmake::Config::new(repo_root)
+            .generator("Ninja")
+            .configure_arg("--preset")
+            .configure_arg(BUILD_VARIANT)
+            .configure_arg("-B")
+            .configure_arg(out_dir.join("build"))
+            .define("MXL_ENABLE_FABRICS_OFI", "ON")
+            .define("BUILD_DOCS", "OFF")
+            .define("BUILD_TESTS", "OFF")
+            .define("BUILD_TOOLS", "OFF")
+            .build();
+
+        println!(
+            "cargo:rustc-link-search={}",
+            dst.join("lib/fabrics/ofi").display()
+        );
+        println!("cargo:rustc-link-lib=mxl");
+        println!("cargo:rustc-link-lib=mxl-fabrics");
+    }
+
+    BindgenSpecs {
+        header,
+        includes_dirs,
+    }
+}
+
 fn main() {
     let bindgen_specs = get_bindgen_specs();
     for include_dir in &bindgen_specs.includes_dirs {
         println!("cargo:include={include_dir}");
     }
+
+    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
 
     let bindings = bindgen::builder()
         .clang_args(
@@ -86,10 +143,31 @@ fn main() {
         .generate()
         .unwrap();
 
-    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
     bindings
         .write_to_file(out_path.join("bindings.rs"))
         .expect("Could not write bindings");
+
+    if cfg!(feature = "mxl-fabrics-ofi") {
+        let fabrics_bindings = bindgen::builder()
+            .clang_args(
+                get_bindgen_specs_fabrics()
+                    .includes_dirs
+                    .iter()
+                    .map(|dir| format!("-I{dir}")),
+            )
+            .header("wrapper-fabrics.h")
+            .derive_default(true)
+            .derive_debug(true)
+            .prepend_enum_name(false)
+            .dynamic_library_name("libmxlfabrics")
+            .dynamic_link_require_all(false)
+            .parse_callbacks(Box::new(CB))
+            .generate()
+            .unwrap();
+        fabrics_bindings
+            .write_to_file(out_path.join("fabrics_bindings.rs"))
+            .expect("Could not write fabrics bindings");
+    }
 }
 
 #[derive(Debug)]
--- a/rust/mxl-sys/build.rs
+++ b/rust/mxl-sys/build.rs
@@ -151,7 +151,7 @@ fn main() {
             .blocklist_var(".*")
             .raw_line("use crate::types::*;")
             .dynamic_library_name("libmxlfabrics")
-            .dynamic_link_require_all(false)
+            .dynamic_link_require_all(true)
             .parse_callbacks(Box::new(CB))
             .generate()
             .unwrap();
--- a/rust/mxl-sys/build.rs
+++ b/rust/mxl-sys/build.rs
@@ -31,6 +31,16 @@ fn get_bindgen_specs() -> BindgenSpecs {
             .to_string_lossy()
             .to_string(),
     ];
+    if cfg!(feature = "mxl-fabrics-ofi") {
+        includes_dirs.push(
+            repo_root
+                .join("lib")
+                .join("fabrics")
+                .join("include")
+                .to_string_lossy()
+                .to_string(),
+        );
+    }
     if cfg!(not(feature = "mxl-not-built")) {
         let out_dir = PathBuf::from(std::env::var("OUT_DIR").unwrap());
         let build_version_dir = out_dir.join("include").to_string_lossy().to_string();
@@ -41,7 +51,8 @@ fn get_bindgen_specs() -> BindgenSpecs {
         let lib_root = repo_root.join("lib");
         println!("cargo:rerun-if-changed={}", lib_root.display());
 
-        let dst = cmake::Config::new(repo_root)
+        let mut config = cmake::Config::new(repo_root);
+        config
             .generator("Ninja")
             .configure_arg("--preset")
             .configure_arg(BUILD_VARIANT)
@@ -50,66 +61,21 @@ fn get_bindgen_specs() -> BindgenSpecs {
             .define("BUILD_DOCS", "OFF")
             .define("BUILD_TESTS", "OFF")
             .define("BUILD_TOOLS", "OFF")
-            .define("CMAKE_INSTALL_LIBDIR", "lib")
-            .build();
+            .define("CMAKE_INSTALL_LIBDIR", "lib");
+        if cfg!(feature = "mxl-fabrics-ofi") {
+            config.define("MXL_ENABLE_FABRICS_OFI", "ON");
+        }
+        let dst = config.build();
 
         println!("cargo:rustc-link-search={}", dst.join("lib").display());
         println!("cargo:rustc-link-lib=mxl");
-    }
-
-    BindgenSpecs {
-        header,
-        includes_dirs,
-    }
-}
-
-fn get_bindgen_specs_fabrics() -> BindgenSpecs {
-    let header = "wrapper-fabrics.h".to_string();
-
-    let manifest_dir =
-        PathBuf::from(env::var("CARGO_MANIFEST_DIR").expect("failed to get current directory"));
-    let repo_root = manifest_dir.parent().unwrap().parent().unwrap();
-    let mut includes_dirs = vec![
-        repo_root
-            .join("lib")
-            .join("include")
-            .to_string_lossy()
-            .to_string(),
-        repo_root
-            .join("lib")
-            .join("fabrics")
-            .join("include")
-            .to_string_lossy()
-            .to_string(),
-    ];
-    if cfg!(not(feature = "mxl-not-built")) {
-        let out_dir = PathBuf::from(std::env::var("OUT_DIR").unwrap());
-        let build_version_dir = out_dir.join("include").to_string_lossy().to_string();
-
-        includes_dirs.push(build_version_dir);
-
-        // Rebuild if any file in lib/ changes
-        let lib_root = repo_root.join("lib");
-        println!("cargo:rerun-if-changed={}", lib_root.display());
-
-        let dst = cmake::Config::new(repo_root)
-            .generator("Ninja")
-            .configure_arg("--preset")
-            .configure_arg(BUILD_VARIANT)
-            .configure_arg("-B")
-            .configure_arg(out_dir.join("build"))
-            .define("MXL_ENABLE_FABRICS_OFI", "ON")
-            .define("BUILD_DOCS", "OFF")
-            .define("BUILD_TESTS", "OFF")
-            .define("BUILD_TOOLS", "OFF")
-            .build();
-
-        println!(
-            "cargo:rustc-link-search={}",
-            dst.join("lib/fabrics/ofi").display()
-        );
-        println!("cargo:rustc-link-lib=mxl");
-        println!("cargo:rustc-link-lib=mxl-fabrics");
+        if cfg!(feature = "mxl-fabrics-ofi") {
+            println!(
+                "cargo:rustc-link-search={}",
+                dst.join("lib/fabrics/ofi").display()
+            );
+            println!("cargo:rustc-link-lib=mxl-fabrics");
+        }
     }
 
     BindgenSpecs {
@@ -124,7 +90,32 @@ fn main() {
         println!("cargo:include={include_dir}");
     }
 
-    let bindings = bindgen::builder()
+    let mut types_builder = bindgen::builder()
+        .clang_args(
+            bindgen_specs
+                .includes_dirs
+                .iter()
+                .map(|dir| format!("-I{dir}")),
+        )
+        .header(bindgen_specs.header.clone())
+        .derive_default(true)
+        .derive_debug(true)
+        .prepend_enum_name(false)
+        .blocklist_function(".*")
+        .parse_callbacks(Box::new(CB));
+    if cfg!(feature = "mxl-fabrics-ofi") {
+        types_builder = types_builder.header("wrapper-fabrics.h");
+    }
+    let types_bindings = types_builder
+        .generate()
+        .expect("Could not generate shared type bindings");
+
+    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
+    types_bindings
+        .write_to_file(out_path.join("types_bindings.rs"))
+        .expect("Could not write shared type bindings");
+
+    let mxl_bindings = bindgen::builder()
         .clang_args(
             bindgen_specs
                 .includes_dirs
@@ -132,32 +123,33 @@ fn main() {
                 .map(|dir| format!("-I{dir}")),
         )
         .header(bindgen_specs.header)
-        .derive_default(true)
-        .derive_debug(true)
-        .prepend_enum_name(false)
+        .allowlist_function("mxl.*")
+        .blocklist_type(".*")
+        .blocklist_var(".*")
+        .raw_line("use crate::types::*;")
         .dynamic_library_name("libmxl")
         .dynamic_link_require_all(true)
         .parse_callbacks(Box::new(CB))
         .generate()
-        .unwrap();
+        .expect("Could not generate libmxl function bindings");
 
-    let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
-    bindings
-        .write_to_file(out_path.join("bindings.rs"))
-        .expect("Could not write bindings");
+    mxl_bindings
+        .write_to_file(out_path.join("mxl_bindings.rs"))
+        .expect("Could not write libmxl function bindings");
 
     if cfg!(feature = "mxl-fabrics-ofi") {
         let fabrics_bindings = bindgen::builder()
             .clang_args(
-                get_bindgen_specs_fabrics()
+                bindgen_specs
                     .includes_dirs
                     .iter()
                     .map(|dir| format!("-I{dir}")),
             )
             .header("wrapper-fabrics.h")
-            .derive_default(true)
-            .derive_debug(true)
-            .prepend_enum_name(false)
+            .allowlist_function("mxlFabrics.*")
+            .blocklist_type(".*")
+            .blocklist_var(".*")
+            .raw_line("use crate::types::*;")
             .dynamic_library_name("libmxlfabrics")
             .dynamic_link_require_all(false)
             .parse_callbacks(Box::new(CB))
--- a/rust/mxl/examples/fabrics-demo.rs
+++ b/rust/mxl/examples/fabrics-demo.rs
@@ -469,8 +469,8 @@ fn main() -> Result<(), anyhow::Error> {
 
     let interface_config = mxl::fabrics::InterfaceConfig::builder(endpoint_address)
         .provider(provider.prov_type().clone())
-        .caps(Capabilities::builder().supports_remote_write().build())
-        .build()?;
+        .caps(Capabilities::default())
+        .build();
 
     let interfaces = fabrics_instance
         .get_interfaces(Some(interface_config))
--- a/rust/mxl/examples/fabrics-demo.rs
+++ b/rust/mxl/examples/fabrics-demo.rs
@@ -16,11 +16,11 @@ use clap::{Parser, Subcommand};
 use base64::{Engine as _, prelude::BASE64_STANDARD};
 
 use mxl::{
-    Error, FlowConfigInfo, FlowInfo, FlowReader, FlowWriter, GrainReader, GrainWriter,
-    MxlFabricsApi, MxlInstance, SamplesReader, SamplesWriter,
+    Error, FlowConfigInfo, FlowInfo, FlowReader, FlowWriter, GrainReader, GrainWriter, MxlInstance,
+    SamplesReader, SamplesWriter,
     config::{get_mxl_fabrics_ofi_so_path, get_mxl_so_path},
     fabrics::{
-        EndpointAddress, FabricsInstance, TargetInfo,
+        Capabilities, EndpointAddress, FabricsInstance, InterfaceConfig, ProviderType, TargetInfo,
         initiator::{self, Initiator},
         target::{self, Target},
     },
@@ -40,24 +40,23 @@ pub struct Cli {
     #[arg(
         short,
         long,
-        default_value = "tcp",
-        help = "The fabrics provider. One of (tcp, verbs or efa)."
+        help = "Force a specific provider (tcp, verbs, efa or shm). Auto-selected if not specified."
     )]
-    pub provider: String,
+    pub provider: Option<String>,
 
     #[arg(
         short,
         long,
-        help = "This corresponds to the interface identifier of the fabrics endpoint, it can also be a logical address. This can be seen as the bind address when using sockets."
+        help = "Filter interface selection by node address. If not set, the best available interface is chosen automatically."
     )]
-    pub node: String,
+    pub node: Option<String>,
 
     #[arg(
         short,
         long,
-        help = "This corresponds to a service identifier for the fabrics endpoint. This can be seen as the bind port when using sockets."
+        help = "Service identifier for the fabrics endpoint (e.g. a port number)."
     )]
-    pub service: String,
+    pub service: Option<String>,
 
     #[command(subcommand)]
     pub command: Command,
@@ -72,6 +71,12 @@ pub enum Command {
             help = "The JSON file which contains the NMOS Flow configuration."
         )]
         flow_file: String,
+        #[arg(
+            long,
+            help = "Output file path for raw target info (optional, always logged as base64)."
+        )]
+        target_info_path: Option<String>,
+        //TODO: flow options?
     },
     /// Run as an initiator (flow reader + fabrics initiator).
     Initiator {
@@ -79,7 +84,7 @@ pub enum Command {
         flow_id: String,
         #[arg(
             long,
-            help = "The target information to send to. Start the target first and paste the printed target info here."
+            help = "Base64-encoded target info, or a path prefixed with '@' to read raw target info from a file."
         )]
         target_info: String,
     },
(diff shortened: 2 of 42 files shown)


mlefebvre1:
As a general question, should we follow the same versioning for the Rust bindings crates as the project itself? Currently the crates are still at version 0.1.0

pedro-alves-ferreira:
> As a general question, should we follow the same versioning for the Rust bindings crates as the project itself? Currently the crates are still at version 0.1.0

Good point. What I would like to is to keep all Rust bindings on par with the SDK version. 

My concern is what do we do if we need to change the bindings, irrespective of any change in the SDK.
We could fix bugs and change the patch version, independently from the SDK.
But, if we need/want to change the Rust API, then it's trickier.
I'm a bit torn here.

garethsb:
I would say a version for the FFI layer should slavishly follow the the SDK version but maybe the higher level Rust API could be separately versioned?

pedro-alves-ferreira:
> I would say a version for the FFI layer should slavishly follow the the SDK version but maybe the higher level Rust API could be separately versioned?

Sounds good to me. And then we would use semantic versioning for the higher level API.

garethsb:
**Not sure how useful this will be. I have not reviewed this in detail, myself. I have asked a series of questions of my 🤖 buddy, and then summarized the conversation. I will come back to this after 1 September.**

---

The direction is right (feature-gated bindgen in `mxl-sys`, RAII wrapper in `mxl`, TCP tests that actually move payload). CI looks good.

The crate sits at the same abstraction level as the existing core `mxl` SDK bindings, which is what we want, but a few FFI/API concerns and an ownership hole vs the core pattern should be fixed before merge.

## Abstraction / layering

This is the same layer as today’s `mxl` crate, not a new session/media API:

- `mxl-sys` = bindgen + `libloading`
- `mxl` = owned handles, `Drop`, `Result`, `Send` and not `Sync`, `Arc` so children keep the instance alive

Call shape matches core (`load_api` → instance → typed handle → C-shaped operations). Grain indices, timeouts, and retry-on-not-ready stay visible. That is appropriate for an SDK binding.

Two divergences from core are worth fixing rather than growing a different style:

1. Core uses **consuming conversions** (`FlowWriter::to_grain_writer()`) so one Rust type owns one C handle. Fabrics uses a phantom type-state (`Initializing` → `Specializing` → `Grain`/`Samples`) that sequences `setup`, but only **borrows** the flow at `setup`. That is more ceremony than core and less ownership.
2. `specialize(&FlowConfigInfo)` does not call C; it only looks at `is_discrete_flow()`. `to_grain_writer()` inspects the **actual** flow. Prefer specializing from the reader/writer passed to `setup` (or return `Either` from `setup` itself).

A separate `mxl-fabrics` crate is only needed if we want the versioning split already discussed on this PR (FFI tracks SDK; high-level semvers independently). Putting fabrics inside `mxl` / `mxl-sys` is consistent with “one SDK, one Rust stack”. Workspace is still `0.1.0`.

Two bindgen runs produce two opaque pointer types, then the high-level code `.cast()`s between them. Sound if both headers agree; one bindgen of `mxl.h` + `fabrics.h` (or a shared alias) would remove the casts.

## Safety (memory / threads)

The high-level API is a good RAII wrapper. It is **not** completely memory- or thread-safe. Safe Rust can still UAF or race.

**What it does make safe**

- Opaque handles are owned (`FabricsInstance`, `Initiator`, `Target`, `TargetInfo`, `Interfaces` all `Drop` the matching C function).
- `FabricsInstanceContext` holds `Arc<InstanceContext>`, so live fabrics objects keep the MXL instance alive; `MxlInstance::destroy()` will refuse if they still exist.
- `OwnedEndpointAddress` / `OwnedInterfaceConfig` keep `CString`s alive across `setup` / `get_interfaces`.
- `InterfaceConfig<'a>` from `Interfaces::iter()` cannot outlive the list.
- `Send` + not `Sync` on initiator/target matches `FlowReader` / `FlowWriter` for a **single** handle.

**What it does not**

Flow reader/writer are not owned by the fabrics objects. C stores the `mxlFlowReader` / `mxlFlowWriter` at `setup`. Rust `Config` only borrows them for that call. After `setup`, `Initiator` / `Target` have no lifetime and no `Arc` to the flow handle. `FlowReader::Drop` calls `release_flow_reader`, so this compiles and is UB on the C side:

```rust
let reader = instance.create_flow_reader(id)?;
let initiator = fabrics.create_initiator()?.setup(&initiator::Config::new(iface, &reader))?;
drop(reader); // releases the C reader
initiator.specialize(&cfg); // initiator still holds that pointer
```

The tests/demo then `to_grain_reader()` / `to_grain_writer()`, moving the same C pointer into a second Rust type and nulling the first. That aliasing is inherent (fabrics + `GrainReader` both need the reader), but the crate never records it. Dropping the `GrainReader` while the `Initiator` is still live is the same UAF.

Until initiator/target **own or share** the flow they register (the core pattern), “safe fabrics API” oversells it.

Threads: comments say the fabrics API as a whole is not thread-safe. `FabricsInstanceContext` is `Send` and not `Sync`, so `Arc<Context>` would normally be neither. `unsafe impl Send` on `FabricsInstance` / `Initiator` / `Target` lets you move a `Target` to thread A and an `Initiator` to thread B and call into the same instance with no compiler complaint. `TargetInfo` and `Provider` also impl `Sync` on the claim that `*_to_string` / `*_from_string` are a thread-safe subset, but `TargetInfo::Drop` still calls `fabrics_free_target_info`. I would not `Sync` anything that `Drop`s fabrics state unless the C API guarantees it.

`Error::NotReady` is not a safety bug, but it fights the type system: C uses it as “pending”, so `?` is wrong and every caller writes the same retry loop. `Poll<()>` or `enum Progress { Done, Pending }` would match the C comments.

## Must fix

**Initiator/target config `version: 0`.** `InterfaceConfig` correctly sets `MXL_FABRICS_API_VERSION`. `initiator::Config` and `target::Config` hardcode `0`. That equals the current `#define`, so tests pass; the next bump will send the wrong version.

**Timeout truncation on initiator `make_progress`.** Target `read` uses `u16::try_from(timeout.as_millis())` → `InvalidArg`. Initiator does `as u16`, which wraps. Same helper everywhere.

**Demo continuous path advances by 1, not by batch.** C `demo.cpp` does `headIndex += batchSize`. The Rust demo does `index += 1` after `transfer(index, count)` with `count = max_sync_batch_size_hint()`.

**Dangling pointers in unused `TryFrom`.**

```rust
interface: *OwnedInterfaceConfig::try_from(&value.interface)?.as_ffi(),
```

`OwnedInterfaceConfig` is dropped at the end of that expression; the copied FFI struct still holds pointers into it. `setup` uses `OwnedInitiatorConfig::new` (safe). Delete the `TryFrom`, or only return an owned wrapper.

**Example CLI vs implementation.** Initiator help says base64 or `@path`; `run()` always base64-decodes. Target only writes `target_info_path` if it starts with `@`, and the inner `starts_with("@")` / else is dead. Match the C demo.

## Should fix

- **Double CMake in `mxl-sys/build.rs`.** Default bindgen already builds without `MXL_ENABLE_FABRICS_OFI`; the fabrics path runs a second Ninja build with it on, same `-B`. When the feature is on, pass `MXL_ENABLE_FABRICS_OFI=ON` in the first cmake and only add include/link dirs the second time.
- **`Error::from_status` is incomplete for fabrics.** Unmapped (become `Unknown(n)`): `STRLEN`, `NO_FABRIC`, `INVALID_STATE`, `INTERNAL`, `EXISTS`, `UNSUPPORTED_OPERATION`, plus existing `PERMISSION_DENIED` / `FLOW_INVALID`. At least map the ones callers will hit (`INVALID_STATE`, `NO_FABRIC`, `UNSUPPORTED_OPERATION`).
- **`ProviderType::from` panics** on an unknown C enum. `TryFrom` + `Error::Unknown` is safer across SDK upgrades.
- **`InterfaceIter` uses `try_into().ok()`** and skips UTF-8 failures with no log. Prefer surfacing the error so a bad `attr` string cannot drop the only usable NIC.
- **`inner()` on reader/writer** is `#[allow(dead_code)]`. Gate with `#[cfg(feature = "mxl-fabrics-ofi")]`.
- **Versioning** as agreed in the thread, if you want the v1.1 backport to be coherent.

## Nits

- `MxlFabricsAPiHandle` → `MxlFabricsApiHandle`.
- Grain vs samples `add_target` / `make_progress` are copy-paste; samples docs still say `TransferGrain`.
- `Either::Sample` vs `Initiator::Samples`.
- `mod.rs` rustdoc sample ends with four backticks and has a space before `.unwrap()`; `--all-targets` does not run doctests.
- `get_mxl_fabrics_ofi_so_path` error string says `MXL_FABRICS_SO_PATH` but reads `MXL_BUILD_DIR`.
- Typos: “Commited”, “initator”, “happend”; commit subject `wip capabilities integration`; Co-authored-by `gmai..com`.
- `dynamic_link_require_all(false)` for fabrics vs `true` for core — comment why, or keep them the same.
- `examples/common` already exists on main (`setup_logging`); `mod common` in the demo is fine.

mlefebvre1:
Well it was useful, I've addressed most of the issues outlined by the robot.

I just have one thing to say about this comment :
> **What it does not**
> 
> Flow reader/writer are not owned by the fabrics objects. C stores the `mxlFlowReader` / `mxlFlowWriter` at `setup`. Rust `Config` only borrows them for that call. After `setup`, `Initiator` / `Target` have no lifetime and no `Arc` to the flow handle. `FlowReader::Drop` calls `release_flow_reader`, so this compiles and is UB on the C side:
> 
> ```rust
> let reader = instance.create_flow_reader(id)?;
> let initiator = fabrics.create_initiator()?.setup(&initiator::Config::new(iface, &reader))?;
> drop(reader); // releases the C reader
> initiator.specialize(&cfg); // initiator still holds that pointer
> ```
> 
> The tests/demo then `to_grain_reader()` / `to_grain_writer()`, moving the same C pointer into a second Rust type and nulling the first. That aliasing is inherent (fabrics + `GrainReader` both need the reader), but the crate never records it. Dropping the `GrainReader` while the `Initiator` is still live is the same UAF.
> 
> Until initiator/target **own or share** the flow they register (the core pattern), “safe fabrics API” oversells it.

I agree with the safety concern, but I think the comment is slightly imprecise about what is retained.

The C fabrics API does not keep a `mxlFlowReader` / `mxlFlowWriter` handle after `setup()`. During setup it uses the flow object to derive the memory regions, and later fabrics operations use those registered regions rather than the flow handle itself.

So the lifetime issue is not “the initiator/target still stores the flow reader/writer pointer” as such. The real issue is that Rust currently does not encode that the underlying flow-backed shared memory used to create those regions must stay alive for as long as the initiator/target may use them.

That means the problem is still an ownership/lifetime bug on the Rust side, but at the level of the underlying flow memory rather than the specific `FlowReader` / `FlowWriter` wrapper type. The `to_grain_*` / `to_samples_*` conversions are relevant because they move ownership of the same underlying flow handle between Rust wrapper types, while fabrics still depends on the memory that handle keeps alive.

I think the right fix is for `Initiator` / `Target` to retain shared ownership of a common internal flow-liveness object. I agree that the current Rust API does not enforce the required lifetime relationship strongly enough. I will try to improve the compile-time guarantees in that direction.

> Two bindgen runs produce two opaque pointer types, then the high-level code `.cast()`s between them. Sound if both headers agree; one bindgen of `mxl.h` + `fabrics.h` (or a shared alias) would remove the casts.

I agree with this also. I will try to re-work the bindgen part.

mlefebvre1:
I looked into whether we could merge the two bindgen generations into a single one, but unfortunately there is no perfect way to achieve this.

The main issue is that we are using runtime loading (libloading) and have two separate libraries. When using Bindgen's `dynamic_library_name`, only a single library can be specified per builder instance. As a result, we cannot generate bindings for both libraries in the same bindgen invocation. Separate bindgen runs are therefore required to generate bindings for each library independently.

This leads to the issue mentioned earlier: we end up with two separate sets of bindings, where libmxlfabrics generates many of the same types and definitions that are already present in libmxl. As a result, the Rust compiler treats these as distinct types, even though they represent the exact same underlying C types. This means that values must be explicitly cast when passed between the two libraries.

I see two possible approaches to address this:

1.  Generate three bindgen outputs

The first option would be to generate a dedicated bindgen output containing all shared types from both libmxl and libmxlfabrics. Since this bindgen would only contain type definitions, there would be no need to use `dynamic_library_name`, eliminating the duplication of types across the two libraries.

We would then generate two additional bindgen outputs:

One containing only the functions specific to libmxl
One containing only the functions specific to libmxlfabrics

Because both function-specific bindings would reference the same shared type definitions, this approach would eliminate the need for casting between the two libraries.

2. Replace runtime loading with load-time linking

The second option would be to stop using `libloading` and switch to load-time linking, allowing us to generate all bindings in a single bindgen invocation.

However, this approach comes with several drawbacks:

* The safe wrappers would need to be reworked.
* It would introduce a breaking API change.
* We would lose the ability to treat the MXL crate as a plugin-like system.

Considering the drawbacks of option 2, I would lean towards option 1. It addresses the type duplication issue, eliminates the need for casting between identical types generated from different bindgens, and preserves the current runtime-loading architecture.

Any thoughts?

mlefebvre1:
> I looked into whether we could merge the two bindgen generations into a single one, but unfortunately there is no perfect way to achieve this.
> 
> The main issue is that we are using runtime loading (libloading) and have two separate libraries. When using Bindgen's `dynamic_library_name`, only a single library can be specified per builder instance. As a result, we cannot generate bindings for both libraries in the same bindgen invocation. Separate bindgen runs are therefore required to generate bindings for each library independently.
> 
> This leads to the issue mentioned earlier: we end up with two separate sets of bindings, where libmxlfabrics generates many of the same types and definitions that are already present in libmxl. As a result, the Rust compiler treats these as distinct types, even though they represent the exact same underlying C types. This means that values must be explicitly cast when passed between the two libraries.
> 
> I see two possible approaches to address this:
> 
>     1. Generate three bindgen outputs
> 
> 
> The first option would be to generate a dedicated bindgen output containing all shared types from both libmxl and libmxlfabrics. Since this bindgen would only contain type definitions, there would be no need to use `dynamic_library_name`, eliminating the duplication of types across the two libraries.
> 
> We would then generate two additional bindgen outputs:
> 
> One containing only the functions specific to libmxl One containing only the functions specific to libmxlfabrics
> 
> Because both function-specific bindings would reference the same shared type definitions, this approach would eliminate the need for casting between the two libraries.
> 
>     2. Replace runtime loading with load-time linking
> 
> 
> The second option would be to stop using `libloading` and switch to load-time linking, allowing us to generate all bindings in a single bindgen invocation.
> 
> However, this approach comes with several drawbacks:
> 
>     * The safe wrappers would need to be reworked.
> 
>     * It would introduce a breaking API change.
> 
>     * We would lose the ability to treat the MXL crate as a plugin-like system.
> 
> 
> Considering the drawbacks of option 2, I would lean towards option 1. It addresses the type duplication issue, eliminates the need for casting between identical types generated from different bindgens, and preserves the current runtime-loading architecture.
> 
> Any thoughts?

Implemented Option 1 in https://github.com/dmf-mxl/mxl/pull/621/commits

pedro-alves-ferreira:
@mlefebvre1 I agree we should go for option 1. Thanks.

rust/mxl/src/fabrics/interface/config.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+}
+impl TryFrom<FabricsInterfaceConfig> for InterfaceConfig {
+    type Error = crate::Error;
+    fn try_from(value: FabricsInterfaceConfig) -> Result<Self, Self::Error> {
+        let provider = (value.provider as FabricsProvider).try_into()?;
+        let caps = value.caps.into();
+        let endpoint_address = EndpointAddress::from(&value.address);
+        let attr =
+            (!value.attr.is_null()).then(|| unsafe { CString::from_raw(value.attr as *mut i8) });
```
garethsb: Does this mean CString takes ownership? Would the C layer not free this?

mlefebvre1: Indeed, this is wrong.

mlefebvre1: Fixed in https://github.com/dmf-mxl/mxl/pull/621/changes/5eff804193be36d7c3b5d6976e945aa7403767d4

garethsb: Thanks. I see you used `into` as opposed to `to_owned` used in similar code elsewhere in the PR but the effect is the same. :)

rust/mxl/src/fabrics/initiator/grain.rs
```diff
@@ -0,0 +1,74 @@
...
+        Error::from_status(unsafe {
+            self.instance
+                .ctx
+                .api()
+                .fabrics_initiator_add_target(self.instance.inner, target.inner)
+        })
+    }
+
+    /// Remove a target from the initiator. This function is always non-blocking. If any additional communication for a graceful shutdown is
+    /// required it will happend during a call to make_progress*(). It is guaranteed that no new grain transfer operations will
```
garethsb: ⬇️ 
```suggestion
    /// required it will happen during a call to make_progress*(). It is guaranteed that no new grain transfer operations will
```

rust/mxl/src/fabrics/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+//! let mxl_api = mxl::load_api(mxl::config::get_mxl_so_path()) .unwrap();
+//! let instance = mxl::MxlInstance::new(mxl_api, "/dev/shm","").unwrap();
+//!
+//! let mxl_fabrics_api = mxl::load_fabrics_api(mxl::config::get_mxl_fabrics_ofi_so_path());
+//! let fabrics_instance = instance.create_fabrics_instance(&mxl_fabrics_api.unwrap()).unwrap();
+//!
+//! // You can now create Targets and Initiators from the fabrics_instance
+//! let target = fabrics_instance.create_target().unwrap();
+//! let initiator = fabrics_instance.create_initiator().unwrap();
+//! ````
```
garethsb:
~~~suggestion
//! ```
~~~

rust/mxl/src/fabrics/target/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+pub struct Target<S: TargetState> {
+    instance: TargetInstance,
+    _marker: PhantomData<S>,
+}
+//SAFETY: A target is safe to be sent across threads, but it's not thread-safe to use its API functions.
+unsafe impl<S: TargetState> Send for Target<S> {}
+
+pub enum TargetFlavor {
+    Grain(Target<Grain>),
+    Sample(Target<Samples>),
```
garethsb: `TargetFlavor::Samples` maybe, like `InitiatorFlavor::Samples`?

rust/mxl/src/fabrics/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! This module provides the Fabrics API extension for this library. The main type is the
+//! [FabricsInstance], which is used to create Targets and Initiators for remote data transfers.
+//! This module is gated by the `mxl-fabrics-ofi` feature flag.
+//!
+//! # Details
+//! - To get a FabricsInstance, you must create it from MXL instance and a loaded Fabrics API.
+//! ```
+//! let mxl_api = mxl::load_api(mxl::config::get_mxl_so_path()) .unwrap();
```
garethsb: ⬇️
```suggestion
//! let mxl_api = mxl::load_api(mxl::config::get_mxl_so_path()).unwrap();
```

rust/mxl/src/error.rs
```diff
@@ -23,6 +23,27 @@ pub enum Error {
     InvalidArg,
     #[error("Conflict")]
     Conflict,
+
```
garethsb, line 32:
```suggestion
    #[error("Permission denied")]
    PermissionDenied,
    #[error("Flow invalid")]
    FlowInvalid,

```

rust/mxl/src/error.rs
```diff
@@ -47,6 +68,18 @@ impl Error {
             mxl_sys::MXL_ERR_TIMEOUT => Err(Error::Timeout),
             mxl_sys::MXL_ERR_INVALID_ARG => Err(Error::InvalidArg),
             mxl_sys::MXL_ERR_CONFLICT => Err(Error::Conflict),
+
```
garethsb, line 79:
```suggestion
            mxl_sys::MXL_ERR_PERMISSION_DENIED => Err(Error::PermissionDenied),
            mxl_sys::MXL_ERR_FLOW_INVALID => Err(Error::FlowInvalid),

```

garethsb (commented):
I think two other non-Fabrics error codes have been added since the Rust bindings were first implemented, so maybe we could add those here at the same time as the Fabrics ones?

rust/mxl-sys/build.rs
```diff
@@ -69,27 +90,75 @@ fn main() {
...
+                    .iter()
+                    .map(|dir| format!("-I{dir}")),
+            )
+            .header("wrapper-fabrics.h")
+            .allowlist_function("mxlFabrics.*")
+            .blocklist_type(".*")
+            .blocklist_var(".*")
+            .raw_line("use crate::types::*;")
+            .dynamic_library_name("libmxlfabrics")
+            .dynamic_link_require_all(false)
```
garethsb: Why `false` here vs. `true` in the `mxl_bindings`? Do we want to load if there are missing symbols just for the Fabrics API?

rust/mxl/examples/fabrics-demo.rs
```diff
@@ -0,0 +1,539 @@
...
+        match self.initiator {
+            InitiatorFlavor::Grain(initiator) => {
+                initiator.add_target(&target_info)?;
+                // Wait to be connected
+                loop {
+                    if !running.load(atomic::Ordering::SeqCst) {
+                        return Ok(());
+                    }
+
+                    if initiator.make_progress(Duration::from_millis(250)).is_ok() {
```
garethsb: Does this intentionally treat `MXL_ERR_NOT_READY` the same as "real" errors? There are a few other instances in this code.

garethsb: My 🤖 buddy suggests a cute "fix" to `make_progress` return `std::task::Poll` (or use a locally defined `enum Progress { Done, Pending }`) to make "not ready" not be an error. I think it means something like this:

```rust
use std::task::Poll;

impl Initiator<Grain> {
    pub fn make_progress(&self, timeout: Duration) -> Result<Poll<()>> {
        match unsafe {
            self.instance.ctx.api().fabrics_initiator_make_progress_blocking(
                self.instance.inner,
                u16::try_from(timeout.as_millis()).map_err(|_| Error::InvalidArg)?,
            )
        } {
            mxl_sys::MXL_STATUS_OK => Ok(Poll::Ready(())),
            mxl_sys::MXL_ERR_NOT_READY => Ok(Poll::Pending),
            status => Err(Error::from_status(status).unwrap_err()),
        }
    }
    pub fn transfer(&self, grain_index: u64, start_slice: u16, end_slice: u16) -> Result<Poll<()>> {
        // same split: OK → Ready, NOT_READY → Pending, else Error
    }
}
```

so then here:

```suggestion
                    if initiator.make_progress(Duration::from_millis(250))?.is_ready() {
```

rust/mxl/examples/fabrics-demo.rs
```diff
@@ -0,0 +1,539 @@
...
+                                return Ok(());
+                            }
+                            Err(Error::NotReady) => {
+                                // Retry
+                                continue;
+                            }
+                            Err(e) => {
+                                return Err(e);
+                            }
+                        }
```
garethsb: I don't think the 🤖 suggestion to turn `make_progress` into `Poll` helps here, because you still have four separate cases, and even if `NotReady` became `Ok(Poll::Pending)`, you'd still have to match `Err(Error::Interrupted)` separately than other errors.

So could well be unnecessary finessing. 🙌

garethsb (commented):
I spotted one place where I wasn't sure if you wanted to handle "not ready" differently than other errors. My 🤖 buddy wondered whether it would be more idiomatic Rust to map `MXL_ERR_NOT_READY` differently, but I'm not sure it helps...

garethsb:
Almost all smaller review comments have been resolved and the bindgen fix looks sensible to me, so I think it's basically just the lifetime/safety questions remaining, that Michael went through in https://github.com/dmf-mxl/mxl/pull/621#issuecomment-5372857576?

On the lifetime bit, my 🤖 buddy suggested a complete fix could look like: `setup` consumes or `Arc`-shares the flow handle, returns `Grain` vs `Samples` from **that** object, and dropping initiator/target is what releases the flow (or the last `Arc`). 

On the thread safety bit, if it's analysis of the problem is correct, 🤖 recommends just documenting that you shouldn't send a `Target` or `Initiator` to thread A and another one on the same fabrics instance to thread B.

rust/mxl/examples/fabrics-demo.rs
mlefebvre1: I would say using Poll probably communicate the intent better than treating NotReady as an error. But, I'm also fine keeping it as an error, since converting an mxlStatus into a concrete Error or Success is seamless, whereas with Poll you need custom error handling at call sites inside the safe wrapper that might return NotReady.

Open to know what others think about this one.

rust/mxl-sys/build.rs
mlefebvre1: Indeed it should be true, this is leftover from the previous code.

mlefebvre1:
> Almost all smaller review comments have been resolved and the bindgen fix looks sensible to me, so I think it's basically just the lifetime/safety questions remaining, that Michael went through in [#621 (comment)](https://github.com/dmf-mxl/mxl/pull/621#issuecomment-5372857576)?
> 
> On the lifetime bit, my 🤖 buddy suggested a complete fix could look like: `setup` consumes or `Arc`-shares the flow handle, returns `Grain` vs `Samples` from **that** object, and dropping initiator/target is what releases the flow (or the last `Arc`).
> 
> On the thread safety bit, if it's analysis of the problem is correct, 🤖 recommends just documenting that you shouldn't send a `Target` or `Initiator` to thread A and another one on the same fabrics instance to thread B.

I have a draft PR that should fix the flow lifetime issue. It wraps mxlFlowReader or mxlFlowWriter in a managing instance (FlowWriterInstance, FlowReaderInstance) that controls its lifecycle. This wrapper is then shared with the concrete FlowReader/FlowWriter and their flavored variants (GrainWriter/Reader, SampleWriter/Reader) as an Arc<T>.
The destroy function is now a no-op, and the flow's lifetime is instead tied to the Arc<T>. The instance is released once the reference count hits zero. This means, once we have that, in Fabrics we can simply clone that Arc<T> and store it in the Target or Initiator, as it was suggested by 🤖.

https://github.com/mlefebvre1/mxl/pull/3/changes

To keep the changes manageable to review, I would make a separate PR just for that.

rust/mxl/src/fabrics/interface/config.rs
mlefebvre1: Changed to `to_owned` for consistency :)

rust/mxl/examples/fabrics-demo.rs
garethsb: @mlefebvre1 I don't know whether the `Poll` refactor is worth it or not, but on this specific line of code, did you want NotReady and other errors to all be treated the same way?

mlefebvre1: Right, this was a bug. Now fixed in https://github.com/dmf-mxl/mxl/pull/621/changes/ee1ae7b2b15f55833106f01034306946186cba53. Also took the opportunity to clean-up duplicate code paths.

garethsb (approved):
LGTM (review only, not tested) 🙌 @mlefebvre1!

pac-work (approved):

pac-work (changes requested):
Previously I approved, but then I started to play with valgrind a bit and was able to get here:
```
==137469== Possible data race during write of size 8 at 0x4ADB8A8 by thread #1
==137469== Locks held: none
==137469==    at 0x5315F78: mxl::lib::fabrics::ofi::FabricsInstance::createTarget() (in /usr/lib/libmxl-fabrics.so.1.1)
==137469==    by 0x530115A: mxlFabricsCreateTarget (in /usr/lib/libmxl-fabrics.so.1.1)
==137469==    by 0x18938D: mxl_sys::fabrics::libmxlfabrics::fabrics_create_target (fabrics_bindings.rs:256)
==137469==    by 0x181CA4: mxl::fabrics::target::create_target (mod.rs:134)
==137469==    by 0x17F3A2: mxl::fabrics::instance::FabricsInstance::create_target (instance.rs:90)
==137469==    by 0x1640A8: fabrics_instance_unsafe::main (fabrics-instance-unsafe.rs:140)
==137469==    by 0x14A371: core::ops::function::FnOnce::call_once (function.rs:250)
==137469==    by 0x14DFF4: std::sys::backtrace::__rust_begin_short_backtrace (backtrace.rs:166)
==137469==    by 0x1443C3: std::rt::lang_start::{{closure}} (rt.rs:206)
==137469==    by 0x1B2AD3: call_once<(), (dyn core::ops::function::Fn<(), Output=i32> + core::marker::Sync + core::panic::unwind_safe::RefUnwindSafe)> (function.rs:287)
==137469==    by 0x1B2AD3: do_call<&(dyn core::ops::function::Fn<(), Output=i32> + core::marker::Sync + core::panic::unwind_safe::RefUnwindSafe), i32> (panicking.rs:581)
==137469==    by 0x1B2AD3: catch_unwind<i32, &(dyn core::ops::function::Fn<(), Output=i32> + core::marker::Sync + core::panic::unwind_safe::RefUnwindSafe)> (panicking.rs:544)
==137469==    by 0x1B2AD3: catch_unwind<&(dyn core::ops::function::Fn<(), Output=i32> + core::marker::Sync + core::panic::unwind_safe::RefUnwindSafe), i32> (panic.rs:359)
==137469==    by 0x1B2AD3: {closure#0} (rt.rs:175)
==137469==    by 0x1B2AD3: do_call<std::rt::lang_start_internal::{closure_env#0}, isize> (panicking.rs:581)
==137469==    by 0x1B2AD3: catch_unwind<isize, std::rt::lang_start_internal::{closure_env#0}> (panicking.rs:544)
==137469==    by 0x1B2AD3: catch_unwind<std::rt::lang_start_internal::{closure_env#0}, isize> (panic.rs:359)
==137469==    by 0x1B2AD3: std::rt::lang_start_internal (rt.rs:171)
==137469==    by 0x1443A6: std::rt::lang_start (rt.rs:205)
==137469==    by 0x16444D: main (in /home/pac/Sources/MXL/mxl_jonasohland/rust/target/debug/examples/fabrics-instance-unsafe)
==137469== 
==137469== This conflicts with a previous write of size 8 by thread #3
==137469== Locks held: none
==137469==    at 0x5316047: mxl::lib::fabrics::ofi::FabricsInstance::destroyTarget(mxl::lib::fabrics::ofi::TargetWrapper*) (in /usr/lib/libmxl-fabrics.so.1.1)
==137469==    by 0x53013AC: mxlFabricsDestroyTarget (in /usr/lib/libmxl-fabrics.so.1.1)
==137469==    by 0x1893CD: mxl_sys::fabrics::libmxlfabrics::fabrics_destroy_target (fabrics_bindings.rs:264)
==137469==    by 0x181550: <mxl::fabrics::target::TargetInstance as core::ops::drop::Drop>::drop (mod.rs:55)
==137469==    by 0x180B26: core::ptr::drop_in_place<mxl::fabrics::target::TargetInstance> (mod.rs:810)
==137469==    by 0x180CAA: core::ptr::drop_in_place<mxl::fabrics::target::Target<mxl::fabrics::target::states::Initializing>> (mod.rs:810)
==137469==    by 0x140678: core::mem::drop (mod.rs:1003)
==137469==    by 0x16206B: fabrics_instance_unsafe::main::{{closure}} (fabrics-instance-unsafe.rs:112)
==137469==  Address 0x4adb8a8 is 24 bytes inside a block of size 56 alloc'd
==137469==    at 0x4849023: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_helgrind-amd64-linux.so)
==137469==    by 0x5300B6C: mxlFabricsCreateInstance (in /usr/lib/libmxl-fabrics.so.1.1)
==137469==    by 0x18941E: mxl_sys::fabrics::libmxlfabrics::fabrics_create_instance (fabrics_bindings.rs:228)
==137469==    by 0x17F46A: mxl::fabrics::instance::create_instance (instance.rs:27)
==137469==    by 0x181701: mxl::instance::MxlInstance::create_fabrics_instance (instance.rs:253)
==137469==    by 0x1637C4: fabrics_instance_unsafe::main (fabrics-instance-unsafe.rs:95)
==137469==    by 0x14A371: core::ops::function::FnOnce::call_once (function.rs:250)
==137469==    by 0x14DFF4: std::sys::backtrace::__rust_begin_short_backtrace (backtrace.rs:166)
==137469==    by 0x1443C3: std::rt::lang_start::{{closure}} (rt.rs:206)
==137469==    by 0x1B2AD3: call_once<(), (dyn core::ops::function::Fn<(), Output=i32> + core::marker::Sync + core::panic::unwind_safe::RefUnwindSafe)> (function.rs:287)
==137469==    by 0x1B2AD3: do_call<&(dyn core::ops::function::Fn<(), Output=i32> + core::marker::Sync + core::panic::unwind_safe::RefUnwindSafe), i32> (panicking.rs:581)
==137469==    by 0x1B2AD3: catch_unwind<i32, &(dyn core::ops::function::Fn<(), Output=i32> + core::marker::Sync + core::panic::unwind_safe::RefUnwindSafe)> (panicking.rs:544)
==137469==    by 0x1B2AD3: catch_unwind<&(dyn core::ops::function::Fn<(), Output=i32> + core::marker::Sync + core::panic::unwind_safe::RefUnwindSafe), i32> (panic.rs:359)
==137469==    by 0x1B2AD3: {closure#0} (rt.rs:175)
==137469==    by 0x1B2AD3: do_call<std::rt::lang_start_internal::{closure_env#0}, isize> (panicking.rs:581)
==137469==    by 0x1B2AD3: catch_unwind<isize, std::rt::lang_start_internal::{closure_env#0}> (panicking.rs:544)
==137469==    by 0x1B2AD3: catch_unwind<std::rt::lang_start_internal::{closure_env#0}, isize> (panic.rs:359)
==137469==    by 0x1B2AD3: std::rt::lang_start_internal (rt.rs:171)
==137469==    by 0x1443A6: std::rt::lang_start (rt.rs:205)
==137469==    by 0x16444D: main (in /home/pac/Sources/MXL/mxl_jonasohland/rust/target/debug/examples/fabrics-instance-unsafe)
==137469==  Block was alloc'd by thread #1
```
Is it possible, that there is a race condition, if I drop a target from a different thread?

Now I am trying to simplify the example, posting early just to potentially block the merge. My current example is safe code only.

pac-work (approved):
Based on the discussion on the TSC, the mentioned problem will get fixed in the c++ parts.

pac-work:
> Based on the discussion on the TSC, the mentioned problem will get fixed in the c++ parts.

Handled here: https://github.com/dmf-mxl/mxl/pull/696

mlefebvre1:
I would probably wait for https://github.com/dmf-mxl/mxl/pull/704 to merge first and then a small modification to hook the new instances in the Target/Initiator

PR #626, opened 2026-07-24
Fabrics: Regular printout of transfer stats in demo app

jonasohland:
When running the demo app, regularly print out some transfer statistics. Also serves as a way to check if anything is happening at all. Before, after establishing the connection, the app would just go silent.

*Example for granular flow*
2026-07-24-114244_hyprshot

*Example for cont. flow*
2026-07-24-114051_hyprshot

### DIFF
--- a/tools/mxl-fabrics-demo/demo.cpp
+++ b/tools/mxl-fabrics-demo/demo.cpp
@@ -56,6 +56,48 @@ void signal_handler(int)
     g_exit_requested = 1;
 }
 
+struct TransferStats
+{
+    std::chrono::steady_clock::time_point lastLog = std::chrono::steady_clock::now();
+    uint64_t totalTransfers = 0;
+    uint64_t totalUnits = 0;
+    uint64_t intervalTransfers = 0;
+
+    void record(uint64_t units)
+    {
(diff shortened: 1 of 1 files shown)


tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -56,6 +56,48 @@ void signal_handler(int)
     g_exit_requested = 1;
 }
 
+struct TransferStats
+{
+    std::chrono::steady_clock::time_point lastLog = std::chrono::steady_clock::now();
+    uint64_t totalTransfers = 0;
```
vt-tv, line 62: AA style init

vt-tv (changes requested):

mlefebvre1 (approved):
LGTM

tools/mxl-fabrics-demo/demo.cpp
jonasohland: This is a struct member, so auto does not work here :(

vt-tv: oh sorry. should not review code on phone. :D

Issue #627, opened 2026-07-24
[Small Improvement] Fabrics: Clean up the demo app


jonasohland:
### Description

The demo app has not been cleaned up in a while, and is more or less unchanged from our initial draft. It does not follow a lot of the coding guidelines and is overall a somewhat hard to follow and understand piece of code. (Which is bad for a demo application).
We should think about splitting the app into separate target and initiator apps, that are very basic and serve as examples, and a third full proxy app that serves as a reference implementation.

felixpou:
TSC
-Not necessary for v1.1 RC, would be good to have it in v1.1 or soon after the release otherwise.

felixpou:
Including https://github.com/dmf-mxl/mxl/issues/318

felixpou:
TSC
- @jonasohland can look at this in maybe in 3 weeks

felixpou:
TSC

- no updates

felixpou:
TCS
- no updates, maybe next week

felixpou:
TSC
- @jonasohland is this on your radar?

felixpou:
TSC

- Still on @jonasohland todo. When time permit.

PR #628, opened 2026-07-24
Fix FI_EINTR (interrupted) corrupting internal state for no reason

### LINKED ISSUE #624: [Small Improvement] Gracefully handle signal interruptions on blocking completion queue reads
### Description

Right now, when a completion queue read is interrupted by a signal, we throw an exception which sets our internal state-variant into a `valueless_by_exception` state, and the initiator or target can no longer be used.
We should handle this case gracefully so target/initiator can be used after being interrupted by a signal.

### Additional Context

_No response_

jonasohland:
Some providers return FI_EINTR from blocking completion or event queue reads. This was (almost correctly) forwarded to the user through an exception, but leave our internal state variant in `valueless_by_exception`.
FI_EINTR is a recoverable error, so the user should be able to continue making progress it they choose to do so.
This PR fixes that by catching FI_EINTR early and returning it as a valid result.

Fixes #624

### DIFF
--- a/tools/mxl-fabrics-demo/demo.cpp
+++ b/tools/mxl-fabrics-demo/demo.cpp
@@ -468,11 +468,6 @@ public:
                 status = makeProgress(std::chrono::milliseconds(10));
                 if (status == MXL_ERR_INTERRUPTED)
                 {
-                    if (g_exit_requested)
-                    {
-                        return MXL_STATUS_OK;
-                    }
-
                     continue;
                 }
 
@@ -566,11 +561,6 @@ public:
                 status = makeProgress(std::chrono::milliseconds(10));
                 if (status == MXL_ERR_INTERRUPTED)
                 {
-                    if (g_exit_requested)
-                    {
-                        return MXL_STATUS_OK;
-                    }
-
                     continue;
                 }
 
@@ -822,11 +812,6 @@ public:
             }
             else if (status == MXL_ERR_INTERRUPTED)
             {
-                if (g_exit_requested)
-                {
-                    return MXL_STATUS_OK;
-                }
-
                 continue;
             }
             else if (status != MXL_STATUS_OK)
@@ -887,11 +872,6 @@ public:
             }
             else if (status == MXL_ERR_INTERRUPTED)
             {
-                if (g_exit_requested)
-                {
-                    return MXL_STATUS_OK;
-                }
-
                 continue;
             }
             else if (status != MXL_STATUS_OK)
--- a/tools/mxl-fabrics-demo/demo.cpp
+++ b/tools/mxl-fabrics-demo/demo.cpp
@@ -51,9 +51,9 @@ struct Config
     InterfaceSelection interface;
 };
 
-void signal_handler(int)
+void signal_handler(int signal)
 {
-    g_exit_requested = 1;
+    g_exit_requested = signal;
 }
 
 struct TransferStats
@@ -359,7 +359,7 @@ public:
             status = makeProgress(std::chrono::milliseconds(250));
             if (status == MXL_ERR_INTERRUPTED)
             {
-                return MXL_STATUS_OK;
+                return status;
             }
 
             if (status != MXL_ERR_NOT_READY && status != MXL_STATUS_OK)
@@ -468,7 +468,7 @@ public:
                 status = makeProgress(std::chrono::milliseconds(10));
                 if (status == MXL_ERR_INTERRUPTED)
                 {
-                    return MXL_STATUS_OK;
+                    return status;
                 }
 
                 if (status != MXL_ERR_NOT_READY && status != MXL_STATUS_OK)
@@ -561,7 +561,7 @@ public:
                 status = makeProgress(std::chrono::milliseconds(10));
                 if (status == MXL_ERR_INTERRUPTED)
                 {
-                    return MXL_STATUS_OK;
+                    return status;
                 }
 
                 if (status != MXL_ERR_NOT_READY && status != MXL_STATUS_OK)
@@ -812,7 +812,7 @@ public:
             }
             else if (status == MXL_ERR_INTERRUPTED)
             {
-                return MXL_STATUS_OK;
+                return status;
             }
             else if (status != MXL_STATUS_OK)
             {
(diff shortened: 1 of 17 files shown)


vt-tv (approved):
LGTM.

mlefebvre1 (approved):
LGTM

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -1004,6 +1004,11 @@ int main(int argc, char** argv)
 
         if (status = app.run(); status != MXL_STATUS_OK)
         {
+            if (status == MXL_ERR_INTERRUPTED)
+            {
+                return 128 + g_exit_requested;
```
KimonHoffmann: This seems flawed to me:
1. When the app is requested to terminate by means of `SIGTERM` or `SIGINT` it should just exit with success error code.
2. The the app is *not* requested to terminate a result of `INTERRUPTED` should not be treated as an error and the app should just continue. Interruptions can happen for a variety of reasons and are not errors.

The best remediation from my perspective would be:
1. Rename `run` to `runOnce` and make it perform exactly one iteration, without checking for `g_exit_requested`.
3. Change this if to a loop like the one sketched out below and move the handling of the various status codes from the `run*()` methods to there.
4. Adapt the error handling to ignore non-lethal errors.

Loop outline:
```cpp
while (!g_exitRequested)
{
    auto const status = app.runOnce();
    if (isLethalError)
    {
        // Report and exit
    }
}
```

Alternativele: Ignore `INTERRUPTED` and other non-lethal errors in the `run*()` loops that already exist and only terminate `run()` upon lethal conditions.

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -1058,6 +1063,11 @@ int main(int argc, char** argv)
...
+            }
```
KimonHoffmann: As above.

KimonHoffmann (changes requested):
Looks good, but the handling of the new status in the demo app appears off to me.

tools/mxl-fabrics-demo/demo.cpp
jonasohland: The intention was to mimic the behaviour of bash when a process is killed because of an uncaught signal. But the more I think about it, I can see that this is more confusing than anything else. So I will return 0 after catching an error signal.

KimonHoffmann: The signal is in this case not uncaught, as `SIGINT` and `SIGTERM` are explicitly handled in order to perform an orderly shutdown, but also `EINTR` is not an error indication, it's just an information that the operation was not completed due to some kind of interruption (which could be due to a signal -- handled or not) and the appropriate thing do is to just resume the operation of there isn't anything more urgent to do.

The common handling pattern of `SIGTERM` is a good example of this. If a blocking syscall (like a read with an infinite timeout) would not return early with `EINTR` applications would not be able to perform an orderly shutdown, because while the handler may be called immediately the main loop of the program would not be able to wind down as a result.

So in this case
* the signal handler commonly sets a shutdown flag,
* the main loop gets interrupted, and before continuing with the next iteration, where it resumes the previously interrupted operation
* checks the flag once and exits as a result (which would be "more urgent [thing] to do" mentioned above).

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -812,7 +822,12 @@ class AppTarget
             }
             else if (status == MXL_ERR_INTERRUPTED)
             {
-                return status;
+                if (g_exit_requested)
```
KimonHoffmann: This is superfluous, as the while loop is already predicated on `!g_exit_requested`.

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -872,7 +887,12 @@ class AppTarget
             }
             else if (status == MXL_ERR_INTERRUPTED)
             {
-                return MXL_STATUS_OK;
+                if (g_exit_requested)
```
KimonHoffmann: As above.

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -561,7 +566,12 @@ class AppInitator
...
-                    return status;
+                    if (g_exit_requested)
+                    {
+                        return MXL_STATUS_OK;
+                    }
+
+                    continue;
                 }
 
                 if (status != MXL_ERR_NOT_READY && status != MXL_STATUS_OK)
```
KimonHoffmann, line 567: Could use parentheses around the binary sub-expressions.

tools/mxl-fabrics-demo/demo.cpp
KimonHoffmann, line 479: Could use parentheses around the binary sub-expressions.

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -468,7 +468,12 @@ class AppInitator
...
```
KimonHoffmann, line 474: Could use parentheses around the binary sub-expressions.

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -468,7 +468,12 @@ class AppInitator
                 status = makeProgress(std::chrono::milliseconds(10));
                 if (status == MXL_ERR_INTERRUPTED)
                 {
-                    return status;
+                    if (g_exit_requested)
+                    {
+                        return MXL_STATUS_OK;
+                    }
```
KimonHoffmann: While I understand that there is some cost to setup and teardown of this run-loop I think it'd be more appropriate to just return with `MXL_ERR_INTERRUPTED` and let the caller decide how to go about this.

tools/mxl-fabrics-demo/demo.cpp
```diff
@@ -561,7 +566,12 @@ class AppInitator
...
```
KimonHoffmann: As above wrt to forwarding the error to the caller.

KimonHoffmann (changes requested):
A few subtleties regarding the handling of interruptions.

jonasohland:
@KimonHoffmann I am going to re-write this demo app anyway (#627), so I just want to get this fix merged with not too many other changes that will conflict with my rewrite... So maybe we can just merge this as-is? I removed the superfluous `g_exit_requested` check.

KimonHoffmann:
> @KimonHoffmann I am going to re-write this demo app anyway (#627), so I just want to get this fix merged with not too many other changes that will conflict with my rewrite... So maybe we can just merge this as-is? I removed the superfluous `g_exit_requested` check.

Alight, that's fine by me.

KimonHoffmann (approved):
Good to go, thank you!

PR #631, opened 2026-07-24
mxl-info: Stop lying about latency!

### LINKED ISSUE #413: [Small Improvement] Remove Last [Read/Write] time from mxl-info output for cont. flows
### Description

These fields in the runtime info are not updated by cont. flow writers. We should hide them in the output of `mxl-info` to avoid confusion.

### Additional Context

_No response_

jonasohland:
The latency number pretends to be accurate, but is calculated from the index (which is based on the flow rate) against a nanosecond time. So the latency we show there is just one possible latency in a range.
Let's show that range instead!

Also remove the output of last read/write time from the output for continuous flows. They are never updated an could be misleading.

Fixes #413

### DIFF
--- a/tools/mxl-info/main.cpp
+++ b/tools/mxl-info/main.cpp
@@ -136,7 +136,8 @@ namespace
                 latencyGrains = headIndex - currentIndex;
             }
 
-            auto const signedDiffMs = [](std::uint64_t a, std::uint64_t b) -> double {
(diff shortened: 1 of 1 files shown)


jonasohland:
*Example Output*
2026-07-24-195416_hyprshot

vt-tv (approved):
Nice idea.

mlefebvre1 (approved):

PR #633, opened 2026-07-26
Fix INSTALL_RPATH target for mxl-fabrics-demo

k0zmo:
Correct a copy-paste error introduced in ecd877e0 that applied the mxl-fabrics-demo INSTALL_RPATH to mxl-data-probe instead.


KimonHoffmann (commented):
Good catch, thank you!

KimonHoffmann (approved):

vt-tv (approved):
lgtm

PR #643, opened 2026-08-03
Fix continuous flow futex notification

### LINKED ISSUE #642: Continuous flow writers are not notifying downstream reader correctly.
In `PosixContinuousFlowWriter` the per-writer `_currentIndex` is set to the undefined value before it is used later in `signalCompletedBatch()` to calculate if the futex should receive a wake-up or not.
This sometimes is not an issue and readers that use the non-blocking API will not observe this issue.

[bug here](https://github.com/dmf-mxl/mxl/blob/4a8a50e76dfe55892348c737774c29d7c0f881c5/lib/internal/src/PosixContinuousFlowWriter.cpp#L114)

jonasohland:
Fixes #642 

The issue can be reproduced by running the tests on commit 7cfaeb0c which is that adds a specific test for the bug without the fix. 5f8052b7 adds the fix.

### DIFF
--- a/lib/internal/src/PosixContinuousFlowWriter.cpp
+++ b/lib/internal/src/PosixContinuousFlowWriter.cpp
@@ -111,7 +111,6 @@ namespace mxl::lib
         {
             auto const flow = _flowData->flow();
             flow->info.runtime.headIndex = _currentIndex;
-            _currentIndex = MXL_UNDEFINED_INDEX;
 
             if (signalCompletedBatch())
             {
@@ -120,6 +119,7 @@ namespace mxl::lib
                 wakeAll(&flow->state.syncCounter);
             }
 
+            _currentIndex = MXL_UNDEFINED_INDEX;
             return MXL_STATUS_OK;
         }
         else
--- a/lib/internal/src/PosixContinuousFlowWriter.cpp
+++ b/lib/internal/src/PosixContinuousFlowWriter.cpp
@@ -111,6 +111,7 @@ namespace mxl::lib
         {
             auto const flow = _flowData->flow();
             flow->info.runtime.headIndex = _currentIndex;
+            _currentIndex = MXL_UNDEFINED_INDEX;
 
             if (signalCompletedBatch())
             {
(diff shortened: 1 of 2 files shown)


lapointejp (approved):
LGTM.
If @KimonHoffmann has some time to review his eyes would be relevant here I think.

KimonHoffmann (approved):
Thank you for finding and fixing this!
The only question I'd have is whether it is more correct to change `signalCompletedBatch()` to use `flow->info.runtime.headIndex` instead of `_currentIndex` instead?!

lib/internal/src/PosixContinuousFlowWriter.cpp
```diff
@@ -120,6 +119,7 @@ namespace mxl::lib
                 wakeAll(&flow->state.syncCounter);
             }
 
+            _currentIndex = MXL_UNDEFINED_INDEX;
```
pedro-alves-ferreira: I think this deserves a comment like:

```
// Note that this must be done _after_ the call to `signalCompletedBatch`.
```

But even with that is still makes me feel uneasy, as it looks too brittle. 

I wonder if the correct fix wouldn't be to use `flow->info.runtime.headIndex` in line 140 and only use `_currentIndex` to signal that a range is currently open. Because the `headIndex` changing is what actually triggers the notification.

pedro-alves-ferreira: 😄 It looks like Kimon beat me to it.

jonasohland:
> Thank you for finding and fixing this! The only question I'd have is whether it is more correct to change `signalCompletedBatch()` to use `flow->info.runtime.headIndex` instead of `_currentIndex` instead?!

Very reasonable. I will do that.

jonasohland:
@KimonHoffmann @pedro-alves-ferreira Changed it to use the head index from the flow data.

pedro-alves-ferreira:
> @KimonHoffmann @pedro-alves-ferreira Changed it to use the head index from the flow data.

Thanks!

KimonHoffmann:
Perfect, thank you!

lib/internal/src/PosixContinuousFlowWriter.cpp
KimonHoffmann, line 149: Oh, this missed another occurrence of `_currentIndex`.

PR #664, opened 2026-08-07
Rust: Fix prototype of mkdtemp in gst-mxl-rs tests to avoid a compilation error on arm64

KimonHoffmann:
[//]: # (SPDX-FileCopyrightText: 2026 Contributors to the Media eXchange Layer project.)
[//]: # (SPDX-License-Identifier: Apache-2.0)

# Details

The current native prototype of `mkdtemp` in one of the rust tests was falsely prototyped as taking a signed byte pointer instead of a character pointer. This caused a type mismatch, when compiling for linux/arm64.

@pac-work provided this fix, which I confirmed to solve the problem, while having no impact on the compilation on linux/x86_64.

# Backport requirements

Should be back ported to v1.1, not sure if the offending code is also present in v1.0.


pac-work (approved):

pedro-alves-ferreira (approved):
LGTM

PR #668, opened 2026-08-13
gst-avsynctest-rs: fix AncillaryMeta DC parity bits

garethsb:
## Summary
- `AncillaryMeta::set_data` only writes DC’s low 8 bits and clears the parity nibble; packing computed a parity-extended DC for the checksum but never called `set_data_count_upper_two_bits`, so attached metas were invalid.
- Centralise SMPTE 291 packing and validity checks in `gst-avsynctest-rs::ancillary`; point `gst-mxl-rs` tests at that helper instead of the duplicate in `tests/common`.
- Add a regression test covering odd and even DC bit counts (`add_ancillary_meta_produces_valid_meta`), plus end-to-end `video_ancillary_metas_valid`.

## Test plan
- [x] `cargo fmt -p gst-avsynctest-rs -- --check`
- [x] `cargo clippy -p gst-avsynctest-rs --all-targets -- -D warnings`
- [x] `cargo test -p gst-avsynctest-rs -- --test-threads=1` (includes `add_ancillary_meta_produces_valid_meta`)

## Backport requirements

Should be backported for v1.1 please

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -10,6 +10,7 @@ resolver = "2"
 edition = "2024"
 publish = false
 version = "0.1.0"
+rust-version = "1.87"
 license = "Apache-2.0"
 license-file = "../LICENSE.txt"
 repository = "https://github.com/dmf-mxl/mxl"
--- a/rust/gst-avsynctest-rs/src/ancillary.rs
+++ b/rust/gst-avsynctest-rs/src/ancillary.rs
@@ -1,0 +1,175 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-FileCopyrightText: 2026 Contributors to the Media eXchange Layer project.
+// SPDX-License-Identifier: Apache-2.0
+
+//! SMPTE 291 ancillary packing and `GstAncillaryMeta` validity helpers.
+
+use glib;
+use gstreamer as gst;
+use gstreamer_video as gst_video;
+
+/// Extend an 8-bit value to a 10-bit ST 291 word with even/odd parity.
+///
+/// Per SMPTE ST 291: b8 is even parity over b7 to b0 (so b0 to b8 have an even
+/// number of 1s), and b9 = NOT b8. Matches `SET_WITH_PARITY` in
+/// gst-plugins-base `video-anc.c`: odd popcount gives `0x100 | v`, even gives
+/// `0x200 | v`.
+fn extend_with_even_odd_parity(v: u8) -> u16 {
+    if v.count_ones() & 1 != 0 {
+        // Odd number of ones in v: b8 = 1, b9 = 0
+        0x1_00 | (v as u16)
+    } else {
+        // Even number of ones in v: b8 = 0, b9 = 1
+        0x2_00 | (v as u16)
+    }
+}
+
+fn has_even_odd_parity(w: u16) -> bool {
+    w == extend_with_even_odd_parity((w & 0xff) as u8)
+}
+
+/// SMPTE 291 checksum over the 10-bit DID, SDID, DC and UDW words.
+fn ancillary_checksum(did: u16, sdid: u16, dc: u16, data: &[u16]) -> u16 {
+    let mut checksum = 0u16;
+    checksum = checksum.wrapping_add(did & 0x1ff);
+    checksum = checksum.wrapping_add(sdid & 0x1ff);
+    checksum = checksum.wrapping_add(dc & 0x1ff);
+    for &w in data {
+        checksum = checksum.wrapping_add(w & 0x1ff);
+    }
+    checksum &= 0x1ff;
+    // b9 = NOT b8 (ST 291 checksum word)
+    if checksum & 0x1_00 == 0 {
+        checksum |= 0x2_00;
+    }
+    checksum
+}
+
+/// Attach one `GstAncillaryMeta` carrying 8-bit `payload` under `did`/`sdid` on
+/// `line`/`offset`, as 10-bit even/odd-parity words plus checksum.
+pub fn add_ancillary_meta(
+    buffer: &mut gst::BufferRef,
+    line: u16,
+    offset: u16,
+    did: u8,
+    sdid: u8,
+    payload: &[u8],
+) {
+    let mut meta = gst_video::video_meta::AncillaryMeta::add(buffer);
+    meta.set_c_not_y_channel(false);
+    meta.set_line(line);
+    meta.set_offset(offset);
+    let did_10bit = extend_with_even_odd_parity(did);
+    let sdid_10bit = extend_with_even_odd_parity(sdid);
+    let dc_10bit = extend_with_even_odd_parity(payload.len() as u8);
+    meta.set_did(did_10bit);
+    meta.set_sdid_block_number(sdid_10bit);
+    let data: Vec<u16> = payload
+        .iter()
+        .copied()
+        .map(extend_with_even_odd_parity)
+        .collect();
+    meta.set_data(glib::Slice::from(data));
+    // set_data writes only the length into data_count's low 8 bits and clears
+    // the upper two; restore even/odd parity so DC is a valid 10-bit word.
+    meta.set_data_count_upper_two_bits((dc_10bit >> 8) as u8);
+    let checksum = ancillary_checksum(
+        meta.did(),
+        meta.sdid_block_number(),
+        meta.data_count(),
+        meta.data(),
+    );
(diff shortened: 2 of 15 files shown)


rust/gst-avsynctest-rs/src/ancillary.rs
```diff
@@ -0,0 +1,111 @@
...
+    meta.set_sdid_block_number(sdid_10bit);
+    let data: Vec<u16> = payload
+        .iter()
+        .copied()
+        .map(extend_with_even_odd_parity)
+        .collect();
+    meta.set_data(glib::Slice::from(data));
+    // set_data writes only the length into data_count's low 8 bits and clears
+    // the upper two; restore even/odd parity so DC is a valid 10-bit word.
+    meta.set_data_count_upper_two_bits((dc_10bit >> 8) as u8);
```
garethsb, line 75: :eyes: This is the one-line fix... but I took the opportunity to increase the test coverage and de-dupe the ancillary meta functions that were in `src/videosrc/imp.rs` and `tests/common/mod.rs`. (And bring in some doc-comment additions.)

rust/gst-avsynctest-rs/src/ancillary.rs
```diff
@@ -0,0 +1,111 @@
...
+//! SMPTE 291 ancillary packing and `GstAncillaryMeta` validity helpers.
+
+use glib;
+use gstreamer as gst;
+use gstreamer_video as gst_video;
+
+/// Extend an 8-bit value to a 10-bit even/odd-parity word (b8 = even parity of
+/// b7–b0, b9 = not b8).
+fn extend_with_even_odd_parity(v: u8) -> u16 {
+    if v.count_ones() & 1 == 0 {
```
k0zmo: The branches should be swapped. Take for example `0x61`, it has odd (3) ones and should be `0x161` - but with current code it becomes `0x261`

rust/gst-avsynctest-rs/src/ancillary.rs
```diff
@@ -0,0 +1,111 @@
...
+        meta.data(),
+    );
+    meta.set_checksum(checksum);
+}
+
+/// SDI forbidden values in a 10-bit word: TRS/sync codes `0x000`–`0x003` and
+/// `0x3FC`–`0x3FF`. UDWs must not use these; DID/SDID/DC with even/odd parity
+/// already avoid them.
+fn is_sdi_reserved_sync_word(w: u16) -> bool {
+    let w = w & 0x3ff;
```
k0zmo: This function takes any `u16` or 10-bit value represented as `u16`? If former, I'd add a defensive check `w <= 0x3fff`

k0zmo (changes requested):

rust/gst-avsynctest-rs/src/ancillary.rs
garethsb: Ouch. You're right. Thanks.

garethsb: I realised the upstream **gst-plugins-rs** also had it wrong - I've submitted a fix ([gstreamer/gst-plugins-rs!3216](https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/merge_requests/3216)) which I was able to get merged quickly.

garethsb: Fixed, thanks!

rust/gst-avsynctest-rs/src/ancillary.rs
garethsb: Reversed polarity and now returns false for `w > 0x3ff`

rust/Cargo.toml
```diff
@@ -10,6 +10,7 @@ resolver = "2"
 edition = "2024"
 publish = false
 version = "0.1.0"
+rust-version = "1.87"
```
pedro-alves-ferreira, line 13: Do we want to set the version here? It will be a bit of a pain to chase all those. 
Would it be better to use a [toolchain file](https://share.google/I6sjhH3xdap9eDPrV)?

rust/gst-avsynctest-rs/src/ancillary.rs
```diff
@@ -0,0 +1,175 @@
...
+    w == extend_with_even_odd_parity((w & 0xff) as u8)
+}
+
+/// SMPTE 291 checksum over the 10-bit DID, SDID, DC and UDW words.
+fn ancillary_checksum(did: u16, sdid: u16, dc: u16, data: &[u16]) -> u16 {
+    let mut checksum = 0u16;
+    checksum = checksum.wrapping_add(did & 0x1ff);
+    checksum = checksum.wrapping_add(sdid & 0x1ff);
+    checksum = checksum.wrapping_add(dc & 0x1ff);
+    for &w in data {
```
pedro-alves-ferreira, line 37: This ampersand looks odd to me. What is the goal?

rust/gst-avsynctest-rs/src/ancillary.rs
```diff
@@ -0,0 +1,175 @@
...
+}
+
+fn has_even_odd_parity(w: u16) -> bool {
+    w == extend_with_even_odd_parity((w & 0xff) as u8)
+}
+
+/// SMPTE 291 checksum over the 10-bit DID, SDID, DC and UDW words.
+fn ancillary_checksum(did: u16, sdid: u16, dc: u16, data: &[u16]) -> u16 {
+    let mut checksum = 0u16;
+    checksum = checksum.wrapping_add(did & 0x1ff);
```
pedro-alves-ferreira, line 34: I'm not sitting at my desk, so I won't try to suggest an alternative, but not using an algorithm like accumulate and repeating 0x1FF irks me a bit 🙂 But, I don't think it's worth it to spend time on it and this is just fine.

rust/gst-avsynctest-rs/src/ancillary.rs
```diff
@@ -0,0 +1,175 @@
...
+        return false;
+    }
+    if !data.iter().all(|&w| is_permitted_word(w)) {
+        return false;
+    }
+    meta.checksum() == ancillary_checksum(did, sdid, dc, data)
+}
+
+/// Whether every `GstAncillaryMeta` on `buffer` passes [`ancillary_meta_valid`].
+pub fn ancillary_metas_valid(buffer: &gst::BufferRef) -> bool {
```
pedro-alves-ferreira: How about `are_ancillary_metas_valid`, to start with a verb and be consistent with `has_` above.

pedro-alves-ferreira (approved):
LGTM, with a couple of comments.

rust/gst-avsynctest-rs/src/ancillary.rs
garethsb: My 🤖 buddy told me it was idiomatic to make the deref clear for the `w & 0x1ff`, but I'm happy to delete it.

rust/gst-avsynctest-rs/src/ancillary.rs
garethsb: Yep, agreed it could be nice with an accumulator struct... but this is the simplest readable implementation and aligned with https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/blob/main/video/closedcaption/src/st2038anc_utils.rs?ref_type=heads#L126-149

rust/gst-avsynctest-rs/src/ancillary.rs
garethsb: I'll change to `is_valid_ancillary_meta` and `has_valid_ancillary_metas` 🙂

rust/Cargo.toml
garethsb: I thought the `Cargo.toml` indicates the MSRV, whereas `rust-toolchain.toml` pins the actual version used to build? If we want the MSRV to easily ripple down to most crates, they would each need `rust-version.workspace = true`. I'll do that.

(I prefer `rust-version = "1.87"` explicitly in the `gst-avsynctest-rs/Cargo.toml` precisely because I'm maintaining that across two repos which had different MSRV, which meant I got different clippy warnings from both.)

k0zmo (approved):

vt-tv (approved):
Approving on behalf of @k0zmo

rust/Cargo.toml
pedro-alves-ferreira: Absolutely. My bad.

rust/gst-avsynctest-rs/src/ancillary.rs
pedro-alves-ferreira: I see. I neglected the fact that you would have to explicitly dereference `w` below.

PR #672, opened 2026-08-14
Fixes for ease of packaging

iSchluff:
# Details
Two fixes to ease packaging of libmxl
- fix one is a basic missing include (not sure why it doesn't blow up in upstream builds)
- fix two is removing the dependency on a stduuid build specifically with UUID_SYSTEM_GENERATOR enabled. This is out-of-scope for the stdlib proposal and not enabled everywhere by default. As it only affects tests/utils right now I think this is fine.

# Backport requirements
The domainwatcher include would be nice2have in 1.0

### DIFF
--- a/lib/tests/fabrics/test_basics.cpp
+++ b/lib/tests/fabrics/test_basics.cpp
@@ -7,6 +7,7 @@
 #include <cstdint>
 #include <cstring>
 #include <algorithm>
+#include <random>
 #include <vector>
 #include <uuid.h>
 #include <catch2/catch_message.hpp>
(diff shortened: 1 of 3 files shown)


jonasohland (approved):
Looks good to me.

pedro-alves-ferreira (approved):
LGTM

lib/tests/fabrics/test_basics.cpp
```diff
@@ -711,7 +712,8 @@ TEST_CASE_PERSISTENT_FIXTURE(mxl::tests::mxlDomainFixture, "Fabrics: Transfer Gr
     for (size_t i = 0; i < nbTargets; i++)
     {
         REQUIRE(mxlFabricsCreateTarget(fabrics, &targets[i]) == MXL_STATUS_OK);
-        flowIds[i] = uuids::to_string(uuids::uuid_system_generator{}());
+        auto gen = std::mt19937{std::random_device{}()};
```
garethsb: I think this is a bit of an anti-pattern. You want to construct the `random_device` and the `mt19937` engine (it's not the `gen`erator, btw), and probably the `uuid_random_generator`, once and use it repeatedly. Constructing them new for each UUID is bad for randomness and performance. Can we have just one generator per test case please?

garethsb (changes requested):

lib/tests/fabrics/test_basics.cpp
iSchluff: now reuses engines and generators.

garethsb (approved):
Thanks, @iSchluff.

PR #678, opened 2026-08-19
gst-mxl-rs: no-op compatible mid-stream set_caps on mxlsink

garethsb:
## Details

`mxlsink` used to treat every mid-stream `set_caps` as a request to create a new flow writer. Compatible CAPS updates (the case I experienced: extra `channel-mask` / `channel-order`) therefore failed while an active writer already owned the UUID.

**Before:** re-`set_caps` always called `create_flow_writer` again and failed with:

```
The writer could not be created, the UUID belongs to a flow with another active writer
```

**Now:**
- If the stored `FlowDef` details are unchanged → no-op: keep the existing writer (`set_caps` succeeds).
- If the media kind is the same but format details changed, or the media type changed (video/audio/data) → fail immediately with a dedicated error instead of attempting recreate:
  - `mxlsink does not support mid-stream format changes after the flow writer is created` (not very likely)
  - `mxlsink does not support mid-stream media type changes after the flow writer is created` (very unlikely!)

## Backport requirements

Version 1.1

### DIFF
--- a/rust/gst-mxl-rs/src/mxlsink/state.rs
+++ b/rust/gst-mxl-rs/src/mxlsink/state.rs
@@ -157,6 +157,14 @@ pub(crate) enum DiscreteFormat {
 pub(crate) struct DiscreteState {
     pub format: DiscreteFormat,
     pub writer: GrainWriter,
+    pub flow_def: DiscreteFlowDef,
+}
+
+/// Video vs data details stored alongside the discrete writer.
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) enum DiscreteFlowDef {
+    Video(FlowDefVideo),
+    Data(FlowDefData),
 }
 
 pub(crate) struct ContinuousState {
@@ -168,6 +176,12 @@ pub(crate) struct ContinuousState {
 pub(crate) struct Context {
     pub state: Option<State>,
 }
+
+const RE_SET_CAPS_NO_CHANGE: &str = "set_caps: flow_def unchanged; keeping existing writer";
+const RE_SET_CAPS_FORMAT_CHANGE: &str =
+    "mxlsink does not support mid-stream format changes after the flow writer is created";
+const RE_SET_CAPS_MEDIA_TYPE_CHANGE: &str =
+    "mxlsink does not support mid-stream media type changes after the flow writer is created";
 
 pub(crate) fn init_state_with_video(
     state: &mut State,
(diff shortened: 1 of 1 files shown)


rust/gst-mxl-rs/src/mxlsink/state.rs
```diff
@@ -228,6 +242,28 @@ pub(crate) fn init_state_with_video(
             },
         ],
     };
+    if let Some(prev_state) = &state.flow_state {
+        match prev_state {
```
pedro-alves-ferreira: I wonder if it would be clearer if you extracted the `return` out of the `match`, to make it clear that the function unconditionally returns early if `prev_state` is some?

```rust
return match prev_state {
(...)
};
```

rust/gst-mxl-rs/src/mxlsink/state.rs
```diff
@@ -293,6 +329,24 @@ pub(crate) fn init_state_with_audio(
         channel_count: channels,
         bit_depth,
     };
+    if let Some(prev_state) = &state.flow_state {
+        match prev_state {
```
pedro-alves-ferreira: Same comment about the early return.

rust/gst-mxl-rs/src/mxlsink/state.rs
```diff
@@ -358,6 +408,28 @@ pub(crate) fn init_state_with_data(
             denominator: framerate.denom(),
         },
     };
+    if let Some(prev_state) = &state.flow_state {
+        match prev_state {
+            FlowState::Discrete(prev_discrete) => match &prev_discrete.flow_def {
```
pedro-alves-ferreira, line 415: Ditto.

rust/gst-mxl-rs/src/mxlsink/state.rs
```diff
@@ -391,8 +463,36 @@ pub(crate) fn init_state_with_data(
     state.flow_state = Some(FlowState::Discrete(DiscreteState {
         format: DiscreteFormat::Data,
         writer,
+        flow_def: DiscreteFlowDef::Data(flow_def_details),
     }));
     state.flow_config = Some(flow);
 
     Ok(())
 }
+
```
pedro-alves-ferreira: Do you want to leave this here or was it just an experiment?

pedro-alves-ferreira (approved):
LGTM, with a few comments.

rust/gst-mxl-rs/src/mxlsink/state.rs
garethsb: Yep, orphaned code after proving out the cause of the issue I was seeing. Deleting...

rust/gst-mxl-rs/src/mxlsink/state.rs
garethsb: Yes, that's much clearer, thanks.

mlefebvre1 (approved):

vt-tv:
@garethsb  if this should be backported please add the required labels.

PR #690, opened 2026-09-01
Day 1 Experience: Add a curl|bash bootstrap for the Docker Compose examples

garethsb:
https://dmfmxl.slack.com/archives/C0812T4044B/p1787912169550229

## Details

- Adds `examples/bootstrap.sh` so a Linux x86_64 host can fetch sources, install Docker Engine and Compose if missing, and `docker compose build` / `up` the example stack without a prior clone or compiler on the host.
- Documents the one-liner in the top-level README and in `examples/README.md`, including distro coverage, `--yes` / `--up` / `--skip-docker`, and how to check logs and stop.
- Adds a repo-root `.dockerignore` so host `build/` / `target/` trees are not copied into the example image (a copied `CMakeCache.txt` breaks the in-container CMake preset).
- Follow-up: when there is no `.git` (GitHub tarball / `git archive`), CMake reads the short hash from `cmake/git-hash.txt` (`export-subst`) so `MXL_VERSION_FULL` is not always `unknown`.

## 🤖-suggested test plan

- [x] From a checkout: `./examples/bootstrap.sh --up` (Docker already present; confirm it skips install and builds/runs).
- [x] Piped: `curl -fsSL …/examples/bootstrap.sh | bash -s -- --yes --up` in an empty directory; confirm `./mxl` is unpacked and Compose comes up.
- [x] `docker logs mxl-example-video-flow-info-1` shows updating flow info; `Ctrl+C` then `docker compose down` from `examples/`.
- [ ] Repeat with Docker missing on one of: Ubuntu/Debian, Fedora/RHEL/Rocky, AlmaLinux, Arch (sudo password prompt, then install + build).
- [x] `--skip-docker` on a machine without Docker exits cleanly; `--yes` skips confirmation prompts but still allows sudo to ask for a password.
- [x] Second run with the same `MXL_REF` reuses the stamped tree; a different `MXL_REF` / `MXL_GITHUB_REPO` does not reuse a mismatched tree.
- [x] Host `build/` present during `docker compose build` does not fail CMake with a `CMakeCache.txt` path mismatch.
- [x] Tarball/archive build: `MXL_VERSION_FULL` includes a 12-char hash, not `unknown`. Checkout build still uses `git rev-parse`.

## Backport requirements

Version 1.1


vt-tv:
LGTM.  Does it need to be backported? if so please add the labels.

vt-tv (approved):
LGTM

PR #695, opened 2026-09-03
fix gst-mxl-rs audio reader spinning

jonasohland:
Found this while building [my domain explorer](https://github.com/jonasohland/mxl-domain-explorer). I might be missing the intention, and maybe also misunderstanding the continuous flow half buffer thing, but without this fix, all my gstreamer `mxlsrc` elements eat 1 entire cpu.

### DIFF
--- a/rust/gst-mxl-rs/src/mxlsrc/create_continuous.rs
+++ b/rust/gst-mxl-rs/src/mxlsrc/create_continuous.rs
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: 2025-2026 Contributors to the Media eXchange Layer project.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::time::{Duration, Instant};
+use std::time::Duration;
 
 use crate::mxlsrc::imp::{CreateState, MxlSrc};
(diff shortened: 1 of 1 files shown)


vt-tv (approved):
LGTM

rust/gst-mxl-rs/src/mxlsrc/create_continuous.rs
```diff
@@ -52,23 +51,16 @@ pub(crate) fn create_continuous(
...
-    wait_for_sample(head, batch, continuous_state)?;
 
     if is_reader_late(head, batch, ring, continuous_state)? {
         resync_state(continuous_state);
     }
 
-    // `get_samples(end, count)` returns the `count` samples at absolute indices
-    // `[end - count, end)` (last is `end - 1`). `continuous_state.index` is the first
-    // sample we want, so pass `index + batch` as the end; `pts_for_index(index)`
-    // then stamps that first returned sample.
```
garethsb, line 64: Could we keep this comment in some form as it helps to document that we are end-exclusive?

garethsb (approved):

PR #704, opened 2026-09-04
Rust: manage flow instance lifecycle automatically

### LINKED ISSUE #182: Fabrics: Rust bindings

mlefebvre1:
# Details

Add shared wrappers around the native flow reader and writer handles to manage their lifetimes automatically. These wrappers can also be shared with future Fabric API objects, such as Target and Initiator, ensuring that the underlying handles remain valid for as long as they are needed.

This change deprecates the explicit destroy() methods and changes their behavior. Calling destroy() now consumes only the current Rust handle and always returns Ok(()); it does not necessarily release the underlying native handle immediately because other related objects may still reference it. The native handle is released automatically when its last owner is dropped.

# Backport requirements

v1.1

### DIFF
--- a/rust/gst-mxl-rs/src/mxlsink/imp.rs
+++ b/rust/gst-mxl-rs/src/mxlsink/imp.rs
@@ -349,27 +349,7 @@ impl BaseSinkImpl for MxlSink {
 
         // Destroy the flow writers before dropping the MXL instance they belong
         // to, then release the instance and clock.
-        if let Some(mut state) = context.state.take() {
-            match state.flow_state.take() {
-                Some(FlowState::Discrete(discrete)) => {
-                    discrete.writer.destroy().map_err(|e| {
-                        gst::error_msg!(
-                            gst::CoreError::Failed,
-                            ["Failed to destroy discrete writer: {}", e]
-                        )
-                    })?;
-                }
-                Some(FlowState::Continuous(continuous)) => {
-                    continuous.writer.destroy().map_err(|e| {
-                        gst::error_msg!(
-                            gst::CoreError::Failed,
-                            ["Failed to destroy continuous writer: {}", e]
-                        )
-                    })?;
-                }
-                None => {}
-            }
-        }
+        context.state.take();
 
         gst::info!(CAT, imp = self, "Stopped");
         Ok(())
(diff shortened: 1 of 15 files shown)


rust/gst-mxl-rs/src/mxlsink/imp.rs
```diff
@@ -349,27 +349,7 @@ impl BaseSinkImpl for MxlSink {
 
         // Destroy the flow writers before dropping the MXL instance they belong
```
jonasohland: This comment is no longer accurate

rust/mxl/src/flow/reader.rs
```diff
@@ -67,51 +110,40 @@ pub(crate) fn get_runtime_info(
...
         }
-        let result = GrainReader::new(self.context.clone(), self.reader);
-        self.reader = std::ptr::null_mut();
+        let result = GrainReader::new(self.reader);
         Ok(result)
     }
 
-    pub fn to_samples_reader(mut self) -> Result<SamplesReader> {
+    pub fn to_samples_reader(self) -> Result<SamplesReader> {
         let flow_type = self.get_info()?.config.value.common.format;
```
jonasohland: Can this also use `.is_discrete_flow()` and `common().data_format()`?

jonasohland: with `get_config_info()`

rust/mxl/src/grain/reader.rs
```diff
@@ -9,39 +9,42 @@ use crate::{
...
 
 pub struct GrainReader {
-    context: Arc<InstanceContext>,
-    reader: mxl_sys::FlowReader,
+    reader: Arc<FlowReaderInstance>,
 }
 
 /// The MXL readers and writers are not thread-safe, so we do not implement `Sync` for them, but
 /// there is no reason to not implement `Send`.
 unsafe impl Send for GrainReader {}
```
jonasohland: Is there any way to make sure that the `FlowReaderInstance` can only be mutated by the GrainReader handle? Because with this change and an Initiator owning an Arc<FlowReaderInstance> on another thread the whole thing becomes kind of unsound.

jonasohland (changes requested):

rust/mxl/src/grain/reader.rs
mlefebvre1: created a wrapper type over the Arc<T> and keeping the Arc<T> private, preventing a Target/Initiator from being able to use it.

PR #707, opened 2026-09-07
gst-mxl-rs: wait for a missing flow in create(), not negotiate()

garethsb:
## Summary

- Move the `FlowNotFound` wait out of `negotiate()` into `create()`, so a live `mxlsrc` can reach PLAYING before the producer creates the flow. Reaching PLAYING first is existing behaviour (`set_live(true)`, `start()` does not attach the reader, and `start_valid_pipeline` starts the consumer first); the bug is that `negotiate()` was the wrong place to block.
- When `unlock()` interrupts that wait, `create()` now returns `FlowError::Flushing` rather than failing negotiation. Previously the interrupted wait made `negotiate()` fail, which basesrc reported as `streaming stopped, reason not-negotiated (-4)` — a bus ERROR that kills the pipeline. `unlock()` is called on FLUSH_START and also on PLAYING to PAUSED, so a consumer waiting for a flow that did not exist yet died as soon as anything paused or flushed it.

## Test plan

New unit test `pause_while_waiting_for_missing_flow_does_not_error`: put `mxlsrc` in PLAYING with a flow id that does not exist, wait until the streaming thread is inside the `FlowNotFound` wait, cycle PAUSED then PLAYING, and assert nothing posts an ERROR on the bus.

- [x] `cargo test -p gst-mxl-rs --lib`, with default parallelism and with `--test-threads=1`
- [x] The new test fails 10 out of 10 runs against the code before this change (`not-negotiated (-4)`), and passes 10 out of 10 with it
- [x] CI `gst-mxl-rs` job

# Backport requirements

Should be backported to `release/v1.1` — labelled `backport/v1.1`.

### DIFF
--- a/rust/gst-mxl-rs/src/mxlsrc/imp.rs
+++ b/rust/gst-mxl-rs/src/mxlsrc/imp.rs
@@ -327,9 +327,8 @@ impl BaseSrcImpl for MxlSrc {
         }
 
         // `start()` does not attach the MXL reader (so PLAYING is reachable
-        // before the producer creates the flow). Attach here on the streaming
-        // thread. `init_mxl_reader` polls until the flow exists; `unlock()` sets
-        // `clock_wait.flushing` so teardown can interrupt that wait.
+        // before the producer creates the flow). Do not wait here; `create()`
+        // attaches the reader.
         let need_init = {
             let context = self
                 .context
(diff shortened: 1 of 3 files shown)


pedro-alves-ferreira (approved):
LGTM

jonasohland (approved):

PR #718, opened 2026-09-14
Reset validSlices when opening a new grain

### LINKED ISSUE #717: `validSlices` not reset when opening a new grain
`mxlFlowWriterOpenGrain()` can return a stale `validSlices` value when a new grain reuses a previously used ring-buffer slot.

When a grain is reopened while it is still incomplete, preserving `validSlices` is correct. However, when a different grain is opened, `validSlices` should start at zero. Currently, a recycled slot retains the `validSlices` value from the previous grain occupying that slot.

This can cause a newly opened grain to appear partially or fully valid before any slices have been written.

Expected behavior:

- A new grain starts with `validSlices == 0`.
- Reopening the same incomplete grain preserves its existing `validSlices`.
- Reusing a physical ring-buffer slot for a new grain resets `validSlices` to zero.

jptrainor:
Reset `validSlices` to zero when opening a grain that is different from the currently open grain. Preserve the existing value when reopening an incomplete grain so that partial write progress is retained.

This prevents a new grain from inheriting stale `validSlices` state when a ring-buffer slot is reused.

Fixes #717

# Backport requirements
Backport to release/v1.0 and release/v1.1.


vt-tv (approved):
Ohhh. good catch.  Thanks for this.

lapointejp (approved):
Good catch indeed. Thank you !

